PR #25, opened 2016-01-01
Interfaces

bruce:
- [x] Interface type definitions
- [x] Checking implementors of interfaces
- [x] TypeMap-like mapping of interfaces -> implementors
- [x] Execution with interface as the type
  - [x] Ensure "when querying for fields on an interface type, only those fields declared on the interface may be queried."
  - [ ] ~~Implementor-specific querying supported via via Fragment `on` syntax~~ (Pushed off to the Fragments PR).


bruce:
- [ ] Check for these two issues in `InterfacesMap` to make sure the interface/implementor relationship is well-defined:

From graphql-js:

```js
    invariant(
      iface instanceof GraphQLInterfaceType,
      `${type} may only implement Interface types, it cannot ` +
      `implement: ${iface}.`
    );
    if (typeof iface.resolveType !== 'function') {
      invariant(
        typeof type.isTypeOf === 'function',
        `Interface Type ${iface} does not provide a "resolveType" function ` +
        `and implementing Type ${type} does not provide a "isTypeOf" ` +
        `function. There is no way to resolve this implementing type ` +
        `during execution.`
      );
    }
```

lib/absinthe/schema/interface_map.ex
```diff
@@ -0,0 +1,75 @@
+defmodule Absinthe.Schema.InterfaceMap do
+  # Builds a mapping of interface names to the types that implement them
+
+  @moduledoc false
+
+  alias Absinthe.Type
+  alias Absithe.Schema.TypeMap
```
rubytree: Typo: `Absithe` -> `Absinthe`

Issue #30, opened 2016-01-06
Rework definition of enum values


bruce:
Looking at the need to introspect `enumValues` [in the spec](https://facebook.github.io/graphql/#sec-Enum), I see deprecation support included, something that I skipped when building out `Type.Enum` initially.

I'm thinking this, plus the fact we're having to provide a mapping of external/internal values when creating an enum probably points at a need for the type definition to be rethought a little, possible with the inclusion of a `fields`/`args`-like function to more easily define values (and, in conjunction with `deprecate`, support deprecating them).

bruce:
- [ ] Support introspection

PR #26, opened 2016-01-04
Introspection

bruce:
From the [specification section on Introspection](https://facebook.github.io/graphql/#sec-Introspection):

All tests will be in `test/specification/introspection`.

- [x] General Principles
  - [x] Naming conventions
  - [x] Documentation (No work needed, all types have `description`)
  - [x] Deprecation (No work needed, all fields/args have deprecation capability)
  - [x] Type Name Introspection (Working with the exception of the Union type; TODO in a separate PR for #28)
- [x] Schema Introspection (Some missing features broken out into separate issues/PRs; see below)
  - [x] The `__Type` type
  - [x] Type Kinds
    - [x] Scalar
    - [x] Object
    - [x] Union
    - [x] Interface
    - [x] Enum (Working with exception of `enum_values`; TODO in separate PR supporting #30)
    - [x] Input Object
    - [x] List
    - [x] Non-null
    - [x] Combining List and Non-Null
  - [x] The `__Field` Type
  - [x] The `__InputValue` Type
  - ~~The `__EnumValue` Type~~ (TODO in separate PR supporting #30)
  - [x] The `__Schema` type (Working with exception of directives, waiting on #32)
    - [x] The `__schema` field
 - ~~The `__Directive` type~~ (Needs resolution definitions, waiting on #32)

PR #34, opened 2016-01-13
Enum definition changes

bruce:
To support #30 and fill in missing Enum features from #26.


bruce:
- [ ] Enum introspection support, filling in gaps from #26

PR #42, opened 2016-01-18
Inline fragments w/ directives

bruce:
- [x] With a type condition
- [x] Without a type condition


bruce:
Had to modify the parser to support anonymous inline fragments with directives, eg:

```
query Q($skipAge: Boolean = false) {
  person {
    name
    ... @skip(if: $skipAge) {
      age
    }
  }
}
```

Issue #43, opened 2016-01-18
Support passing the current target to all resolver functions


bruce:
Right now to get the "current object" in a resolver, you need to do something like this:

```elixir
resolve: fn
  _args, execution ->
    # do something with execution.resolution.target
end
```

Since the object is really the current scope of the schema node in question, this happens often enough, and the internals of the `Execution.t` are best considered a "private API," I think it makes sense to pass it as the first argument to the resolver, eg:

```elixir
resolve: fn
  obj, _args, _execution ->
    # do something with obj
end
```

(This would bring it generally in line with other implementations.)

Or, alternately, have some other contract we expect from resolvers.

What do you think, @benwilson512 ?

bruce:
@benwilson512 note this might go hand-in-hand with #19

benwilson512:
resolve: {Mod, :fun}
resolve &Mod.fun/3

benwilson512:
Actually let's ditch the `{Mod, :fun}` version. There's an uncomfortable similarity to an archiac way of calling functions in erlang, and it makes it hard to see how one might wrap the specific resolution in pre or post resolve hooks.

bruce:
Some notes from our last discussion:

- While getting the current scope/object is important, it's not always needed (commonly it's just the arguments)
- There are other bits of info that would also be nice to get access to:
  - The current "context," for authentication/authorization, etc
  - The child selections, for query optimization (eg, avoiding N+1)
- The fewer arguments the better. While we'd like to support these things, we don't want to do what's done in JS:

```javascript
resolve: (root, {id}, source, fieldASTs) => {
  var projections = getProjection(fieldASTs);
  return User.findById(id, projections);
}
```

Four magic arguments is just too many.

Let's plan on retaining the 2 arity approach, but restructure the second value passed to the resolution to be more user-friendly, eg (and ignore the obvious Ecto pseudocode):

```elixir
# Using the AST selections under this field to select additional data
resolve: fn %{id: id}, info ->
  found = User.projections(info.selections)
  |> Repo.get(found)
  {:ok, value}
end

# Using the current object/scope (source)
resolve: fn _, info ->
  found = User |> Repo.get(info.source.author_id)
  {:ok, found}
end
```

Draft values that would be available on `info`, above:

- `:root` - The root object (See #22)
- `:source` - The parent object (source of this field)
- `:selections` - The AST selections under this field
- `:context` - The context passed to `Absinthe.run`)
- `:schema` - The current schema in use
[...]

PR #49, opened 2016-01-29
Pass more information to resolvers

bruce:
To support #43.

Add a `Absinthe.Execution.Field` struct to be passed to every resolver as the second argument. This struct will contain:

  - `:adapter` (`:: atom`) The adapter
  - `:ast_node` - (`:: Absinthe.Language.t`) The current AST node.
  - `:context` - (`:: map`) The context passed to `Absinthe.run`.
  - `:definition` - (`:: Absinthe.Type.Field.t`) The current field definition
  - `:root` - (`:: any`) The root object passed to `Absinthe.run`, if any.
  - `:schema` - (`:: Absinthe.Schema.t`) The current schema.
  - `:source` - (`:: any`) The resolved parent object; source of this field.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,21 @@
 # Changelog
 
+## v0.5.0
+
+The following changes are required if you're upgrading from the previous version:
+
+### Resolution Functions
+
+The second argument passed to resolution functions has changed from
+`Absinthe.Execution.t` to a flatter, simpler data structure,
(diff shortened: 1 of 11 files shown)


bruce:
- [ ] Fix introspection-related resolution using abstract classes (may need to pass along `parent_type`)

bruce:
Add `:parent_type` (useful for introspection resolution of composite types, especially)

PR #53, opened 2016-02-03
Schema compilation

bruce:
Supporting compilation of schema modules including:
- Compile-type type map generation, to include baking in dependent types to limit external calls
- Schema verification (eg, interface checking)
- Exceptions during compilation in the event the schema in invalid

This was driven primarily by:
- Dynamic type generation needs in CargoSense/absinthe_relay#1 leading us to use macros anyway
- Performance considerations around schema building; preparing, verifying, and caching and complexities caused by anonymous function context.


bruce:
Currently this is using the following format to set a custom name:

```elixir
object [foo: "Bar"], [
  fields: [
    baz: [type: :string]
  ]
]
```

- [] Switch to this instead, as it's more conventional and less ugly:

```elixir
object :foo, [
  name: "Bar",
  fields: [
    baz: [type: :string]
  ]
]
```

bruce:
- [ ] Ensure `description` values are populated from `@doc` attributes (w/ tests)
- [ ] Fix all compilation warnings
- [ ] Update documentation

PR #54, opened 2016-02-27
Just a thought

benwilson512:
Might be faster? Don't have time to complete now, will look later.

In order to sandbox the changes I just added the new stuff with fast_prefix instead of changing everything.

### DIFF
--- a/lib/absinthe/schema/notation.ex
+++ b/lib/absinthe/schema/notation.ex
@@ -63,7 +63,6 @@ defmodule Absinthe.Schema.Notation do
 
     block
     |> expand(env)
-    |> IO.inspect
 
     fast_close_scope(kind, env, identifier)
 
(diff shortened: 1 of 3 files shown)


lib/absinthe/schema/notation.ex
```diff
@@ -48,6 +48,94 @@ defmodule Absinthe.Schema.Notation do
...
+    fast_scope(__CALLER__, :object, identifier, [], block)
+  end
+
+  @doc false
+  # Define a notation scope that will accept attributes
+  def fast_scope(env, kind, identifier, attrs, block) do
+    fast_open_scope(kind, env, identifier, attrs)
+
+    block
+    |> expand(env)
```
benwilson512: expansion effectively evaluates the field calls, meaning we can just build the scope now instead of putting stuff in the module body that builds scope later.

PR #56, opened 2016-02-29
Update links to absinthe_plug

karlin:
These appeared to be pointing at a stale site structure on `hex.pm`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Goals:
 
 ## Related Projects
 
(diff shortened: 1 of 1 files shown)


benwilson512:
Hey! We're in the middle of moving these to their own organizations in preparation for the 1.0 release later today. There will undoubtedly be some mis-linked stuff between now and then, but it should be sorted out by later today.

Thanks!

benwilson512:
Oops, I should have read the PR, this is a different thing, and is correct. Thanks! :D

PR #59, opened 2016-02-29
updates

benwilson512:


benwilson512:
Remaining:
- [ ] Figure out why this module still exists https://github.com/CargoSense/absinthe/blob/update-docs/lib/absinthe/type/definitions.ex#L1
- [ ] Figure out why this module uses the previous one https://github.com/CargoSense/absinthe/blob/update-docs/lib/absinthe/type/field.ex

benwilson512:
- [ ] Review `deprecate`

PR #70, opened 2016-03-04
Allows for List of Lists (of Lists of... Whatever)

maxdec:
This allows the use of List of Lists, like:

```
query do:
    field :list_of_list_of_numbers, list_of(list_of(:integer)) do
      resolve fn _, _ -> {:ok, [[1, 2, 3], [4, 5, 6]]} end
    end
end
```

That example uses scalars but it's also working for list of lists of objects, as you can see in the specs.

I'm new to Elixir and Absinthe, so I hope I'm not breaking anything (especially by keeping the `ast_node` in the new `Resolution`).
I added some specs and the whole suite still passes.

### DIFF
--- a/lib/absinthe/execution/resolution/list.ex
+++ b/lib/absinthe/execution/resolution/list.ex
@@ -2,6 +2,10 @@ defimpl Absinthe.Execution.Resolution, for: Absinthe.Type.List do
 
   alias Absinthe.Execution.Resolution
 
+  def resolve(%{of_type: %{of_type: _} = wrapped_type}, execution) do
+    # That's a list of list, we unwrap and pass it while keeping the selection_set
+    do_resolve(wrapped_type, wrapped_type, execution)
+  end
   def resolve(%{of_type: wrapped_type}, %{resolution: %{ast_node: %{selection_set: nil}}} = execution) do
     # There's no selection set, do inner resolution on the wrapped type
     do_resolve(wrapped_type, wrapped_type, execution)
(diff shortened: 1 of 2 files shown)


benwilson512:
Hey! First off I'm very impressed that you waded through this stuff, resolution can be a bit crazy.

I'm in the middle of refactoring how arguments work at the moment and will look at this more closely after.

One question I have is if there's a way to make it more general. For example, this handles 2 levels of `list_of`, but not more as far as I can tell. There's probably an approach using recursion that could make that nest to arbitrary depth.

maxdec:
Actually it should work for any depth, as the clause:

```
# lib/absinthe/execution/resolution/lists.ex
def resolve(%{of_type: %{of_type: _} = wrapped_type}, execution) do
```
catches as long as a list contains another list (which could also contain a list).

I've tested with this:

```
field :big_nesting_of_numbers, list_of(list_of(list_of(list_of(:integer)))) do
  resolve fn _, _ -> {:ok, [[0],[[[1, 2, 3], [4, 5, 6]], [], [[7, 8, 9], [10, 11, 12]]]]} end
end
```

And it looks good so far. I'll add it.

But I don't know what to expect with `objects` instead of `scalars` and in case `resolve` doesn't return a list with the proper nesting/depth.

maxdec:
Actually it should work for any depth, as the clause:

```
# lib/absinthe/execution/resolution/lists.ex
def resolve(%{of_type: %{of_type: _} = wrapped_type}, execution) do
...
```
catches as long as a list contains another list (which could also contain a list).

I've tested with a nesting of 4 (with `:number`) and it looks good so far. I'll add a spec for that.

But I don't know what to expect with `objects` instead of `scalars` and in case `resolve` doesn't return a list with the proper nesting/depth.

benwilson512:
This looks good, thanks!

PR #71, opened 2016-03-05
Better arguments

benwilson512:
Solves issues with heavily nested arguments and variables.


benwilson512:
TODO:
- [ ] Verify if the built in directives should have non null inputs w/o defaults
- [ ] Add deprecation checking
- [ ] Add missing checking back to arguments
- [ ] Figure out the best route for returning `{:error, values}` or not from an execution run.

benwilson512:
On that last note @bruce it looks like Absinthe.run is supposed to return `{:ok, stuff}` for nearly all cases, which is fine.

bruce:
RE directives: looking at the [spec](https://facebook.github.io/graphql/#sec-Type-System.Directives), I'd say the `if` argument for both `@skip` and `@include` should be `non_null(:boolean)`

bruce:
RE errors -- correct -- we almost always want to push errors to the API user via the [errors property of the response](https://facebook.github.io/graphql/#sec-Errors).

benwilson512:
This test case https://github.com/absinthe-graphql/absinthe/blob/master/test/lib/absinthe/type/directive_test.exs#L91 would seem to indicate that for these boolean directives not including them should default to false.

lib/absinthe/type/built_ins/scalars.ex
```diff
@@ -102,13 +102,10 @@ defmodule Absinthe.Type.BuiltIns.Scalars do
...
-  defp parse_string(value) do
-    try do
-      {:ok, to_string(value)}
-    rescue
-      Protocol.UndefinedError -> :error
-    end
+  defp parse_string(value) when is_binary(value) do
+    {:ok, value}
   end
+  defp parse_string(_), do: :error
```
benwilson512: @bruce I think this is a better route for parsing strings. `Kernel.to_string` is entirely too forgiving with inputs. For example it'll take a list of integers `[1,2,3]` and turn it into `<<1,2,3>>` which is definitely not gonna be considered normal.

I could see an argument for coercing floats / integers to binaries, but I think that ought to be the limit.

PR #74, opened 2016-03-08
Compile-time extensibility support for Field and Object

bruce:
- Add a namespaced `__private__` key for both `Type.Field` and `Type.Object` for extensibility
- Support resolution wrapping: Passing any field resolver defined by the schema designer to a extension-defined resolver registered at the field's `[:__private__, Absinthe, :resolve]` (this allows extensions to handle pre-processing arguments passed to a resolver and post-processing any result). Later we may turn this into a more middleware like stack.
- Support a new placement rule for `Notation` macros -- `:lookup_private`, which can be used by extensions to ensure a macro defining an ancestor scope has set a certain value.
- Fix Adapter issues relating to InlineFragment traversal

### DIFF
--- a/lib/absinthe/adapter.ex
+++ b/lib/absinthe/adapter.ex
@@ -77,6 +77,10 @@ defmodule Absinthe.Adapter do
         %{node |
           selection_set: load_ast_node(node.selection_set)}
       end
+      defp load_ast_node(%Language.InlineFragment{} = node) do
+        %{node |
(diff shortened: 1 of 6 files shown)


bruce:
Tests all pass. Getting this in to support initial release of [Absinthe.Relay](https://github.com/absinthe-graphql/absinthe_relay). :rocket:

PR #80, opened 2016-03-09
Support generation of schema.json files

bruce:
Adds an `absinthe.schema.json` mix task, uses standard introspection query document from graphql-js.


bruce:
@benwilson512 :eyes:

bruce:
:rocket:

PR #81, opened 2016-03-09
add back in the ability to type check a schema at compiletime

benwilson512:


lib/absinthe/schema/rule/type_names_are_valid.ex
```diff
@@ -0,0 +1,104 @@
...
+  @moduledoc false
+
+  @description """
+  Types must exist if referenced.
+  """
+
+  def explanation(%{data: %{identifier: identifier, parent: parent}}) do
+    artifact_name = String.capitalize(parent.name)
+    """
+    #{artifact_name} #{inspect identifier} is not defined in your schema.
```
benwilson512: @bruce is this a sufficient error message?

test/support/schemas/prefix_schema.exs
```diff
@@ -3,12 +3,12 @@ defmodule PrefixSchema do
   use Absinthe.Schema
 
   query do
-    field :foo do
+    field :foo, :integer do
```
benwilson512: @bruce we may want to have some additional checks in the field macro that make sure a type of _some_ kind is set. Honestly I'd just raise an exception, it's basically a function call error.

bruce: I ran into this _exact_ thing earlier today; silent failure to add the field. Definitely need something.

lib/absinthe/schema/rule/type_names_are_valid.ex
bruce: :+1:

PR #84, opened 2016-03-11
Refactor selection set logic to handle fragment merges

bruce:
I encountered an error when trying to run the following query, generated by Relay:

```graphql
query Viewer{viewer{id,...F1}}
fragment F0 on Viewer{name,id}
fragment F1 on Viewer{id,...F0}
```

This ran afoul of the complicated selection set flattening and merge logic -- which had grown organically to support complex types like unions and interfaces.

This refactors that logic, handling cases like these and at least making things marginally easier to read.

### DIFF
--- a/lib/absinthe/execution/directives.ex
+++ b/lib/absinthe/execution/directives.ex
@@ -29,8 +29,8 @@ defmodule Absinthe.Execution.Directives do
     end)
     |> reduce_results
   end
-  def check(_, _) do
-    :ok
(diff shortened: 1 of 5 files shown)


bruce:
@benwilson512 :eyes: when you have a chance; all tests pass, but should be tested against real world code, too.

spider

bruce:
Nevermind, I ran our app test suite against it. :rocket: If you see any issues, write them up :-)

PR #89, opened 2016-03-17
Custom default resolvers

### LINKED ISSUE #77: Support overriding default resolver
The default resolver does a `Map.get/2` to retrieve a field value. In some cases (see #76) it may be desirable to modify the default resolver in some way, and in a single location.

bruce:
Resolves #77.

Adds a macro, `default_resolve`, than can be set in a schema to override the built-in default resolve function.

Eg, to support string [instead of atom] keyed maps:

```elixir
default_resolve fn
  _, %{source: source, definition: %{name: name}} ->
    case source do
      %{} ->
        {:ok, Map.get(source, name)}
       _ ->
        {:ok, nil}
    end
end
```

@benwilson512 Would appreciate a sanity and performance pass. I've tried to limit the overhead, but it's likely you'll find some improvements with some benchmarking.

### DIFF
--- a/lib/absinthe/execution/resolution/field.ex
+++ b/lib/absinthe/execution/resolution/field.ex
@@ -9,10 +9,19 @@ defimpl Absinthe.Execution.Resolution, for: Absinthe.Language.Field do
 
   @spec resolve(Absinthe.Language.Field.t,
                 Absinthe.Execution.t) :: {:ok, map} | {:error, any}
-  def resolve(%{name: name} = ast_node, %{strategy: :serial, resolution: %{parent_type: parent_type}} = execution) do
+  def resolve(%{name: name} = ast_node, %{strategy: :serial, resolution: %{parent_type: parent_type, target: target}, schema: schema} = execution) do
     field = find_field(ast_node, execution)
-    case field do
-      %{resolve: _} ->
+    case {field, schema.__absinthe_custom_default_resolve__} do
+      {%{resolve: nil}, nil} ->
+        # No custom default resolve function, we can skip arguments
+        case target do
+          %{} ->
+            {:ok, Map.get(target, name |> String.to_existing_atom)}
+          _ ->
+            {:ok, nil}
+        end
+        |> process_raw_result(ast_node, field, execution)
+      {%{}, _} ->
         case Execution.Arguments.build(ast_node, field.args, execution) do
           {:ok, args, exe} ->
             Type.Field.resolve(field, args, field_info(field, ast_node, execution))
@@ -23,7 +32,7 @@ defimpl Absinthe.Execution.Resolution, for: Absinthe.Language.Field do
             |> skip_as(:invalid, invalid, name, ast_node)
             |> Flag.as(:skip)
         end
-      nil ->
+      {nil, _} ->
         if Introspection.type?(parent_type) do
           {:skip, execution}
         else
--- a/lib/absinthe/execution/resolution/field.ex
+++ b/lib/absinthe/execution/resolution/field.ex
@@ -9,21 +9,9 @@ defimpl Absinthe.Execution.Resolution, for: Absinthe.Language.Field do
 
   @spec resolve(Absinthe.Language.Field.t,
                 Absinthe.Execution.t) :: {:ok, map} | {:error, any}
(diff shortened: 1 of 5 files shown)


lib/absinthe/execution/resolution/field.ex
```diff
@@ -9,22 +9,19 @@ defimpl Absinthe.Execution.Resolution, for: Absinthe.Language.Field do
...
-              ArgumentError -> nil
-            end
+            {:ok, Map.get(target, name |> String.to_existing_atom)}
           _ ->
-            nil
+            {:ok, nil}
         end
-        |> result(ast_node, field, execution)
-      %{resolve: _} ->
+        |> process_raw_result(ast_node, field, execution)
```
bruce: To support adding, eg, deprecation notices

lib/absinthe/execution/resolution/field.ex
```diff
@@ -9,22 +9,19 @@ defimpl Absinthe.Execution.Resolution, for: Absinthe.Language.Field do
...
-      %{resolve: nil} ->
+    case {field, schema.__absinthe_custom_default_resolve__} do
+      {%{resolve: nil}, nil} ->
+        # No custom default resolve function, we can skip arguments
         case target do
           %{} ->
-            try do
-              target |> Map.get(name |> String.to_existing_atom)
-            rescue
-              ArgumentError -> nil
```
bruce: Because we should probably just blow up if `String.to_existing_atom` fails, no?

lib/absinthe/schema/notation/writer.ex
```diff
@@ -29,10 +31,22 @@ defmodule Absinthe.Schema.Notation.Writer do
...
         def __absinthe_exports__, do: unquote(exports)
-      end
+      end,
+      custom_default_resolve(default_resolve_func)
     ]
   end
 
+  defp custom_default_resolve(nil) do
+    quote do
+      def __absinthe_custom_default_resolve__, do: nil
```
bruce: My first go at this had a `__absinthe_default_resolve__` that would return the built-in default resolver unless it had been overridden. The fact we can skip argument building for the built-in default resolver meant just using this for the custom resolver made more sense, performance-wise.

lib/absinthe/execution/resolution/field.ex
```diff
@@ -9,22 +9,19 @@ defimpl Absinthe.Execution.Resolution, for: Absinthe.Language.Field do
 
   @spec resolve(Absinthe.Language.Field.t,
                 Absinthe.Execution.t) :: {:ok, map} | {:error, any}
-  def resolve(%{name: name} = ast_node, %{strategy: :serial, resolution: %{parent_type: parent_type, target: target}} = execution) do
+  def resolve(%{name: name} = ast_node, %{strategy: :serial, resolution: %{parent_type: parent_type, target: target}, schema: schema} = execution) do
     field = find_field(ast_node, execution)
-    case field do
-      %{resolve: nil} ->
+    case {field, schema.__absinthe_custom_default_resolve__} do
```
benwilson512: Do we actually need `schema. __absinthe_custom_default_resolve__ ` to be part of this? Down on `lib/absinthe/type/field.ex:140` we set the resolve to be `schema. __absinthe_custom_default_resolve__ ` if resolve is nil, so all we need to do here is check to see if `resolve` is still nil right?

lib/absinthe/schema.ex
```diff
@@ -100,6 +100,44 @@ defmodule Absinthe.Schema do
...
+  ```
+  default_resolve fn
+    _, %{source: source, definition: %{name: name}} ->
+      case source do
+        # Only from sources that are maps
+        %{} ->
+          {:ok, Map.get(source, String.upcase(name))}
+        _ ->
+          {:ok, nil}
+      end
```
benwilson512: Because I love pattern matching on function heads:
```
fn
  _, %{source: source, definition: %{name: name}} when is_map(source) ->
    {:ok, Map.get(source, String.upcase(name))}

  _, _ ->
    {:ok, nil}
end
```

bruce: I thought you might do that ;-)

lib/absinthe/execution/resolution/field.ex
bruce: Yes, because if no _custom_ default resolver is defined, we can skip the overhead associated with argument and field info building and just resolve immediately using the built-in logic below.

PR #93, opened 2016-03-21
IDL Generation

bruce:
This is builds the foundation for conversions between schema types and their AST representations (eg, GraphQL IDL AST) -- and the eventual output of the raw IDL to the filesystem (eg, as `schema.graphql` is done for Relay in the JavaScript implementation).

There are follow-on plans in the opposite direction (IDL -> Schema, at least via static generation) that will be built later on top of these features.

## Progress 

Progress per type/concern, showing stats of (schema type -> IDL AST -> IDL iodata):

- [x] Object type
  - [x] Basic generation
  - [x] Fields
    - [x] Basic generation
    - [x] Arguments
      - [x] Basic
      - [x] NonNull
      - [x] List
  - [x] Declared interfaces (eg, _implements_)
- [x] Input Object type
  - [x] Basic 
  - [x] Defaults
    - [x] Scalars
    - [x] Lists
    - [x] Enum values
    - [x] Objects
- [x] Enum type
- [x] Union type

### DIFF
--- a/lib/absinthe/adapter.ex
+++ b/lib/absinthe/adapter.ex
@@ -212,7 +212,7 @@ defmodule Absinthe.Adapter do
         nil
       end
       def adapt(other, _) do
-        Logger.warn "Could not adapt #{inspect other}"
+        Logger.warn "Absinthe: #{__MODULE__} could not adapt #{inspect other}"
         other
       end
 
--- a/lib/absinthe/adapter.ex
+++ b/lib/absinthe/adapter.ex
@@ -57,6 +57,8 @@ defmodule Absinthe.Adapter do
 
   defmacro __using__(_) do
     quote do
+      require Logger
+
       @behaviour unquote(__MODULE__)
 
       alias Absinthe.Execution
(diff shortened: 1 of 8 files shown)


bruce:
I'm pondering pulling these functions into two protocols (eg, implementations for `Type.t` of `to_idl_ast`, implementations for `Language.t`of `to_idl_iodata`), or at least pulling them out into two modules, as they work on fundamentally different structural elements -- and they can be easily confused.

bruce:
- [ ] Adapt field names
- [ ] Refactor to `IDL.IodataConversion` and `IDL.ASTConversion` protocols (or similar)
- [ ] `@moduledoc false` / `@doc false`

bruce:
@benwilson512 Can you glance over this?

lib/absinthe/adapter.ex
```diff
@@ -250,6 +303,22 @@ defmodule Absinthe.Adapter do
...
+
+  ## Examples
+
+  ```
+  def adapt(%{definitions: definitions} = document, :load) do
+    %{document | definitions: definitions |> Enum.map(&your_custom_loader/1)}
+  end
+  ```
+  """
+  @callback adapt(Absinthe.Language.t, :load | :dump) :: Absinthe.Language.t
```
benwilson512: :+1: to these changes.

lib/absinthe/language/idl.ex
```diff
@@ -0,0 +1,287 @@
...
+    ]
+  end
+
+  defp arguments_idl_iodata([]) do
+    []
+  end
+  defp arguments_idl_iodata(arguments) do
+    [
+      "(",
+      Enum.intersperse(Enum.map(arguments, &to_idl_iodata/1), ", "),
```
benwilson512: TIL

lib/absinthe/adapter.ex
```diff
@@ -96,82 +98,131 @@ defmodule Absinthe.Adapter do
...
+        %{node | values: Enum.map(node.values, &adapt(&1, adaptation))}
+      end
+      def adapt(%{__struct__: str} = node, _) when str in @ignore_nodes do
+        node
+      end
+      def adapt(nil, _) do
+        nil
+      end
+      def adapt(other, _) do
+        Logger.warn "Could not adapt #{inspect other}"
```
bruce: This ended up being _super_ handy, btw.

benwilson512: Yeah I think it makes sense here. I do think this should be `"Absinthe: Could not adapt"` because it may be confusing if it shows up in the wild for folks.

benwilson512:
Other than the Logger note, :shipit:

lib/absinthe/adapter.ex
bruce: Ah yeah, I guess `Logger` doesn't tag with the source.

PR #94, opened 2016-03-21
Variable bug fix

benwilson512:


lib/absinthe/execution/variable.ex
```diff
@@ -132,6 +140,26 @@ defmodule Absinthe.Execution.Variable do
...
+
+      {:ok, value, meta} ->
+        inner_list_values(rest, inner_type, type_stack, var_ast, [value | acc], meta)
+
+      {:error, meta} ->
+        # No need to put an error here because which ever build_map_value clause
+        # failed should have added it already.
+        inner_list_values(rest, inner_type, type_stack, var_ast, acc, meta)
+    end
+  end
```
benwilson512: I need to try to consolidate these walk helpers but this gets stuff done for now.

PR #99, opened 2016-04-08
[WIP] rebar3 support for other BEAM languages

yurrriq:


yurrriq:
We may end up finding a nice, general solution here.

/cc @tsloughter, @talentdeficit

bruce:
@benwilson512 My only real concern is the inclusion of the `.app` file, which to me smells strongly of committing a build artifact -- but I'm not familiar enough with the best practices/issues surrounding rebar3 to offer an alternative.

yurrriq:
> ... I'm not familiar enough with the best practices/issues surrounding rebar3...

We generally don't commit `.app` files either.

benwilson512:
Yeah I'm hoping one of the guys you pinged is more familiar with the best way to make this library accessible from the rebar world because I'm not in familiar territory here. Maybe @fishcakez has an idea too?

talentdeficit:
moving this `absinthe.app` to `lib/absinthe.app.src` and changing it to:
```erlang
{application,absinthe,
             [{registered,[]},
              {description,"absinthe"},
              {vsn,"1.1.4"},
              {modules,[]},
              {applications,[kernel,stdlib,elixir,logger]}]}.
```
will make it so rebar3 will autopopulate the modules section with all modules in `lib`. that should reduce the maintenance burden

benwilson512:
@talentdeficit mentioned that he may add a rebar plugin that does what this PR does in a more generic way, so I'm gonna leave this as a branch for now until it becomes clear that this will either be available via rebar or it definitely should happen here.

Hopefully that shouldn't be a big convenience @yurrriq :)

talentdeficit:
i've got a proof of concept that can generate a .app.src from a mix.exs, i'm going to try to package it up into a rebar3 plugin that should make this pr redundant. i will keep you updated @yurrriq

benwilson512:
Hey folks! I'm gonna go ahead and close this because it seems like a working solution has been developed on the rebar side.

PR #103, opened 2016-04-19
Serialize Default Values when Introspecting

benwilson512:
…when introspecting


benwilson512:
@bruce this seems like the right call here. In theory we could specify default values in their client side form, but since the schema is internal I think server side form is much better, plus it means you can use stuff like `Timex.Date.now`. This requires serializing however when introspecting.

bruce:
@benwilson512 Is this done or in-progress?

benwilson512:
This is done, I just want you to look at it and see if it made sense.

bruce:
:shipit: if all tests are green

PR #110, opened 2016-04-26
escape the default value

### LINKED ISSUE #109: defaultValue Introspection bug
@benwilson512 Since your 924248b change I've run into a similar error to the following in a Relay app trying to parse the `schema.json` (blocking all progress):

https://github.com/graphql/graphiql/issues/89

Based on the final response, my best guess is that instead of serializing the default value as, eg `"2016-04-26T06:20Z"` it needs to be wrapped in quotes, eg: `"\"2016-04-26T06:20Z\""`

benwilson512:
fixes #109 

@bruce how does this look?

### DIFF
--- a/lib/absinthe/type/built_ins/introspection.ex
+++ b/lib/absinthe/type/built_ins/introspection.ex
@@ -259,8 +259,11 @@ defmodule Absinthe.Type.BuiltIns.Introspection do
 
         _, %{schema: schema, source: %{default_value: value, type: type}} ->
           case Absinthe.Schema.lookup_type(schema, type, unwrap: true) do
-            %{serialize: serializer} -> {:ok, serializer.(value)}
-              _ -> {:ok, to_string(value)}
+            %{serialize: serializer} ->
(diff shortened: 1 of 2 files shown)


lib/absinthe/type/built_ins/introspection.ex
```diff
@@ -259,8 +259,11 @@ defmodule Absinthe.Type.BuiltIns.Introspection do
 
         _, %{schema: schema, source: %{default_value: value, type: type}} ->
           case Absinthe.Schema.lookup_type(schema, type, unwrap: true) do
-            %{serialize: serializer} -> {:ok, serializer.(value)}
-              _ -> {:ok, to_string(value)}
+            %{serialize: serializer} ->
+              {:ok, "#{inspect serializer.(value)}"}
```
bruce: Why the interpolation? Wouldn't `inspect(serializer.(value))` work here?

bruce:
If GraphIQL likes it, :shipit:

Issue #102, opened 2016-04-14
Default values and missing variables


benwilson512:
Suppose:
```
field :thing, :thing do
  arg :id, :id, default_value: "1"
end

query Foo($id: ID) {
  thing(id: $id)
}
```
If you don't pass in an `id` variable the default is not selected.

bruce:
@benwilson512 Take a peek at this; believe this is handled in the v1.2 branch (if so, close it).

bruce:
@benwilson512 Confirm, this is being handled in #160.

benwilson512:
Fixed in v1.2 beta 0

PR #114, opened 2016-05-09
Use default_values if variables are missing

scrogson:
Hey folks, I started taking a stab at fixing #102 last night as I noticed this issue when experimenting with querying an app I'm building.

I was able to get a new test to pass, but I've introduced ~60 other failures.

I'm not yet that familiar with the codebase...but I can continue working on this to fix up the remaining failures to get my feet wet.

I believe the main issue is that the `add_argument` function clauses are expecting the "inner type" but in order to get access to the `default_value`,
the whole `Absinthe.Type.Argument` needs to be passed around.

Let me know if I'm missing something.

### DIFF
--- a/lib/absinthe/execution/arguments.ex
+++ b/lib/absinthe/execution/arguments.ex
@@ -41,7 +41,7 @@ defmodule Absinthe.Execution.Arguments do
 
   defp add_argument(%Language.Argument{value: value} = ast_node, %Type.Argument{type: inner_type} = type, type_stack, meta) do
     meta = meta |> add_deprecation_notice(type, inner_type, [type.name | type_stack], ast_node)
-    add_argument(value, inner_type, [type.name | type_stack], meta)
+    add_argument(value, type, [type.name | type_stack], meta)
   end
 
   defp add_argument(%Language.ListValue{values: values}, %Type.List{of_type: inner_type}, type_stack, meta) do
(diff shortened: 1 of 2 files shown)


lib/absinthe/execution/arguments.ex
```diff
@@ -41,7 +41,7 @@ defmodule Absinthe.Execution.Arguments do
 
   defp add_argument(%Language.Argument{value: value} = ast_node, %Type.Argument{type: inner_type} = type, type_stack, meta) do
     meta = meta |> add_deprecation_notice(type, inner_type, [type.name | type_stack], ast_node)
-    add_argument(value, inner_type, [type.name | type_stack], meta)
+    add_argument(value, type, [type.name | type_stack], meta)
```
benwilson512: We need to pass the inner type along because that's the type of the argument.

lib/absinthe/execution/arguments.ex
```diff
@@ -106,15 +106,14 @@ defmodule Absinthe.Execution.Arguments do
...
-    |> case do
+    full_type_stack = fillout_stack(schema_type.type, [], meta.schema)
+    
+    case Map.get(meta.variables, name) do
       # The variable exists, and it has the same
       # type as the argument in the schema.
       # yay! we can use it as is.
       %{value: value, type_stack: ^full_type_stack} ->
-        do_retrieve_variable(value, schema_type, type_stack, ast, meta)
+        do_retrieve_variable(value, schema_type.type, type_stack, ast, meta)
```
benwilson512: So one issue here is that a whole lot of types don't have a type field, so this isn't gonna work. schema_type isn't the variable, it's the type from the schema that the variable is supposed to be.

test/lib/absinthe/execution/arguments_test.exs
```diff
@@ -88,6 +88,14 @@ defmodule Absinthe.Execution.ArgumentsTest do
         end
       end
 
+      field :default_list_args, list_of(:integer) do
+        arg :first, list_of(:integer), default_value: [1,2]
+        arg :second, list_of(:integer), default_value: [3,4]
+        resolve fn %{first: first, second: second}, _ ->
+          {:ok, [first | second]}
```
benwilson512: This needs to be `first ++ second` otherwise you're gonna get `[[1, 2], 3, 4]`

bruce:
Any update on this?

@benwilson512, can you determine if we should build on, or close and restart a resolution for #114?

scrogson:
@bruce @benwilson512 feel free to close and work on a proper resolution. Unfortunately I don't have enough time to tackle this myself.

benwilson512:
Hey sorry for the apparent silence in this, we're actually in the middle of a large series of PRs that re-work how execution and validation happen. Those will dramatically simplify a lot of the logic here, make it both easier for us to fix this problem and (we hope) others to contribute in the future.

PR #115, opened 2016-05-09
April 2016 Support

### LINKED ISSUE #105: Update Absinthe to follow the updated GraphQL spec
The GraphQL spec has been updated in April, the list of changes can be found here:
https://github.com/facebook/graphql/releases/tag/April2016

Hopefully this is not too painful...

bruce:
Working to resolve #105

### DIFF
--- a/lib/absinthe/type/built_ins/introspection.ex
+++ b/lib/absinthe/type/built_ins/introspection.ex
@@ -1,5 +1,6 @@
 defmodule Absinthe.Type.BuiltIns.Introspection do
   use Absinthe.Schema.Notation
+  alias Absinthe.Type
 
   @moduledoc false
   object :__schema do
--- a/lib/absinthe/type/built_ins/introspection.ex
+++ b/lib/absinthe/type/built_ins/introspection.ex
@@ -56,8 +56,7 @@ defmodule Absinthe.Type.BuiltIns.Introspection do
       type: :boolean,
       resolve: fn
         _, %{source: source} ->
-          # TODO: support :query, :mutation, :subscription in `on`
-          {:ok, Enum.member?(source.on, Absinthe.Language.OperationDefinition)}
+          {:ok, Enum.any?(source.locations, &Enum.member?([:query, :mutation, :subscription], &1))}
       end
 
     field :on_fragment,
@@ -65,8 +64,7 @@ defmodule Absinthe.Type.BuiltIns.Introspection do
       type: :boolean,
       resolve: fn
         _, %{source: source} ->
-          # TODO: support :fragment_spread in `on`
-          {:ok, Enum.member?(source.on, Absinthe.Language.FragmentSpread)}
+          {:ok, Enum.member?(source.locations, :fragment_spread)}
       end
 
     field :on_field,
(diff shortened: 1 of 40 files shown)


lib/absinthe/type/built_ins/introspection.ex
```diff
@@ -51,28 +52,38 @@ defmodule Absinthe.Type.BuiltIns.Introspection do
...
         _, %{source: source} ->
+          # TODO: support :field in `on`
           {:ok, Enum.member?(source.on, Absinthe.Language.Field)}
       end
 
+    field :locations, list_of(:__directive_location)
+
   end
 
+  enum :__directive_location, values: Type.Directive.valid_location_values
```
bruce: @benwilson512 I'd like to piggyback off of the values defined in `Type.Directive.valid_location_values`, but of course the value of the attribute is quoted. We can't unquote all attr values (because we need to support resolvers, I figure); do you have a quick fix for this?

```
== Compilation error on file lib/absinthe/type/built_ins/introspection.ex ==
** (FunctionClauseError) no function clause matching in Absinthe.Type.Enum.Value.build/1
    lib/absinthe/type/enum/value.ex:31: Absinthe.Type.Enum.Value.build({{:., [line: 85], [{:__aliases__, [line: 85], [:Type, :Directive]}, :valid_location_values]}, [line: 85], []})
```

The tests won't pass after this is fixed -- there other issues -- but should get past this error.

benwilson512: Right yeah I see the issue. The easiest fix for the moment is to make valid_location_values a macro that returns the literal list of atoms. Otherwise we need to be leaving some stuff in the module body that calls these functions after macro expansion.

bruce:
Note rather than _not supporting_ introspection of directive `onOperation`, `onFragment`, and `onField` values, those fields are deprecated in favor of `locations`. This should be a bit less jarring to non-updated consumers of schema information.

bruce:
- [ ] Ensure IDL generation of directive information includes `location`, does not include the `on*` fields

bruce:
- [ ] Ensure `absinthe.schema.json` output includes `onOperation`, `onFragment`, `onField`, _and_ `locations`

bruce:
Pulling all the items from [the spec changelog](https://github.com/facebook/graphql/releases/tag/April2016) here, to make sure nothing is missed:

## Breaking

- [x] Introspection of Directives now queries for a locations Enum instead of onField, onOperation, onFragment booleans.
- [x] The @skip and @include directives are no longer allowed to be used directly on Fragment Definitions.

## Changes:

- [ ] The order of fields in the response is now well defined by the query.
- [ ] Implementation of an interface allows for additional field arguments, provided that those arguments are not required.
- [ ] Implementation of an interface allows for a covariant rather than invariant return type.
- [ ]  Directives are allowed to be different on overlapping fields.
- [ ] Two overlapping fields which can never be queried on the same object are allowed to have different arguments.
- Two overlapping fields must have compatible response shapes rather than equal return types.
- [x] An operation must not define two variables of the same name.

## Clarifications:

- [ ] Better explanation for when validation must occur, including support for memoizing the result of validation before execution.
- [ ] Provided variable values need to be coerced before execution.
- [ ] When @skip and @include directives are both used, one does not have precedence over the other.
- [ ] Overlapping field validation is recursive.
- [ ] When a non-null field throws an error, that error is propagated.

bruce:
While this isn't complete, it's work in the right direction and fixes various issues in the process. I'm going to merge this down and create discrete issues for the remaining work, especially around the changes to interfaces.

PR #120, opened 2016-06-24
shim to enable testing on 1.3 w/o removing ex-spec

benwilson512:


benwilson512:
I need to confirm that this doesn't break on 1.2.X

bruce:
Tested on v1.2.6 and v1.3.0.

PR #124, opened 2016-06-28
Blueprint: Intermediate Representation

bruce:
**This is an experimental work-in-progress.**

This defines an intermediate representation to be used as the data backbone for a more discrete phase-based approach.

Short term goal: Discussion and benchmarking.


bruce:
- [ ]  Make sure we squash this on merge -- this was based on `april-2016`, which was merged into `master` squashed, and there are a number of superfluous commits in this branch.

bruce:
- [ ] IDL: Add `description` to all the things
- [ ] IDL: Add deprecation (probably another struct) to all the things

bruce:
Goals for Absinthe phases:
- Will be more fine-grained and focused than in graphql-js
- Will work by transforming data structures 
- Will support insertion of middleware before/after each phase
- Will support ignoring phases (eg, Validation)
- Will support wholesale replacement (for the really ambitious)

Phase descriptions (we'll create good names for these later):
- Parsing (binary -> AST)
- Intermediate Representation Building (AST -> IR)
- Schema Application (IR -> IR, adding schema nodes)
- Validation (IR -> IR, adding errors)
- Directive Expansion (IR -> IR, modifying nodes based on directives)
- Execution (IR -> Execution Result; Streams? Other?)

bruce:
- [ ] IDL: Model `extend` in Blueprint
- [ ] IDL: Model `schema` (maps toplevel operation type fields) in Blueprint
- [ ] FieldGroup (InlineFragment and FragmentSpread) modeling in Blueprint

lib/absinthe/phase/operation/input.ex
```diff
@@ -0,0 +1,129 @@
...
+        {
+          update_in(node.errors, &[error(@error_not_provided) | &1]),
+          acc
+        }
+      # If it's optional, assign its (bare, unwrapped) default value to the
+      # variables map
+      {_, nil} ->
+        {
+          node,
+          %{acc | variables: Map.put(acc.variables, node.name, Blueprint.Input.unwrap(node.default_value))}
```
benwilson512: This won't be bare in the same sense that a value in JSON is bare. Default values would still look like `%Input.Integer{value: 4}` whereas variables from JSON are just `4`

Also, unwrapping eliminates any list wrapping which we definitely don't want to do here.

lib/absinthe/phase/operation/input.ex
```diff
@@ -0,0 +1,129 @@
...
+    }
+  end
+
+  # All other nodes: return unchanged
+  defp handle_node(node, acc) do
+    {node, acc}
+  end
+
+  defp process_variable_definition(node, acc) do
+    case {node.type, Map.get(acc.variables, node.name)} do
```
benwilson512: We need to ensure that a variable name is unique. IE you can't do `Foo($id: ID, $id: ID)`

lib/absinthe/phase/operation/input.ex
```diff
@@ -0,0 +1,129 @@
...
+  defp handle_node(%Blueprint.Input.Argument{value: %Blueprint.Input.Variable{name: variable_name}} = node, %{active_operation: true} = acc) do
+    {
+      struct(node, provided_value: Map.get(acc.variables, variable_name)),
+      acc
+    }
+  end
+  # Argument not using a variable: Set provided value from the literal value
+  defp handle_node(%Blueprint.Input.Argument{} = node, %{active_operation: true} = acc) do
+    {
+      struct(node, provided_value: Blueprint.Input.unwrap(node.value)),
```
benwilson512: Minor thing, but why are we doing this vs

```elixir
%{node | provided_value: Blueprint.Input.unwrap(node.value)},
```

benwilson512: So provided value is basically the raw JSON parsed variables yes? I don't think we want unwrapping here. You can't specify non null in JSON, and lists we want to keep.

benwilson512: Ah oops no sorry assigning the variable is line 79, this is if it's a literal value. I still don't think we need unwrapping though.

lib/absinthe/phase/operation/input.ex
```diff
@@ -0,0 +1,129 @@
...
+    `provided_value` of `"Things"`.
+  """
+
+  alias Absinthe.{Blueprint, Phase}
+
+  @error_not_provided "value not provided"
+
+  @spec run(Blueprint.t, %{operation_name: nil | String.t, variables: %{String.t => any}}) :: {:ok, Blueprint.t}
+  def run(input, options) do
+    acc = Map.put(options, :active_operation, false)
```
benwilson512: I think instead of doing this we should filter all selected operations from the blueprint operations list and then call prewalk on that. IE
```elixir
input.operations
|> Enum.filter(&(&1.name == options.operation_name)) # This easily allows batching
|> Blueprint.Mapper.prewalk(%{variables: options.variables}, &handle_node/2)
```

benwilson512: If we do this we no longer need to handle `process_variable_definition` as a different function or extra manual walking step. I think the only change required to support that is having the variable_definitions children walked before operation fields but that's 100% the right order to do that anyway.

benwilson512: After further thought I do have some reservations about both of these approaches, let's try to chat later.

lib/absinthe/phase/operation/input.ex
bruce: I think you're mistaking the purpose of the unwrapping -- or it's just a bad name (as I realize the uprise of "unwrapping" *is* to remove list types elsewhere). Input.unwrap should really be Input.turn_struct_into_bare_value and not I have a comment on it already saying that collections aren't supported yet.

lib/absinthe/phase/operation/input.ex
bruce: Sure, but that's a structural validation -- which makes me think -- on second thought, perhaps I shouldn't be doing the validation I'm doing here anyway, but rather just storing the provided value on the definition as well, then doing these validation in named rules used by a validation step.

lib/absinthe/phase/operation/input.ex
bruce: Not comment above that "unwrap" was a bad choice in naming and that, as noted in the code comment, the "convert to plain value" code doesn't support collection yet.

lib/absinthe/phase/operation/input.ex
bruce: Yeah, I knew we'd want to walk this differently -- but at 2am I was looking for a single pass I could easily follow. Totally fine with this being broken up. My main concern was ensuring variable definitions had the chance to add stuff to the accumulator before its sibling fields were walked, without relying on an ordering configured somewhere else.

lib/absinthe/phase/operation/input.ex
bruce: As for the use of "struct" -- just a wild hair, as it explicitly updates structs.

bruce:
- [ ] Finish converting the NoFragmentCycles tests (& remove `@tag :pending`)
- [ ] Support phases returning a set of sub-phases that should be executed (eg, for validation rules)
- [ ] Add Validation to standard document phase pipeline
- [ ] Store the list of completed phases in the Blueprint.t

bruce:
- [ ] Add execution instruction to fields and fragments
- [ ] Don't traverse `:skip` fragments/fields for flatten

bruce:
I'm closing this PR -- but only because this is really a long-running branch we'd like to make more discrete PRs against instead of an epic.

Feel free to follow the other PRs we'll be doing against `intermediate-representation` (hopefully just for the next week or so) or the final merge PR that will come out. We're stoked about the progress we're making.

PR #126, opened 2016-07-15
Foundation for directive expansion phase

bruce:
- [x] Basic Schema phase to support directive expansion
- [x] Support `expand` function in Absinthe.Type.Directive.t while maintaining backwards compat
- [x] Add `flags` for data sharing across phases
- [x] Directive expansion phase pushing nodes through directive expansion

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,10 @@
 language: elixir
-elixir: 1.2.0
+elixir: 1.3.0
 notifications:
   recipients:
     - bruce.williams@cargosense.com
     - ben.wilson@cargosense.com
 otp_release:
   - 18.0
+  - 19.0
 script: "MIX_ENV=test mix local.hex --force && MIX_ENV=test mix do deps.get, test"
--- a/lib/absinthe/blueprint.ex
+++ b/lib/absinthe/blueprint.ex
@@ -8,6 +8,7 @@ defmodule Absinthe.Blueprint do
     directives: [],
     fragments: [],
     errors: [],
+    schema: nil,
   ]
 
   @type t :: %__MODULE__{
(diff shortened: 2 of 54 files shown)


bruce:
@benwilson512 I added a `Blueprint.find` -- mostly for tests to more easily grab and test pieces of the blueprint tree. (Note it's fairly naive and doesn't stop traversing when a node is matches, etc.)

bruce:
- [ ] Add schema nodes to arguments in the `Document.Schema` phase
- [ ] Generate map of argument values to be passed to expansion function

bruce:
A note on argument values and the phases that manipulate them:

- An argument in a blueprint document has a `value` -- if a literal value is in the document for the argument, `value` is set to the nested `Blueprint.Input.t` that's converted over from the AST. Notably this means it can also be a `Blueprint.Input.Variable.t`.
- In the `Phase.Document.Arguments` phase, variable values (which were put into the variable nodes by the `Phase.Document.Variables` phase) and defaults are normalized into the same `Blueprint.Input.t` structure as if the value had been present literally in the document. This normalized value is then put into `provided_value`.
- In the `Phase.Document.Schema` phase, the `provided_value` structure is walked and `schema_node` values added to it.
- In the `Phase.Document.Directives` phase, we need to distill the structures in `provided_value` into real ("Elixir") values so that a map of argument values can be passed to the `expand` functions that directives support. Presumably we'll also need these same real values for execution. My thought is that each "real" argument value should end up alongside `value` and `provided_value` in the Argument struct, done in a phase between `Phase.Document.Schema` and `Phase.Document.Directives`, since we'll need the schema nodes to correctly parse/transform the values (although it still won't do validation -- it will just skip anything bad). 

To this end, perhaps the value fields should be more explicitly named. It's too easy to forget and type `an_argument.value` and wonder why you're not getting the result you expect.

Perhaps:

- `value` should become `literal_value`
- `provided_value` should become `normalized_value`
- The real value should be stored as `processed_value`, `data`, or something similar.

bruce:
Merging this down so we can split PRs for execution and the directive expansion phase.

PR #128, opened 2016-07-18
Generate data structure from Blueprint argument values

bruce:
_This PR is part of the v1.2 effort to restructure Absinthe's execution model around discrete, swappable phases and is not ready for use in applications._

Goals:

- [x] Rename Argument `provided_value` to `normalized_value`
- [x] Rename Argument `value` to `literal_value`
- [x] Create an Argument `data_value` field to store the data
- [x] Rename `Phase.Document.Arguments` to `Phase.Document.Arguments.Normalize`
- [x] Create `Phase.Document.Arguments.Data` and store [valid] data structure from each argument (that matches the schema) in the argument's `data_value` field.

### DIFF
--- a/lib/absinthe/blueprint/input/argument.ex
+++ b/lib/absinthe/blueprint/input/argument.ex
@@ -9,6 +9,7 @@ defmodule Absinthe.Blueprint.Input.Argument do
     # Added by phases
     schema_node: nil,
     normalized_value: nil,
+    data_value: nil,
     errors: [],
   ]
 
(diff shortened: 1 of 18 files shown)


bruce:
We'll likely encounter cases where certain combinations of wrapping types don't get distilled down into `data_value` correctly, and notably values that are invalid for the spec are still getting distilled down -- because it will be up to a validation phase put earlier to handle flagging them. I've left a couple pending tests in to check that later.

PR #129, opened 2016-07-21
initial {version} test passed

benwilson512:


lib/absinthe.ex
```diff
@@ -240,24 +240,33 @@ defmodule Absinthe do
...
 
-  def run(doc, schema, options \\ [])
-  @spec run(binary | Absinthe.Language.Source.t | Absinthe.Language.Document.t, Absinthe.Schema.t, run_opts) :: {:ok, Absinthe.Execution.result_t} | {:error, any}
-  def run(%Absinthe.Language.Document{} = document, schema, options) do
-    case Absinthe.Validation.run(document) do
-      {:ok, errors, doc} ->
-        execute(schema, doc, errors, options)
-      {:error, errors, _} ->
-        {:ok, %{errors: errors}}
+  if System.get_env("NEW_EXECUTION") do
```
benwilson512: @bruce note this flag. I'm running tests as
```
NEW_EXECUTION=true MIX_ENV=test mix do compile, test
```

Do note that you may need to do an initial `mix clean` because if `absinthe.ex` is already compiled merely changing the environment variable is not enough to trigger recompilation.

benwilson512: This is obviously a hack, and will be removed before this is merged down.

PR #130, opened 2016-07-21
Argument Validation

bruce:
Build out core validation phases for v1.2.

Prerequisites:

- [x] Build out test schema/harness helpers, as in graphql-js

Validations:

- [x] ArgumentsOfCorrectType
- [x] KnownArgumentNames
- [x] ProvidedNonNullArguments
- [x] UniqueArgumentNames
- [x] UniqueInputFieldNames


lib/absinthe/blueprint/input.ex
```diff
@@ -65,5 +67,25 @@ defmodule Absinthe.Blueprint.Input do
...
+  end
+  def inspect(%Input.Enum{value: value}) do
+    value
+  end
+  def inspect(%Input.List{values: values}) do
+    contents = values
+    |> Enum.map(&inspect/1)
+    |> Enum.join(", ")
+    "[#{contents}]"
+  end
```
benwilson512: @bruce do we want to just implement the inspect protocol?

bruce: No; perhaps I should call it something else, but I don't want to implement
Inspect and screw up debugging elsewhere.

bruce:
- [ ] Refactor: Use validation helper in existing validations

bruce:
@benwilson512 Ready for a second pair of eyes on this.

- Merged in changes from master (import fields and such) a few days ago, which I'm sure you saw in the messy diff; I shouldn't have done this -- we should have just handled it in `intermediate-representation`, but I don't think it's causing any issue.
- Merged in `intermediate-representation` so I could handle any conflicts here. We have more failing tests on this than on `intermediate-representation`, but not sure what's the cause (and we have enough failing test noise with, eg, old error messages that it's hard to tell).

Would be good to review this with you tomorrow.

bruce:
Merging down; we'll iterate on test failures in a new PR targeting `intermediate-representation`.

PR #131, opened 2016-07-23
Add make_string function to make graphql schema generation works

fazibear:


benwilson512:
This ought not to be necessary, there's some other issue here.

PR #133, opened 2016-08-15
Import fields macro

benwilson512:


benwilson512:
TODO:
- [ ] error handling.

benwilson512:
@bruce you want to look over this at all?

bruce:
I'll take a peek.

lib/absinthe/schema/notation/field_importer.ex
```diff
@@ -0,0 +1,93 @@
...
+          {:ok, %{obj | attrs: Keyword.update!(obj.attrs, :fields, fn _ -> fields end)}}
+        end
+    end
+  end
+
+  # Walk through the items we want to import fields from, get their fields,
+  # and walk to any objects THEY import to get their fields, etc.
+  defp import_fields(nil, _ctx, fields), do: {:ok, fields}
+  defp import_fields([], _ctx, fields), do: {:ok, fields}
+  defp import_fields([{obj_ref, _opts} | rest], ctx, existing_fields) do
```
benwilson512: `opts` is here in the event that we want something like

```
import_fields :foo, only: [:foo,:bar,:baz]
# OR
import_fields :foo, except: [:blah]
```

Do you think that should be included now?

bruce:
Maybe add some testing verifying that you can `import_fields` from a type brought in by `import_types` ?

benwilson512:
Good call.

lib/absinthe/schema/notation/field_importer.ex
bruce: I'd support opts, even if we're not currently using anything, just so the signature doesn't change later.

bruce:
Seems straightforward enough. I prefer being able to report all cycles (like we do in v1.2) vs just the first to break on `add_edge`, but it's a minor nit.

benwilson512:
Ah. I'll look into how we do it on the 1.2branch.

bruce:
https://github.com/absinthe-graphql/absinthe/blob/intermediate-representation/lib/absinthe/phase/document/validation/no_fragment_cycles.ex

benwilson512:
This turns out to be the wrong approach. I'm gonna grab the test file out of here and open a new branch.

PR #141, opened 2016-09-07
Include root `subscriptionType` in built-in introspection query

### LINKED ISSUE #140: Built-in introspection query should include root `subscriptionType` object
The reference introspection query at https://github.com/graphql/graphql-js/blob/master/src/utilities/introspectionQuery.js includes `subscriptionType`. Without it, attempting to use subscriptions in the `babel-relay-plugin` causes an exception deep in the babel plugin's code, which was hard to debug and caused me some frustration. For consistency with the JS and Ruby graphql libraries, the introspection query should include it.

JakeBecker:
The built-in introspection query is used by the mix tasks to generate `schema.(graphql|json)` files, and it should match up with the reference introspection query at https://github.com/graphql/graphql-js/blob/master/src/utilities/introspectionQuery.js

fixes #140

### DIFF
--- a/lib/absinthe/type/built_ins/introspection.ex
+++ b/lib/absinthe/type/built_ins/introspection.ex
@@ -26,6 +26,13 @@ defmodule Absinthe.Type.BuiltIns.Introspection do
           {:ok, Absinthe.Schema.lookup_type(schema, :mutation)}
       end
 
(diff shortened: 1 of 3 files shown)


bruce:
👍 Thanks for this!

PR #142, opened 2016-09-08
Test cases for custom scalar parsing to `nil`

jesenko:
Variables that resolve to `{:ok, nil}` in scalar `parse` function are
filtered from map of parameters passed to resolver.


bruce:
@jesenko Thanks for this! Our work on the v1.2 validation and execution framework (which we're working on wrapping up in the `intermediate-representation` branch) shifts away quite significantly in the way it handles processing arguments and how their data is passed on resolvers (including what `nil` means).

I'm going to close this for the moment, merely because I think it will need to be reworked pretty significantly in the next week anyhow. Let's readdress this after v1.2 lands.

PR #146, opened 2016-09-10
Execution test cleanup

bruce:
Goals
- Clean out old execution code and tests, moving the useful bits over for to `absinthe_test.exs`, etc.
- Restructure resolution to store `Blueprint.Document.Result` types on the current operation
- Continue working through test cases relating to execution (mostly schema, flatten, and execution data phase work)


bruce:
This puts the failed test count at 2 (!) (ignoring the ~40 relating to the old error formatting). The last 2 relate to inline fragments without type conditions, which will be handled in a separate PR.

PR #147, opened 2016-09-10
coerce enum literals to string types when its a string type

benwilson512:
This exists to fix https://github.com/absinthe-graphql/absinthe/blob/intermediate-representation/test/lib/absinthe_test.exs#L213

Note that the argument `id: foo` does not have quotes around `foo`. Is this a valid test? This branch does the requisite changes to make it pass but it seems wrong to coerce enum literals in this case.

Thoughts? @bruce

### DIFF
--- a/lib/absinthe/phase/document/arguments/coercion.ex
+++ b/lib/absinthe/phase/document/arguments/coercion.ex
@@ -21,8 +21,21 @@ defmodule Absinthe.Phase.Document.Arguments.Coercion do
     {:ok, node}
   end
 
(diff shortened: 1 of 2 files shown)


aweiker:
According to the spec, Enum values are not quoted: 

> Enum values are represented as unquoted names (ex. MOBILE_WEB)

https://github.com/facebook/graphql/blob/master/spec/Section%202%20--%20Language.md#enum-value

bruce:
This was actually fixed in my other branch. @aweiker is correct about enums, but this was about an an ID anyway (which should be quoted -- pre-v1.2 we're a bit _too_ lenient).

PR #162, opened 2016-09-17
Add validation "Variables in Allowed Position"

bruce:
Ensures variables are only used in positions whose types match (or where the variable type is a subtype).

Will require changes being made in #160 to complete.


bruce:
@benwilson512 This will be easier to implement with your `Input.Value` changes. It would be nice, btw, if your `Input.List.Item` structs had schema_nodes (eg, to use IDL notation, for an Input.List with a schema_node of `[Contact]`, its list items would have the `Contact` schema node). Then I can consistently extract the "host" node schema types where variables are placed.

bruce:
Closing this as stale. Should be part of a wider effort to add any missing validations later.

PR #166, opened 2016-09-21
Simplify/unify options to phases

bruce:
After playing with extending pipelines (for testing and absinthe_plug) for a bit, I discovered it was difficult to remember which options to pass to which phases, and which phases took different types of options. This branch:

- Simplifies `Pipeline.for_document/2` so that a nicely normalized set of options is passed to phases that need it.
- Makes the second argument to a phase consistently a keyword list
- Adds a validation to guard against empty docs


bruce:
Set one test to pending (testing tighter constraint) as the issue it addressed will be affected by #160 .

Issue #161, opened 2016-09-17
v1.1: Not returning an error when querying list of objects without fields


ScrimpyCat:
I was unsure if v1.1 is still making bug fixes, if not then this issue is fixed in v1.2.

In v1.1 if you perform an invalid query on a field that returns a list of objects, it causes an MatchError exception to be raised. Instead of returning an error.

The example schema:

```elixir
defmodule Schema do
    use Absinthe.Schema

    object :number do
        field :value, :integer
    end

    query do
        field :numbers, type: list_of(:number) do
            resolve fn args, env ->
                { :ok, [%{ value: 0 }, %{ value: 1 }, %{ value: 2 }] }
            end
        end
    end
end
```

The query: `Absinthe.run("{ numbers }", Schema)`

The MatchError occurs here https://github.com/absinthe-graphql/absinthe/blob/1371c878efecf2bceb99eed9c07b34122e9ddbbf/lib/absinthe/execution/resolution/list.ex#L24 and it is due to the `Resolution.resolve/2` function returning a `{:skip, execution}` result.

bruce:
@ScrimpyCat We're focused on v1.2 right now, but would accept a PR that fixes this for v1.1.

I put your test case (thank you for providing a solid example, BTW) in v1.2, and you'll be happy to know you'll get the following error as part of a normal result:

```
Field "numbers" of type "[Number]" must have a selection of subfields. Did you mean "numbers { ... }"?
```

ScrimpyCat:
Yeh I saw that, the new error messages look great 😄 

If I make a pr for v1.1, what would be the correct error to return? Should it return something like: `Field 'numbers': Did not resolve to match {:ok, _} or {:error, _}`? Like the current error message an object field returns if you don't specify its subfields. Or should I copy the better v1.2 message?

benwilson512:
Pretty sure this was fixed by #169

PR #169, opened 2016-09-25
v1.1 Fix: return error on list of object queries with no subfields #161

ScrimpyCat:
This fixes the issue brought up in https://github.com/absinthe-graphql/absinthe/issues/161 I copied the error message that v1.2 presents for this particular error.


benwilson512:
Looks awesome, thank you!

PR #180, opened 2016-10-25
Fixed type resolution of wrapped types

howleysv:
Type resolution was failing for wrapped union types, as it was trying to look up `%Absinthe.Type.NonNull{of_type: :foo}` from the `schema.__absinthe_type__` which returned `nil`

### DIFF
--- a/lib/absinthe/phase/document/execution/resolution.ex
+++ b/lib/absinthe/phase/document/execution/resolution.ex
@@ -54,7 +54,12 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
   defp resolve_fields(parent, bp_root, info, source) do
     parent_type = case parent.schema_node do
       %Type.Field{} = schema_node ->
-        info.schema.__absinthe_type__(schema_node.type)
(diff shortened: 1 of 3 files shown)


lib/absinthe/phase/document/execution/resolution.ex
```diff
@@ -54,7 +54,12 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
   defp resolve_fields(parent, bp_root, info, source) do
     parent_type = case parent.schema_node do
       %Type.Field{} = schema_node ->
-        info.schema.__absinthe_type__(schema_node.type)
+        type = if Type.wrapped?(schema_node.type) do
+          Type.unwrap(schema_node.type)
+        else
+          schema_node.type
+        end
+        info.schema.__absinthe_type__(type)
```
benwilson512: You can just do
```elixir
schema_node.type
|> Type.unwrap
|> info.schema.__absinthe_type__
```

There isn't much value to checking if it's wrapped or not.

benwilson512:
Thank you!!!

PR #182, opened 2016-10-27
Fixed bugs with merging lists of fragments

howleysv:
Fixed bugs with merging fragments of unions

### DIFF
--- a/lib/absinthe/phase/document/execution/resolution.ex
+++ b/lib/absinthe/phase/document/execution/resolution.ex
@@ -271,24 +271,18 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
   # The condition in an Object type and the current scope is a Union; Verify
   # that the Union has the Object type as a member.
   defp passes_type_condition?(%Type.Object{} = condition, %Type.Union{} = type, source, schema) do
-    case Type.Union.member?(type, condition) do
-      true ->
-        source_type = Type.Union.resolve_type(type, source, %{schema: schema})
-        passes_type_condition?(condition, source_type, source, schema)
-      false ->
-        false
+    with true <- Type.Union.member?(type, condition) do
+      source_type = Type.Union.resolve_type(type, source, %{schema: schema})
+      passes_type_condition?(condition, source_type, source, schema)
     end
   end
   # The condition is an Object type and the current scope is an Interface; verify
   # that the Object type is a member of the Interface and that the current source
   # object's concrete type matched the condition Object type.
   defp passes_type_condition?(%Type.Object{} = condition, %Type.Interface{} = type, source, schema) do
-    case Type.Interface.member?(type, condition) do
-      true ->
-        concrete_type = Type.Interface.resolve_type(type, source, %{schema: schema})
-        passes_type_condition?(condition, concrete_type, source, schema)
-      other ->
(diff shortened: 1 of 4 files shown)


lib/absinthe/phase/document/execution/resolution.ex
```diff
@@ -270,8 +270,14 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
...
-    Type.Union.member?(type, condition)
+  defp passes_type_condition?(%Type.Object{} = condition, %Type.Union{} = type, source, schema) do
+    case Type.Union.member?(type, condition) do
+      true ->
+        source_type = Type.Union.resolve_type(type, source, %{schema: schema})
+        passes_type_condition?(condition, source_type, source, schema)
+      false ->
+        false
+    end
   end
```
benwilson512: This could be more succinctly:
```elixir
with true <- Type.Union.member?(type, condition) do
  source_type = Type.Union.resolve_type(type, source, %{schema: schema})
  passes_type_condition?(condition, source_type, source, schema)
end
```

lib/absinthe/phase/document/result.ex
```diff
@@ -79,8 +79,13 @@ defmodule Absinthe.Phase.Document.Result do
     Map.merge(left, right, &do_deep_resolve/3)
   end
 
-  defp do_deep_resolve(_key, [%{} = left], [%{} = right]) do
-    [do_deep_merge(right, left)]
+  defp do_deep_resolve(_key, %{} = left, %{} = right) do
+    do_deep_merge(left, right)
+  end
+
+  defp do_deep_resolve(_key, [_|_] = left, [_|_] = right) when length(left) == length(right) do
```
benwilson512: When would length(left) != length(right) ?

howleysv: I wasn't sure if it was guaranteed so decided to err on the side of caution

howleysv: I think it may be necessary to check that the lists contain maps

benwilson512: This seems like the wrong place to be doing that kind of checking. I'm gonna try to dig into this later and see if I can figure out if we can compute some of this stuff ahead of time so we don't have to do so much work at this point in the pipeline.

benwilson512: Ah yeah that makes sense, if the resolution functions have some kind of side cause like loading from the DB, and the record happens to change between 2 invocations of the function, you're out of luck.

Quite honestly I think this needs to be solved at the blueprint level if possible. However it may mean enumerating all possible combinations of type conditions on a selection set. While algorithmically obnoxious, it seems like in practice the N of type conditions is wildly smaller than the M of actual results. Need to think about this more.

redjohn:
FYI, I've got a Relay app that was creating queries that weren't being resolved properly by Absinthe without this change. I understand from the comments here that this may not be the best solution, but this solution at least allows me to use the app with Absinthe 1.2. I can help provide more details if you need any.

bruce:
@redjohn We've talked this over in our Slack channel a bit with the author. This is a post v1.2.0 release priority, we just need to change the implementation to fix the problem using a slightly different approach.

PR #183, opened 2016-10-28
Resolution Plugin System. Includes Async

benwilson512:


lib/absinthe/blueprint.ex
```diff
@@ -12,7 +12,7 @@ defmodule Absinthe.Blueprint do
     # Added by phases
     flags: %{},
     errors: [],
-    result: Blueprint.Document.Result.new
+    resolution: %Blueprint.Document.Resolution{}
```
benwilson512: @bruce note this relatively minor change. It was a lot clearer to talk about the `blueprint.resolution.result` and `blueprint.resolution.acc`

PR #184, opened 2016-10-28
Batching

benwilson512:


benwilson512:
TODO:
- [ ] Configurable async await timeout value

PR #186, opened 2016-10-31
moduledoc false all the things

benwilson512:


benwilson512:
@bruce this is all just a matter of marking stuff moduledoc false, and moving existing moduledoc in those modules to comments. I vote we merge this in and then do a separate PR for content based changes, cause those will be clearer that way.

PR #187, opened 2016-11-01
Metadata utility

bruce:
This adds a `meta` notation macro to add metadata for:

- Type.Object
- Type.InputObject
- Type.Field
- Type.Enum
- Type.Scalar
- Type.Union
- Type.Interface

Example usage, taken from `schema_test.exs`:

```elixir
  defmodule MetadataSchema do
    use Absinthe.Schema

    object :foo do
      meta :sql_table, "foos"
      field :bar, :string do
        meta :nice, "yup"
      end
    end

    input_object :input_foo do
      meta :is_input, true
      field :bar, :string do
        meta :nice, "nope"
      end
    end

    enum :color do
      meta :rgb_only, true
      value :red
      value :blue
      value :green
    end

    scalar :my_scalar do
      meta :is_scalar, true
      # Missing parse and serialize
    end

    interface :named do
      meta :is_interface, true
      field :name, :string do
        meta :is_name, true
      end
    end

    union :result do
      types [:foo]
      meta :is_union, true
    end

  end
```

Metadata is stored under the existing `:__private__` field (added to some types) that is used for other internal purposes, and keyed under `:meta` (private values are namespaced by an "owner" feature).

A utility function, `Absinthe.Type.meta/2` has been added to support querying supported types for metadata information:

```
foo = Schema.lookup_type(MetadataSchema, :foo)
assert Type.meta(foo, :sql_table) == "foos"
```

If a value is not present, `nil` is returned, as with `get_in`.

### DIFF
--- a/lib/absinthe/schema/notation.ex
+++ b/lib/absinthe/schema/notation.ex
@@ -551,12 +551,22 @@ defmodule Absinthe.Schema.Notation do
     :ok
   end
 
-  @placement {:private, [under: [:field, :object]]}
+  @placement {:private, [under: [:field, :object, :input_object, :enum, :scalar, :interface, :union]]}
   @doc false
   defmacro private(owner, key, value) do
     __CALLER__
     |> recordable!(:private, @placement[:private])
     |> record_private!(owner, key, value)
+  end
+
+  @placement {:meta, [under: [:field, :object, :input_object, :enum, :scalar, :interface, :union]]}
+  @doc """
+  Defines a metadata key/value pair for a custom type.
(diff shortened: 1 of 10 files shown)


tlvenn:
It might be useful to add a `Absinthe.Type.meta/1` to lookup the whole meta map for a given type. When you need to deal with multiple metas, it's generally more convenient to lookup the whole map and operate on it rather than calling multiple individual lookups.

tlvenn:
Any chance to land this for 1.2 @bruce ?

PR #193, opened 2016-11-03
Add test for passing a scalar as an input object.

aweiker:


benwilson512:
merged into a branch so I can work on it.

PR #194, opened 2016-11-04
Stricter, spec-compliant scalar parsing rules

jparise:
The previous code supported more scalar input coercions that are allowed
by the spec. Specifically:

 - Floats are no longer considered valid inputs for an :integer field.
 - Numbers are no longer considered valid inputs for a :boolean field.

A test suite for the built-in scalar types has been included. It covers
both serialization and parsing.

This change also includes a minor description text change for the Int
type as well as two small @spec changes for parse_int and parse_float
(because, like the other parse_* functions, they can handle any input
type).


bruce:
This looks really good; narrow and clear. Good tests. Thanks especially for updating the `Int` `@desc`.  Merging.

PR #197, opened 2016-11-07
Fix interface union

howleysv:
This PR fixes a bug where a query accesses an interface fragment on an "abstract" type (union or interface) in the schema. Previous behaviour was that the types were deemed incompatible and the fragment was not resolved


bruce:
👍

PR #190, opened 2016-11-01
Added test for querying an interface implemented by a union type

howleysv:
Test currently failing, but covers querying an interface implemented by a type inside a union.

PR #199, opened 2016-11-08
Always use the field resolvers of the concrete object type

howleysv:
This fixes a bug so that when querying on unions or interfaces, the resolvers defined in the schema on the type of the concrete object are always used.

Previous behaviour was to attempt to use the resolvers defined on the union/interface field (of which there are none) and then to fall back to the default `Map.fetch` resolver.

### DIFF
--- a/lib/absinthe/phase/document/execution/resolution.ex
+++ b/lib/absinthe/phase/document/execution/resolution.ex
@@ -170,20 +170,16 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
   defp call_resolution_function(args, %{schema_node: %{name: "__" <> _}} = field, info, _) do
     field.schema_node.resolve.(args, info)
   end
-  # Interface Field
-  defp call_resolution_function(args, %{schema_node: schema_node} = field, %{parent_type: %Type.Interface{}} = info, source) when not is_nil(schema_node) do
-    concrete_type = Type.Interface.resolve_type(info.parent_type, source, info)
-    concrete_schema_node = Map.fetch!(concrete_type.fields, field.schema_node.__reference__.identifier)
-    # Try again, using the concrete type/field schema node
-    call_resolution_function(
-      args,
-      put_in(field.schema_node, concrete_schema_node),
-      put_in(info.parent_type, concrete_type),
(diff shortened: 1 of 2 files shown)


bruce:
@howleysv This looks really good. Could you add another test case or two that uses a schema-specified, non-introspection field, just because the resolution mechanism for those [special built-ins] may change somewhat in v1.3 and I want to guard against any regressions in this behavior.

howleysv:
@bruce Thanks for the feedback. Happy to add more tests, could you be more specific on what test cases you think should be covered? The current test covers both the introspection field `__typename` and the schema field `name`.

bruce:
@howleysv Ah, you're right about `name`. Perhaps a test on a union, in addition to the interface you're testing?

howleysv:
@bruce There is a further issue that prevents querying an interface on a union field because the type condition currently fails, but it is solved by https://github.com/absinthe-graphql/absinthe/pull/197

I can push the additional test to this PR, but just so you know, it will fail on master until the other fix is merged.

bruce:
@howleysv I've merged #197, and if you get the conflict fixed here + the test, happy to get this in as well. :-)

howleysv:
@bruce Actually this test from https://github.com/absinthe-graphql/absinthe/pull/197 covers the exact case we want to test:
https://github.com/thusfresh/absinthe/blob/f0ca908387ff65f5b01852d370fdc3b1a21bd217/test/lib/absinthe/union_fragment_test.exs#L106-L126

Which would have otherwise failed once the custom field resolvers were added.

bruce:
@howleysv Ok -- and what about #190? (Considering these are all so interrelated, they probably would have been better to review as one PR.)

howleysv:
@bruce https://github.com/absinthe-graphql/absinthe/pull/190 is a test for fragment merging (which isn't currently working on master). I was just leaving that open so whoever is working on the bugfix can utilise the test. This was discussed a bit on https://github.com/absinthe-graphql/absinthe/pull/182 and on slack.

bruce:
Ah, that's right -- sorry! 😄 

Getting this in.

PR #201, opened 2016-11-11
Support for custom error fields.

sylane:


lib/absinthe/phase/error.ex
```diff
@@ -15,18 +16,21 @@ defmodule Absinthe.Phase.Error do
...
   @spec new(Absinthe.Phase.t, String.t, loc_t | [loc_t]) :: t
-  def new(phase, message, location) do
+  @spec new(Absinthe.Phase.t, String.t, loc_t | [loc_t], Keyword.t) :: t
+  def new(phase, message, location, extra \\ []) do
     %__MODULE__{
       phase: phase,
       message: message,
-      locations: List.wrap(location)
+      locations: List.wrap(location),
+      extra: Enum.filter(extra, &filter_extra_item/1)
```
benwilson512: I'm not sure I understand why this is here. Why are we filtering the extra, particularly with such complicated filtering rules?

lib/absinthe/phase/document/execution/resolution.ex
```diff
@@ -143,6 +143,17 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
     |> to_result(bp_field, full_type)
     |> walk_result(acc, bp_field, full_type, info)
   end
+  defp build_result({:error, %Error{} = err}, acc, bp_field, info, _) do
```
benwilson512: I'm not entirely sure we actually need a struct here. I think it could just as easily be:
```
{:error, message, extra}
```

sylane: I am fine with what you are proposing. Not structure then ?

sylane: Would extra be `Keyword.t`, `map` or any of the two ?

lib/absinthe/phase/error.ex
sylane: Do you think of a better place to enforce that ? Do you think passing a bad extra should cause some sort of error instead of just dropping the extra field ?

lib/absinthe/phase/document/execution/resolution.ex
bruce: I would rather it's always a 2-tuple. Rather than "extra" fields, if you want to have a more stuff, provide a Keyword.t or map (or possibly a list of them) that must have :message in it.

bruce: "Flagged" tuples with the same atom but different #s of members always seem odd.

bruce: So, `{:error, String.t | Keyword.t | map | [String.t | Keyword.t | map]}`, essentially, and we normalize it.

sylane: I don't understand the `[String.t | Keyword.t | map]` part.

bruce: @sylane: yes, precisely. If you're going to add extra stuff, explicitly call out the :message, maintain the same tuple size.

bruce: The list there is just so you can provide multiple errors at once if you want to -- I know that was an idea out of the initial issue.

sylane: Supporting returning multiple error is interesting but would require quite some more changes.

lib/absinthe/phase/error.ex
bruce: I don't think filtering to enforce JSON compatibility makes sense here.

It's quite possible for people to use Absinthe w/out serializing to JSON. Absinthe doesn't even _know_ about JSON serialization -- that's all handled by Absinthe.Plug.

Personally, I'd opt for things blowing up instead of silent removal of information, which doesn't feel very idiomatic -- but if filtering is important, I think it should be a separate, general purpose phase (focused on more than just errors, probably) that's packaged with Absinthe.Plug. Thoughts, @benwilson512 ?

bruce: Also, I'm not excited about the prospect of putting everything that's additional in `extra` -- is this a convention from somewhere else? The spec just states that "GraphQL servers may provide additional entries to error as they choose to produce more helpful or machine‐readable errors, however future versions of the spec may describe additional entries to errors."

I'd argue that if I give `{:error, message: "Bad", something: "Else"}`, I should get the `{"message": "Bad", "something": "Else", "locations": [...]}` as the JSON error -- no need to sandbox the additional stuff under `"extras"` -- if people want to do that, they can use, eg, `{:error, message: "Bad", extras: [something: "Else"]}` themselves.

bruce: BTW, RE filtering: if the intention is to support being able to pass on special objects that later get transformed into better error messages (or other JSON-compatible error details), that's fine -- but the filtering out should be done in whatever phase does that transformation, not left to a generic "clean the errors" part later.

bruce:
Marking this as WIP until some issues get resolved.

lib/absinthe/phase/error.ex
sylane: I am fine with removing filtering. About extra, this is just the name of the struct field containing the extra fields, they are merged into the error object. You would get what you describe as the desired behaviour.

sylane:
After review I can merge all the changes in a single commit if you want.

lib/absinthe/phase/document/execution/resolution.ex
```diff
@@ -143,16 +143,13 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
     |> to_result(bp_field, full_type)
     |> walk_result(acc, bp_field, full_type, info)
   end
+  defp build_result({:error, params}, acc, bp_field, info, _)
+   when is_list(params) or is_map(params) do
```
bruce: Just to match with the existing style in Absinthe, please stick the guard on the same line as the function definition (yeah, I know it makes for a long line).

lib/absinthe/phase/document/execution/resolution.ex
```diff
@@ -143,16 +143,13 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
...
+    {[message: msg], extra} = Keyword.split(Enum.to_list(params), [:message])
```
bruce: If `:message` isn't provided, the "Resolution function did not return" error (defined as the fallthrough for `build_result/5` should be used -- and:

- [ ] That error should be updated to reflect the new possible return values
- [ ] Also, the `Absinthe.Type.Field.error_output` typespec needs to be changed

lib/absinthe/phase/document/execution/resolution.ex
```diff
@@ -256,11 +265,12 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
     schema.__absinthe_type__(schema_type)
   end
 
-  def error(node, message) do
+  def error(node, message, extra \\ []) do
```
bruce: Possible change this to `%{}` if we store `extra` as a map, as I note in other comment.

tlvenn:
@bruce, any chance to review the last commit made by @sylane ? Kinda eager to play with this :)

koleksiuk:
Hey @bruce, could you create a release with this change? Thanks. :)

benwilson512:
There are a couple other PRs we want to include prior to a 1.2.2 release. Should be this week however.

PR #202, opened 2016-11-11
Fix escaped characters in input literals (#165)

### LINKED ISSUE #165: Escaped characters in input literals not correctly interpreted by lexer
Writing a custom scalar for arbitrary JSON. Escaping the quotes in the JSON string results in the scalar parser receiving data with extraneous slashes. 

It appears absinthe is escaping the \'s even though they're being used to escape the quotes.

GraphQL: 

``` graphql
mutation {
    item(data: "{\"foo\": \"bar\"}") {
        id
        data
    }
}
```

The scalar parse function receives `"{\\\"foo\\\":  \\\"bar\\\"}`.

nietaki:
Fixes #165: Unescapes contents of StringValues according to https://facebook.github.io/graphql/#sec-String-Value inside the parser.

I had an idea of converting Unicode codepoints to UTF-8 bytes by hand instead of using `unicode:characters_to_binary/1`, but it seemed like a premature optimisation.

### DIFF
--- a/src/absinthe_parser.yrl
+++ b/src/absinthe_parser.yrl
@@ -263,7 +263,24 @@ extract_atom({Value, _Line}) -> Value.
 extract_binary(Value) when is_binary(Value) -> Value;
 extract_binary({Token, _Line}) -> list_to_binary(atom_to_list(Token));
 extract_binary({_Token, _Line, Value}) -> list_to_binary(Value).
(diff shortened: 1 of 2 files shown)


bruce:
This looks like great work, thanks! I can't even begin to tell you how much I appreciate parser patches 😊

I'll review this in more depth here this morning.

bruce:
Thanks! This will make it into v1.2.1.

PR #203, opened 2016-11-11
Fix argument normalization in fragments

bismark:
Adds a initialization step to the argument normalization phase to fetch
the provided values from the Operation before traversing the blueprint
to normalize all inputs.

Current behavior is dependent on ordering of node traversal in the
blueprint i.e. Operation node is expected to be visited before Input
nodes. However, Fragment nodes are visited before the Operation node so
the provided values are empty.

I am not sure this is the best solution so I am open to feedback.

### DIFF
--- a/lib/absinthe/phase/document/arguments/normalize.ex
+++ b/lib/absinthe/phase/document/arguments/normalize.ex
@@ -17,18 +17,30 @@ defmodule Absinthe.Phase.Document.Arguments.Normalize do
 
   @spec run(Blueprint.t, Keyword.t) :: {:ok, Blueprint.t}
   def run(input, _options \\ []) do
-    acc = %{provided_values: %{}}
+    acc = %{provided_values: fetch_provided_values(input)}
     {node, _} = Blueprint.prewalk(input, acc, &handle_node/2)
     {:ok, node}
   end
 
-  @spec handle_node(Blueprint.node_t, map) :: {Blueprint.node_t, map}
-  defp handle_node(%Blueprint.Document.Operation{} = node, acc) do
+  @spec fetch_provided_values(Blueprint.t) :: map
+  defp fetch_provided_values(input) do
+    default = %{}
+    {_, provided_values} = Blueprint.prewalk(input, default, &find_operation/2)
+    provided_values
+  end
+
+  @spec find_operation(Blueprint.node_t, map)
+   :: {Blueprint.node_t, map} | {:halt, Blueprint.node_t, map}
+  defp find_operation(%Blueprint.Document.Operation{} = node, _) do
     {
+      :halt,
       node,
-      %{acc | provided_values: node.provided_values}
+      node.provided_values
     }
   end
+  defp find_operation(node, acc), do: {node, acc}
+
+  @spec handle_node(Blueprint.node_t, map) :: {Blueprint.node_t, map}
   # Argument using a variable: Set provided value
   defp handle_node(%Blueprint.Input.Value{literal: %Blueprint.Input.Variable{name: variable_name}} = node, acc) do
     {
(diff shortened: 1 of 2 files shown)


bruce:
@benwilson512 Assigning to you, since you focused on this area most recently.

test/lib/absinthe/phase/document/arguments/normalize_test.exs
```diff
@@ -43,6 +55,19 @@ defmodule Absinthe.Phase.Document.Arguments.NormalizeTest do
...
+      things {
+        ... thingsFragment
+      }
+    }
+    fragment thingsFragment on Things {
+      items(id: $id) {
+        id
+      }
+    }
+  """
```
benwilson512: I'm curious how this tests illustrates the issue, because the operation is before the fragments here.

bismark: So without the fix this test fails:

```
 1) test when providing an input to a fragment normalizes the input (Absinthe.Phase.Document.Arguments.NormalizeTest)
     test/lib/absinthe/phase/document/arguments/normalize_test.exs:96
     Assertion with == failed
     code: %Blueprint.Input.String{value: "baz"} == id_argument.input_value().normalized()
     lhs:  %Absinthe.Blueprint.Input.String{errors: [], flags: %{},
            schema_node: nil, source_location: nil, value: "baz"}
     rhs:  nil
     stacktrace:
       test/lib/absinthe/phase/document/arguments/normalize_test.exs:101: (test)
```

Essentially I wanted to ensure that the arguments are correctly normalized in to the fragment node.

The ultimate behavior is that during query resolution for the above query, the arguments map passed to the resolver is empty. I can write a higher level test for that as well to help illustrate the issue if you want. (If so, any pointers to where that test should live?)

lib/absinthe/phase/document/arguments/normalize.ex
```diff
@@ -17,18 +17,30 @@ defmodule Absinthe.Phase.Document.Arguments.Normalize do
 
   @spec run(Blueprint.t, Keyword.t) :: {:ok, Blueprint.t}
   def run(input, _options \\ []) do
-    acc = %{provided_values: %{}}
+    acc = %{provided_values: fetch_provided_values(input)}
```
benwilson512: I think this can just be
```
%{provided_values: Blueprint.current_operation(input).provided_values}
```

bismark: Unfortunately this suggestion didn't work as it caused the existing `normalize_test` cases to fail. This is because the blueprint has >1 operations and no provided `operation_name` needed to mark one of the operations as `current=true`. This did expose a flaw in my original implementation that was only collecting `provided_values` from the first operation, so I've updated my solution and the test to reflect that.

benwilson512:
There must be one and only one current operation, so this is a valid
assumption.

bismark:
Is there possibly a bug in the logic here then? https://github.com/absinthe-graphql/absinthe/blob/master/lib/absinthe/phase/document/current_operation.ex#L29

The blueprint can finish the `CurrentOperation` phase without any operations being marked as `current=true` so the "must be one and only one current operation" invariant is violated.

benwilson512:
That phase selects the current operation, but it's a later phase that adds the error. By convention only phases in the Validation namespace actually add errors, it helps us keep track of them. https://github.com/absinthe-graphql/absinthe/blob/master/lib/absinthe/phase/document/validation/selected_current_operation.ex

That particular phase could probably check for some additional cases, but at a minimum it does ensure that no more than one operation is current.

bismark:
Just checking if there are any other changes required here. Cheers!

benwilson512:
Thank you!

PR #210, opened 2016-11-18
Added the ability to mutate the context variables inside root mutation resolvers

howleysv:
Added integration test to cover this behaviour, while validating that non-mutation resolvers are forbidden from attempting to mutate the context.

### DIFF
--- a/lib/absinthe/phase/document/execution/resolution.ex
+++ b/lib/absinthe/phase/document/execution/resolution.ex
@@ -143,6 +143,18 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
     |> to_result(bp_field, full_type)
     |> walk_result(acc, bp_field, full_type, info)
   end
+  ## Mutation resolver may return a response to update the context
+  defp build_result({:ok, result, context}, acc, bp_field, info, source) do
+    unless is_mutation?(info) do
+      raise Absinthe.ExecutionError, """
+      Only mutation resovers may return `{:ok, val, context}`
+      Resolving field: #{bp_field.name}
+      Resolving on: #{inspect source}
+      Got: #{inspect {:ok, result, context}}
+      """
+    end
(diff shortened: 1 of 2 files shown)


bruce:
@howleysv Can you provide more detail on the motivation for this change?

howleysv:
@bruce The use cases I'm thinking about is where you are storing authorization data inside the `context`, and your mutation necessitates an update to this state.

E.g.

- Logging in/out
- Mutation result means I now have access/lost access to resource X

tlvenn:
Not sure I understand this @howleysv. The context lifecycle is a given request so at the end of your mutation, the context will be destroyed so what the point of trying to update it ?

howleysv:
@tlvenn This new context would be available when resolving the **query** portion of the mutation response, after the mutation resolver completed successfully.

benwilson512:
Right, the idea is that if you do something like:
```
mutation LogOut {
  logOut { someFieldWithAuthRequirements someOtherField}
}
```

`someOtherField` would return fine but `someFieldWithAuthRequirements` would not. This is OK for mutations; queries would not permit mutating the context. This is also supported by Sangria for example.

tlvenn:
ha got it, thanks for the clarification both.

bruce:
My main issue with this (and @benwilson512, we've talked about this sort of thing before): I don't like differently-sized tuples with the same atom "flag" -- it feels far too much like "magic arguments" for my taste. This is especially true with `{:ok, _, _this_right_here}` which has, in the past, been proposed for providing _errors_ in addition to a value. This is a sign, in my opinion, that we shouldn't differ the # of tuple members to indicate different behaviors.

Also: this feels a lot like resolution plugin territory.

howleysv:
@bruce I'm open to implementing it differently, although I'm not sure the existing resolution plugin framework would facilitate updating the context.

benwilson512:
Closing this in favor of doing this via middleware. Docs forthcoming.

PR #213, opened 2016-11-21
Handle schema application more gracefully when provided a bad list

### LINKED ISSUE #212: Schema phase raises exception when [Int] given for expected Int input
The schema phase gives the following exception:

```
     ** (KeyError) key :of_type not found in: %Absinthe.Type.Scalar{__private__: [], __reference__: %{identifier: :integer, location: %{file: "/Users/bruce/Code/CS/absinthe/lib/absinthe/type/built_ins/scalars.ex", line: 6}, module: Absinthe.Type.BuiltIns.Scalars}, description: "The `Int` scalar type represents non-fractional signed whole numeric values.\nInt can represent values between `-(2^53 - 1)` and `2^53 - 1` since it is\nrepresented in JSON as double-precision floating point numbers specified\nby [IEEE 754](http://en.wikipedia.org/wiki/IEEE_floating_point).", name: "Int", parse: #Function<10.65559679/1 in Absinthe.Type.BuiltIns.Scalars.parse_with/2>, serialize: #Function<9.65559679/1 in Absinthe.Type.BuiltIns.Scalars.__absinthe_type__/1>}
```

When attempting to run this test:

```elixir
defmodule Absinthe.Phase.SchemaTest do
  use Absinthe.Case, async: true

  defmodule IntegerInputSchema do
    use Absinthe.Schema

    query do

      field :test, :string do
        arg :integer, :integer
        resolve fn
          _, args, _ ->
            {:ok, "ayup"}
        end
      end

    end

  end


  @query """
  { test(integer: [1]) }
  """

  it "gives a useful error message" do
    assert {:ok, _} = run(@query)
  end

  def run(query) do
    with {:ok, value} <- Absinthe.Phase.Parse.run(query),
         {:ok, value} <- Absinthe.Phase.Blueprint.run(value) do
      Absinthe.Phase.Schema.run(value, schema: IntegerInputSchema)
    end
  end

end
```

My best guess is that this is a recent regression.

bruce:
Closes #212

### DIFF
--- a/test/lib/absinthe/phase/schema_test.exs
+++ b/test/lib/absinthe/phase/schema_test.exs
@@ -1,0 +1,41 @@
+defmodule Absinthe.Phase.SchemaTest do
+  use Absinthe.Case, async: true
+
+  describe "when given [Int] for Int schema node" do
+
+    defmodule IntegerInputSchema do
+      use Absinthe.Schema
+
+      query do
+
+        field :test, :string do
+          arg :integer, :integer
+          resolve fn
+            _, _, _ ->
+              {:ok, "ayup"}
+          end
+        end
+
+      end
+
+    end
+
+
+    @query """
+    { test(integer: [1]) }
+    """
+
+    it "doesn't raise an exception" do
+      assert {:ok, _} = run(@query)
+    end
+
+  end
+
+  def run(query) do
+    with {:ok, value} <- Absinthe.Phase.Parse.run(query),
+         {:ok, value} <- Absinthe.Phase.Blueprint.run(value) do
+      Absinthe.Phase.Schema.run(value, schema: IntegerInputSchema)
+    end
+  end
+
+end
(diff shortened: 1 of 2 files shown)


benwilson512:
I don't think this is right. It shouldn't be getting to the resolution phase at all. Giving a list instead of a scalar should be a validation error, and it should jump from the validation result phase.

bruce:
This is just schema application. This is pre-validation.

test/lib/absinthe/phase/schema_test.exs
```diff
@@ -0,0 +1,41 @@
...
+
+    end
+
+
+    @query """
+    { test(integer: [1]) }
+    """
+
+    it "doesn't raise an exception" do
+      assert {:ok, _} = run(@query)
```
benwilson512: We should be testing for an error here right?

bruce: No, this is the result of the phase -- see the definition of `run` below.

benwilson512:
I'm dumb, ignore me....

PR #214, opened 2016-11-22
Fix input coercion on wrapped types

howleysv:
All but one of these tests current fail against master.

Coercion behaviour on enum input variables fails after this recent commit (seems to be related to unwrapping): https://github.com/absinthe-graphql/absinthe/commit/3294417faf2f1b54245eff8e7a8fd926a29caf53#diff-341f71b30306dc9c9354226bb6241eb3

Coercion behaviour on list inputs works on non-wrapped types, but fails for wrapper types and when input variables are used.

I suspect this can be resolved by doing the appropriate type unwrapping in `Absinthe.Phase.Document.Arguments.Coercion`, I hacked away on it a bit but I'm unclear on the reasons for the different implementations of list and enum coercion and I couldn't get them to play nicely together without breaking the other.

Any pointers?


howleysv:
I've updated this PR with fixes for the original issue.

`Coercion` is now split into 2 phases `CoerceEnums` and `CoerceLists` with tests for both phases

benwilson512:
Hey, this looks good, thanks!

PR #223, opened 2016-12-09
Fix `import_types` compilation on elixir 1.4

### LINKED ISSUE #220: Broken with Elixir v1.4.0rc1
Thanks for the great work on this.  It's been working well for us when using Elixir 1.3.4, but we recently decided to attempt to migrate to Elixir 1.4 (1.4.0rc1, https://github.com/elixir-lang/elixir/releases/tag/v1.4.0-rc.1) and getting the following compilation error:

```
== Compilation error on file lib/absinthe/type/built_ins.ex ==
** (CompileError) lib/absinthe/type/built_ins.ex: invalid quoted expression: %Absinthe.Schema.Notation.Definition{attrs: [name: "include"], builder: nil, category: :directive, file: "/docker/mount/deps/absinthe/lib/absinthe/type/built_ins.ex", identifier: :include, line: 17, opts: [], source: Absinthe.Type.BuiltIns.Directives}
    expanding macro: Absinthe.Schema.Notation.import_types/1
    lib/absinthe/type/built_ins.ex:17: Absinthe.Type.BuiltIns (module)
    (elixir) lib/kernel/parallel_compiler.ex:117: anonymous fn/4 in Kernel.ParallelCompiler.spawn_compilers/1

could not compile dependency :absinthe, "mix compile" failed. You can recompile this dependency with "mix deps.compile absinthe", update it with "mix deps.update absinthe" or clean it with "mix deps.clean absinthe"
make: *** [runtest] Error 1
```

I'd love to help fix this, but I'd probably need some help to be pointed in the right direction.

Any help is greatly appreciated.

vic:
The macro is not intended to return code to be expanded
and instead just register some module attributes on the
caller context.

Fixes #220

### DIFF
--- a/lib/absinthe/schema/notation.ex
+++ b/lib/absinthe/schema/notation.ex
@@ -995,6 +995,9 @@ defmodule Absinthe.Schema.Notation do
     type_module_ast
     |> Macro.expand(env)
     |> do_import_types(env)
+    # this macro does not produce any code to be expanded
+    # do_import_types was used to put caller module attributes
+    nil
   end
 
   defp do_import_types(type_module, env) when is_atom(type_module) do
--- a/lib/absinthe/schema/notation.ex
+++ b/lib/absinthe/schema/notation.ex
@@ -992,12 +992,11 @@ defmodule Absinthe.Schema.Notation do
   """
   defmacro import_types(type_module_ast) do
     env = __CALLER__
-    type_module_ast
-    |> Macro.expand(env)
-    |> do_import_types(env)
-    # this macro does not produce any code to be expanded
-    # do_import_types was used to put caller module attributes
-    nil
+    {:ok, _} =
+      type_module_ast
+      |> Macro.expand(env)
+      |> do_import_types(env)
+    :ok
   end
 
   defp do_import_types(type_module, env) when is_atom(type_module) do


lib/absinthe/schema/notation.ex
```diff
@@ -992,9 +992,11 @@ defmodule Absinthe.Schema.Notation do
   """
   defmacro import_types(type_module_ast) do
     env = __CALLER__
-    type_module_ast
-    |> Macro.expand(env)
-    |> do_import_types(env)
+    {:ok, _} =
+      type_module_ast
+      |> Macro.expand(env)
+      |> do_import_types(env)
```
vic: I guess, it was returning `{:ok, tuple}`  where the tuple had some structs not actually quoted expressions, and  when the macro result was being expanded an error: `%Absinthe.Schema.Notation.Definition{...} is not a valid quoted expression` was raised. Looks like previous elixir did not failed with this kind of things :/

bruce:
Good catch!

benwilson512:
Nice! Apparently the Elixir v1.4 changelog missed out on this change, I'm getting a PR built for that. I'll merge this in and get a release out here shortly, thanks!

benwilson512:
Actually after further discussion, the Elixir change is being reverted. This is a good PR anyway however, because I think explicitly returning `:ok` here makes the fact that it's just doing a side effect even more clear. Thanks!

PR #224, opened 2016-12-09
Fix Elixir 1.4 warnings on missing parens for function calls.

vic:
Automatically fixed with a little utility I just wrote:

https://gist.github.com/vic/7764646521bfa4d4449e8539e3a25892


benwilson512:
Thanks!

Issue #172, opened 2016-10-02
Add schema rule to ensure schema always has a query type


bruce:
According to https://facebook.github.io/graphql/#sec-Initial-types, a query type must be provided. We should add a schema rule (under `lib/absinthe/schema/rule`) to ensure this rule is enforced.

This change will apply cleanly to master (v1.1) and v1.2, as no changes to schema compilation were made in v1.2.

bruce:
I'm going to toss this into the v1.3 bucket, built into the planned phase-based schema validation (rules) system.

benwilson512:
Fixed in master.

PR #226, opened 2016-12-11
Schema ensure query type

Logisig:
This is for #172, i think that this is the right way to do it. Need to update the other tests aswell. But would like some input before i start fixing them all.


benwilson512:
This looks good to me, do you need any help getting the tests to pass?

Logisig:
Thanks for looking at it, just needed to get some feedback before going through 27 tests. I am going to go through the rest and see if i can get them all to pass.

bruce:
@Logisig Any update on this?

Logisig:
Sorry for the late update, I fixed the last tests and now they all pass.

benwilson512:
Gonna merge this into next.

PR #228, opened 2016-12-13
Add failing case for camel case strings resolution

sorentwo:
When a default resolver is used it doesn't play nicely with the `LanguageConventions` adapter. In this case, using camel case string keys causes an atom lookup error because the key hasn't been converted to snake case.

The initial commit adds a failing test to `default_resolver_test.exs` illustrating the issue. I don't entirely understand the flow of queries through to resolution and where/what to alter to get this working. It seems like key conversion should happen before fields are resolved, right?

Within the app I'm working on I'm manually calling `to_internal_name/2` to work around the issue:

```elixir
  default_resolve fn
    _, %{source: source, definition: %{name: name}} when is_map(source) ->
      name = LanguageConventions.to_internal_name(name, nil)

      {:ok, Map.get(source, name) || Map.get(source, String.to_existing_atom(name))}
    _, _ ->
      {:ok, nil}
  end
```

Though I could dynamically get the adapter, this seems unnecessary.

### DIFF
--- a/test/lib/absinthe/execution/default_resolver_test.exs
+++ b/test/lib/absinthe/execution/default_resolver_test.exs
@@ -1,8 +1,8 @@
 defmodule Absinthe.Execution.DefaultResolverTest do
   use Absinthe.Case, async: true
 
-  @root %{:foo => "baz", "bar" => "quux"}
-  @query "{ foo bar }"
+  @root %{:foo => "baz", "bar" => "quux", "foo_bar" => "quix"}
+  @query "{ foo bar fooBar }"
 
   describe "without a custom default resolver defined" do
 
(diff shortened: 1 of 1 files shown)


benwilson512:
I guess I'm not sure what you're trying to show here. If you want the snake case value you can just get it off of the schema node identifier.

sorentwo:
In the documentation (the specific part of which I can't find currently) it shows the definition for a `default_resolve` to automatically handle getting values out of maps. That example works perfectly, until a `name` comes through that is camel case.

Apparently that is on the `schema_node` identifier, but I never would have known that without you saying something. This`default_resolve` definition works perfectly:

```elixir
  default_resolve fn
    _, %{source: source, definition: %{schema_node: %{name: name}}} when is_map(source) ->
      {:ok, Map.get(source, name) || Map.get(source, String.to_existing_atom(name))}
    _, _ ->
      {:ok, nil}
  end
```

PR #230, opened 2016-12-14
Improve nil_value_error

skosch:
Adds the blueprint to nil_value_error output to make it easier to identify which one of the fields is missing


benwilson512:
Thanks!

benwilson512:
Should have waited for travis to finish, the function argument hadn't been uncommented so it wouldn't compile.

PR #237, opened 2016-12-21
Adding failing tests for enforcing non_null fields in input variables

nietaki:
Currently if an input object is passed as a variable and one of its `non_null` fields has a null value, Absinthe returns an error. If the `non_null` field is omitted altogether, the mutation succeeds. 

IIRC, this this behaviour doesn't manifest itself when the input object is inlined instead of being passed as a variable.

This PR adds tests verifying that the mutation returns an error in both cases.

### DIFF
--- a/test/lib/absinthe_test.exs
+++ b/test/lib/absinthe_test.exs
@@ -202,8 +202,8 @@ defmodule AbsintheTest do
 
   it "enforces non_null fields in input passed as variable" do
     query = """
-    mutation UpdateStuff($input: InputStuff!) {
-      stuff_result: update_stuff(stuff: $input)
+    query Stuff($input: InputStuff!) {
+      stuff_result: stuff(stuff: $input)
     }
     """
     result = run(query, Things, variables: %{"input" => %{"value" => 5, "nonNullField" => nil}})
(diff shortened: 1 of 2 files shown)


test/lib/absinthe_test.exs
```diff
@@ -200,6 +200,19 @@ defmodule AbsintheTest do
     assert_result {:ok, %{data: %{"thing" => %{"name" => "Foo", "value" => 100}}}}, result
   end
 
+  it "enforces non_null fields in input passed as variable" do
+    query = """
+    mutation UpdateStuff($input: InputStuff!) {
```
benwilson512: Presumably whether or not it's a mutation doesn't actually matter? or does it?

nietaki: Might not, I was porting this from one of our tests, where it happened to be a mutation. I'll see and simplify to a query if that's the case.

Should I create an Issue for tracking?

benwilson512: No need for an issue. Let's go with a query just to make it clear that it isn't something mutation specific.

nietaki:
BTW, when I tested this with absinthe_relay running against absinthe 1.2.1, this didn't seem to be breaking. Could have been introduced since.

benwilson512:
Oh are you on master then? Because 1.2.1 is still the most recently release on hex.

nietaki:
Yeah, some of the bugfixes were critical for us, so we were running off of master for a while, now we're running off of [thusfresh:mainframe/master](https://github.com/thusfresh/absinthe/tree/mainframe/master) .

benwilson512:
FYI this test also fails on v1.2.1

PR #238, opened 2016-12-22
Fix: enforcing non_null fields in input variables

nietaki:
These are the tests from https://github.com/absinthe-graphql/absinthe/pull/237 plus a proposed fix.

I was thinking of handling both cases in one function head, like this:
```elixir
defp populate_node(%Blueprint.Input.Object{fields: fields, schema_node: %{} = schema_node} = node, adapter, schema) do
  %{fields: schema_fields} = Type.unwrap_non_null(schema_node)
  fields = fill_missing_nodes(Blueprint.Input.Field, fields, schema_fields, node.source_location, adapter, schema)
  %{node | fields: fields}
end
```
but this approach failed another test:
```
1) test literal arguments boolean arguments returns a correct error when passed the wrong type (Absinthe.Execution.ArgumentsTest)
    test/lib/absinthe/execution/arguments_test.exs:398
```

I'm not sure if this is the best way of implementing the fix - hence the separate PR in case you want to use the tests and change how the fix is implemented.

### DIFF
--- a/lib/absinthe/blueprint/input/object.ex
+++ b/lib/absinthe/blueprint/input/object.ex
@@ -17,7 +17,7 @@ defmodule Absinthe.Blueprint.Input.Object do
   @type t :: %__MODULE__{
     fields: [Blueprint.Input.Field.t],
     flags: Blueprint.flags_t,
-    schema_node: nil | Absinthe.Type.InputObject.t,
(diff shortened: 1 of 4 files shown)


benwilson512:
Superseded by https://github.com/absinthe-graphql/absinthe/pull/239

PR #242, opened 2016-12-23
Fix `nil` arguments collection passed to the resolvers if all optional arguments are omitted.

nietaki:
In the resolvers we are assuming that the arguments are a map and I think it's a sane assumption to make. It used to be the case before 1.2.2, but https://github.com/absinthe-graphql/absinthe/commit/55d023cc6df48386fddf41cd15e5e9ca1d2f1813#diff-9e88b6c0fafa4e1d518da81f1c8f16a8L106 introduced the change in behaviour.

Without the fix, the change pushes the responsibility of `nil`-checking the arguments onto the individual resolvers.


benwilson512:
Crap sorry, should have caught that

PR #248, opened 2017-01-06
Support multiple field errors / Better exception message

bruce:
- Support for returning multiple field errors during resolution
- Clean up and expand exception message when a field resolver returns an invalid result.

Here's an example exception (which also includes details about the possible, valid shapes of values returned from resolvers):

```
** (Absinthe.ExecutionError) Invalid value returned from resolver.

     Resolving field:

         FailingThing

     Defined at:

         /Users/bruce/Code/CS/absinthe/test/support/things.ex:34

     Resolving on:

         %{}

     Got value:

         {:error, [code: 42]}

     ...

     You're returning an :error tuple, but did you forget to include a `:message`
     key in every custom error (map or keyword list)?

     The result must be one of the following...

     ## For a valid result

     `{:ok, any}` result will do.

     ### Examples:

     A simple integer result:

         {:ok, 1}

     Something more complex:

         {:ok, %Model.Thing{some: %{complex: :data}}}

     ## For an error result

     One or more errors for a field can be returned in a single `{:error, error_value}` tuple.

     `error_value` can be:
     - A simple error message string.
     - A map containing `:message` key, plus any additional serializable metadata.
     - A keyword list containing a `:message` key, plus any additional serializable metadata.
     - A list containing multiple of any/all of these.

     ### Examples

     A simple error message:

         {:error, "Something bad happened"}

     Multiple error messages:

         {:error, ["Something bad", "Even worse"]

     Single custom errors (note the required `:message` keys):

         {:error, message: "Unknown user", code: 21}
         {:error, %{message: "A database error occurred", details: format_db_error(some_value)}}

     Three errors of mixed types:

         {:error, ["Simple message", [message: "A keyword list error", code: 1], %{message: "A map error"}]}
```

### DIFF
--- a/lib/absinthe/phase/document/execution/resolution.ex
+++ b/lib/absinthe/phase/document/execution/resolution.ex
@@ -14,6 +14,77 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
   alias Absinthe.Phase
   use Absinthe.Phase
 
+  @resultdoc """
+  ## For a data result
+
+  `{:ok, any}` result will do.
+
+  ### Examples:
+
+  A simple integer result:
+  
+      {:ok, 1}
+
+  Something more complex:
+  
+      {:ok, %Model.Thing{some: %{complex: :data}}}
+  
+  ## For an error result
+
+  One or more errors for a field can be returned in a single `{:error, error_value}` tuple.
+
+  `error_value` can be:
+  - A simple error message string.
+  - A map containing `:message` key, plus any additional serializable metadata.
+  - A keyword list containing a `:message` key, plus any additional serializable metadata.
+  - A list containing multiple of any/all of these.
+
+  ### Examples
+
+  A simple error message:
+
+      {:error, "Something bad happened"}
+
+  Multiple error messages:
+
+      {:error, ["Something bad", "Even worse"]
+
+  Single custom errors (note the required `:message` keys):
+
+      {:error, message: "Unknown user", code: 21}
+      {:error, %{message: "A database error occurred", details: format_db_error(some_value)}}
+
+  Three errors of mixed types:
+
+      {:error, ["Simple message", [message: "A keyword list error", code: 1], %{message: "A map error"}]}
+      
+  """
+
+  @typedoc @resultdoc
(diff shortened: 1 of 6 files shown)


bruce:
- Add details on `:plugin` result to types and exception message
- Use the new `@spec`s

bruce:
- [ ] Update/move resolver function types; see https://github.com/absinthe-graphql/absinthe/blob/master/lib/absinthe/type/field.ex#L27

lib/absinthe/phase/document/execution/resolution.ex
```diff
@@ -135,49 +206,102 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
...
-        build_error_result(msg, extra, acc, bp_field, info)
-    end
+  # Error result; force wrap of single, single-value Keyword.t errors
+  defp build_result({:error, [{_, _}] = error_value} = err, acc, bp_field, info, source) do
+    build_error_result(err, [error_value], acc, bp_field, info, source)
   end
-  defp build_result({:error, msg}, acc, bp_field, info, _) do
-    build_error_result(msg, [], acc, bp_field, info)
+  # Error result; force wrap of single, multiple-value Keyword.t errors
+  defp build_result({:error, [{_, _} | _] = error_value} = err, acc, bp_field, info, source) do
```
benwilson512: This would also match the case on line `218`, since `[{:foo, 1}]` is `[{:foo, 1} | []]`

missing_rules.rb
```diff
@@ -0,0 +1,37 @@
...
+  result = File.basename(f, '.js')
+  js[result] = filesize(f)
+  result
+}
+
+puts "Implemented #{implemented.size}/#{needed.size} rules"
+puts "Done:"
+puts implemented.sort.map { |f| "- #{f} (.js LOC: #{js[f]}, .ex LOC: #{ex[f]})" }
+puts "Missing:"
+puts (needed - implemented).sort.map { |f| "- #{f} (.js LOC: #{js[f]})"}
```
benwilson512: @bruce did you mean to check this in?

bruce: Whoops, this is a helper script. I'll remove it and force push here soon.

PR #256, opened 2017-01-28
Date and Time Scalars

hellomika:
I added 4 built-in scalars:

* `utc_datetime` parses to a `DateTime` struct and serializes to an ISO8601 date and time + UTC timezone marker string
* `datetime` parses to a `NativeDateTime` struct and serializes to an ISO8601 date and time string
* `date` parses to a `Date` struct and serializes to an ISO8601 date string
* `time` parses to a `Time` struct and serializes to an ISO8601 time string

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: elixir
-elixir: 1.3.0
+elixir: 1.4.1
 notifications:
   recipients:
     - bruce.williams@cargosense.com
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: elixir
-elixir: 1.4.1
+elixir: 1.4.2
 notifications:
   recipients:
     - bruce.williams@cargosense.com
--- a/lib/absinthe/type/built_ins/scalars.ex
+++ b/lib/absinthe/type/built_ins/scalars.ex
@@ -91,6 +91,16 @@ defmodule Absinthe.Type.BuiltIns.Scalars do
     parse parse_with([Absinthe.Blueprint.Input.Date], &parse_date/1)
   end
 
+  scalar :time do
+    description """
+    The `Time` scalar type represents a time. The Time appears in a JSON
+    response as an ISO8601 formatted string.
+    """
+
+    serialize &Time.to_iso8601/1
+    parse parse_with([Absinthe.Blueprint.Input.Time], &parse_time/1)
+  end
+
   # Integers are only safe when between -(2^53 - 1) and 2^53 - 1 due to being
   # encoded in JavaScript and represented in JSON as double-precision floating
   # point numbers, as specified by IEEE 754.
@@ -173,6 +183,17 @@ defmodule Absinthe.Type.BuiltIns.Scalars do
     end
   end
   defp parse_date(_) do
+    :error
+  end
+
+  @spec parse_time(any) :: {:ok, Time.t} | :error
+  defp parse_time(value) when is_binary(value) do
+    case Time.from_iso8601(value) do
+      {:ok, time} -> {:ok, time}
+      _error -> :error
+    end
+  end
+  defp parse_time(_) do
     :error
   end
 
--- a/lib/absinthe/type/built_ins/scalars.ex
+++ b/lib/absinthe/type/built_ins/scalars.ex
@@ -81,6 +81,16 @@ defmodule Absinthe.Type.BuiltIns.Scalars do
     parse parse_with([Absinthe.Blueprint.Input.NaiveDateTime], &parse_datetime/1)
   end
 
+  scalar :date do
+    description """
+    The `Date` scalar type represents a date. The Date appears in a JSON
+    response as an ISO8601 formatted string.
+    """
+
+    serialize &Date.to_iso8601/1
+    parse parse_with([Absinthe.Blueprint.Input.Date], &parse_date/1)
+  end
+
   # Integers are only safe when between -(2^53 - 1) and 2^53 - 1 due to being
   # encoded in JavaScript and represented in JSON as double-precision floating
   # point numbers, as specified by IEEE 754.
@@ -152,6 +162,17 @@ defmodule Absinthe.Type.BuiltIns.Scalars do
     end
   end
   defp parse_datetime(_) do
+    :error
+  end
+
+  @spec parse_date(any) :: {:ok, Date.t} | :error
+  defp parse_date(value) when is_binary(value) do
+    case Date.from_iso8601(value) do
+      {:ok, date} -> {:ok, date}
+      _error -> :error
+    end
+  end
+  defp parse_date(_) do
     :error
   end
 
--- a/lib/absinthe/type/built_ins/scalars.ex
+++ b/lib/absinthe/type/built_ins/scalars.ex
@@ -59,48 +59,6 @@ defmodule Absinthe.Type.BuiltIns.Scalars do
     parse parse_with([Absinthe.Blueprint.Input.Boolean], &parse_boolean/1)
   end
 
-  scalar :utc_datetime, name: "DateTime" do
-    description """
-    The `UTC DateTime` scalar type represents a date and time in the UTC
-    timezone. The DateTime appears in a JSON response as an ISO8601 formatted
-    string, including UTC timezone ("Z").
-    """
-
-    serialize &DateTime.to_iso8601/1
-    parse parse_with([Absinthe.Blueprint.Input.DateTime], &parse_utc_datetime/1)
-  end
-
-  scalar :datetime, name: "NaiveDateTime" do
-    description """
-    The `DateTime` scalar type represents a naive date and time without
-    timezone. The DateTime appears in a JSON response as an ISO8601 formatted
-    string.
-    """
-
-    serialize &NaiveDateTime.to_iso8601/1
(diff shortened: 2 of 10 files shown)


lib/absinthe/type/built_ins/scalars.ex
```diff
@@ -59,6 +59,48 @@ defmodule Absinthe.Type.BuiltIns.Scalars do
     parse parse_with([Absinthe.Blueprint.Input.Boolean], &parse_boolean/1)
   end
 
+  scalar :utc_datetime, name: "DateTime" do
```
aweiker: Was there a specific reason to need to opt-in to timezone? I know for me, I prefer to default to being overly specific which would be using a timezone by default. This also goes against the assumptions of the Elixir conventions of DateTime is specific while NaiveDateTime is not specific. Is there any reason not to use `:naive_datetime` as the identifier to keep that consistency?

bruce:
I'm open to the idea of packaging additional, useful types (provided they only use the standard library), but I'm not keen on the idea of including types that go beyond the GraphQL Specification as built-ins -- both because of the possibility of future collision with "official" types and because we already have users with types that match these names but may have different accepted date formats, etc.

I think the reasonable choices here come down to one of:

- We launch a separate package (eg, `absinthe_contrib`) that provides these types.
- We package them along with Absinthe proper, but somewhere other than in `Absinthe.Type.BuiltIns`

In either case, use of the types by schema designers should probably be opt-in via `import_types`.

lib/absinthe/type/built_ins/scalars.ex
hellomika: I was following the same conventions than Ecto (which I concede might be driven by legacy). I was also considering what an API consumer would see during introspection and I thought this choice would make the most sense. That said being more specific doesn't hurt :)

hellomika:
@bruce that make sense, I didn't consider the standard scalars from the official specs. They even mention ISO-8601 time string as an example of custom scalar.

I like both ideas. They could be packaged in `Absinthe.Utils.Scalar`. Using one of them would look like `import_types Absinthe.Utils.Scalar.NaiveDateTime`.

lib/absinthe/type/built_ins/scalars.ex
benwilson512: Yeah given that we're using the native Elixir date time handling I think we should go with `:datetime` as always UTC.

benwilson512:
I think we make an `Absinthe.Type.Extensions` module that contains all the date time types we have here. We can already do `import_types Absinthe.Type.Extensions, only: [:datetime]` if someone wants just one type or another, so we don't need to give each type its own module.

hellomika:
Update up for review. The build is still failing due to Elixir 1.3.0.

lib/absinthe/type/extensions.ex
```diff
@@ -0,0 +1,111 @@
...
+    """
+
+    serialize &Time.to_iso8601/1
+    parse parse_with([Absinthe.Blueprint.Input.Time], &parse_time/1)
+  end
+
+  @spec parse_datetime(any) :: {:ok, DateTime.t} | :error
+  defp parse_datetime(value) when is_binary(value) do
+    case DateTime.from_iso8601(value) do
+      {:ok, datetime, _offset} -> {:ok, datetime}
```
jparise: I have something similar in my code, but I also consider a non-zero offset to be an input error:

```elixir
      case DateTime.from_iso8601(string) do
        {:ok, dt, 0}      -> {:ok, dt}
        {:ok, _, _}       -> {:error, :unsupported_offset}
        {:error, reason}  -> {:error, reason}
      end
```

That policy might be too strict for a general implementation, but I think it's slightly more honest about how offset information is treated by the server.

benwilson512:
Can you update the .travis.yml to Elixir 1.4?

lib/absinthe/type/extensions.ex
```diff
@@ -0,0 +1,112 @@
...
+
+  scalar :datetime, name: "DateTime" do
+    description """
+    The `DateTime` scalar type represents a date and time in the UTC
+    timezone. The DateTime appears in a JSON response as an ISO8601 formatted
+    string, including UTC timezone ("Z").
+    """
+
+    serialize &DateTime.to_iso8601/1
+    parse parse_with([Absinthe.Blueprint.Input.DateTime], &parse_datetime/1)
```
benwilson512: I think you're misunderstanding what `parse_with` does. These are all gonna be `Input.String` types because that's the Blueprint type that these datetime strings will initially come in as. They get parsed into datetimes sure, but prior to that they're just strings. The list you pass to `parse_with` is a list of the basic input types that are acceptable.

hellomika: Got it! I should have better reviewed when I moved the code from builtins to extensions :)

I'll update soon.

jparise:
This is a very minor comment, but I have a small preference for `.Custom` over `.Extensions` to keep the terminology in-line with the GraphQL spec's language. How does that sit with the rest of you?

tlvenn:
@hellomika any update on this PR ? Thanks in advance.

benwilson512:
We are about to release a 1.3 beta, and it would be nice if we could include these in here. I may just merge them into the 1.3 branch and then make some edits tomorrow, we'll see.

hellomika:
Hey all, apologies for the delay, time goes by fast :) I'm working on it right now.

.travis.yml
```diff
@@ -1,5 +1,5 @@
 language: elixir
-elixir: 1.3.0
+elixir: 1.4.1
```
jparise: May as well go to Elixir 1.4.2 now.

lib/absinthe/type/custom.ex
```diff
@@ -0,0 +1,96 @@
...
+  This module contains additional data types.
+
+  To use: `import_types Absinthe.Type.Custom`.
+  """
+
+  scalar :datetime, name: "DateTime" do
+    description """
+    The `DateTime` scalar type represents a date and time in the UTC
+    timezone. The DateTime appears in a JSON response as an ISO8601 formatted
+    string, including UTC timezone ("Z").
```
jparise: It might be helpful to include a brief sentence about the input format's expectations, too. Parsing an input string with a timezone offset other than 0 is considered an error.

test/lib/absinthe/type/custom_test.exs
```diff
@@ -0,0 +1,171 @@
...
+    it "serializes as an ISO8601 date and time string with UTC timezone marker" do
+      assert "2017-01-27T20:31:55Z" == serialize(:datetime, @datetime)
+    end
+
+    it "can be parsed from an ISO8601 date and time string including timezone" do
+      assert {:ok, @datetime} == parse(:datetime, "2017-01-27T20:31:55Z")
+      assert {:ok, @datetime} == parse(:datetime, "2017-01-27 20:31:55Z")
+    end
+
+    it "cannot be parsed without UTC timezone marker" do
```
jparise: Please also include a test for parsing a non-0 offset (which should also be an error).

lib/absinthe/type/custom.ex
```diff
@@ -0,0 +1,96 @@
...
+    The `Time` scalar type represents a time. The Time appears in a JSON
+    response as an ISO8601 formatted string.
+    """
+
+    serialize &Time.to_iso8601/1
+    parse &parse_time/1
+  end
+
+  @spec parse_datetime(any) :: {:ok, DateTime.t} | :error
+  defp parse_datetime(value) when is_binary(value) do
```
benwilson512: These functions are not passed simply a string IE "2016-01-01T12:12:12". Rather they are passed an input blueprint struct. You'll want to pattern match that a string input has been given by doing:
```
defp parse_datetime(%Absinthe.Blueprint.Input.String{value: value}) do
```

When you do so, there isn't really any need to do the `is_binary` check.

test/lib/absinthe/type/custom_test.exs
```diff
@@ -37,6 +38,11 @@ defmodule Absinthe.Type.CustomTest do
       assert {:ok, @datetime} == parse(:datetime, "2017-01-27 20:31:55Z")
     end
 
+    it "cannot be parsed when a UTC offset is included" do
```
jparise: Sorry to be pedantic, but I think should should be "non-zero UTC offset". I don't believe this string will not error:

`2017-01-27T20:31:55+00:00"`

hellomika: Thanks, great feedback :)

hellomika:
Changes up for review. Let me know if the high level tests are adequate!

benwilson512:
Merged into the 1.3 branch.

harmon25:
hey folks,
Looking at using the custom date scalar.
I guess 1.3 beta.0 does not have this feature implemented?
Getting the following error:

```
== Compilation error on file lib/longport/graphql/objects.ex ==
** (UndefinedFunctionError) function Absinthe.Type.Custom.__absinthe_types__/0 is undefined (module Absinthe.Type.Custom is not available)
    Absinthe.Type.Custom.__absinthe_types__()
    lib/absinthe/schema/notation.ex:1065: Absinthe.Schema.Notation.do_import_types/2
```

I will use the 1.3 branch as a mix dependency worst case...

Thanks!

benwilson512:
@harmon25 you should try the beta.2, I believe it's part of that release.

PR #260, opened 2017-02-03
Introduce complexity analysis phase

fishcakez:
By default adds 1 per field in the query. This can be overriden by
using a custom complexity analyser that is 2-arity function. The first
argument is the field's context and the second argument is the
complexity of the children.

Heavily influnced by Sangria.


lib/absinthe/phase/document/complexity.ex
```diff
@@ -0,0 +1,58 @@
+defmodule Absinthe.Phase.Document.Complexity do
+  @moduledoc false
+
+  # Analyses document complexity.
+
+  alias Absinthe.{Blueprint, Phase}
+
+  use Absinthe.Phase
+
+  @default_complexity 1
```
fishcakez: 1 doesnt seem good default if using integers because the only number less than 1 is 0, and the next number is double. should it be floats? Or default to bigger number?

lib/absinthe/phase/document/complexity.ex
```diff
@@ -0,0 +1,58 @@
...
+
+  @doc """
+  Run complexity analysis.
+  """
+  @spec run(Blueprint.t, Keyword.t) :: Phase.result_t
+  def run(input, options \\ []) do
+    %{operations: ops} = result = Blueprint.postwalk(input, &handle_node/1)
+    complexity = sum_complexity(ops)
+    max = Keyword.get(options, :max_complexity, :infinity)
```
fishcakez: Please help picking default max - if any

lib/absinthe/phase/document/complexity.ex
```diff
@@ -0,0 +1,58 @@
...
+    @default_complexity + child_complexity
+  end
+  defp field_complexity(%{complexity: complexity}, arg, child_complexity)
+       when is_function(complexity, 2) do
+    complexity.(arg, child_complexity)
+  end
+  defp field_complexity(%{complexity: {mod, fun}}, arg, child_complexity) do
+    apply(mod, fun, [arg, child_complexity])
+  end
+
```
fishcakez: Should add nice error?

lib/absinthe/pipeline.ex
```diff
@@ -92,6 +92,7 @@ defmodule Absinthe.Pipeline do
       # Prepare for Execution
       Phase.Document.CascadeInvalid,
       Phase.Document.Flatten,
+      {Phase.Document.Complexity, options},
```
fishcakez: Should this be optional phase? I think current implementation *must* go in this position.

test/lib/absinthe/phase/document/complexity_test.exs
```diff
@@ -0,0 +1,88 @@
+defmodule Absinthe.Phase.Document.Arguments.ComplexityTest do
```
fishcakez: typo in module name oops, will fix on review

lib/absinthe/blueprint/document/operation.ex
```diff
@@ -18,6 +18,7 @@ defmodule Absinthe.Blueprint.Document.Operation do
     # Populated by phases
     flags: %{},
     schema_node: nil,
+    complexity: nil,
```
benwilson512: Would it make sense to initialize this to the value of the default complexity?

Doing so would simplify a couple of function clauses in the phase.

lib/absinthe/phase/document/complexity.ex
benwilson512: The default resolution function is essentially just a Map.get, and it is for all intents and purposes impossible to resolve any faster than that. Thus I think a default complexity of 1 isn't so bad, since it doesn't seem feasible to go lower anyway.

lib/absinthe/phase/document/complexity.ex
```diff
@@ -0,0 +1,58 @@
...
+  use Absinthe.Phase
+
+  @default_complexity 1
+
+  @doc """
+  Run complexity analysis.
+  """
+  @spec run(Blueprint.t, Keyword.t) :: Phase.result_t
+  def run(input, options \\ []) do
+    %{operations: ops} = result = Blueprint.postwalk(input, &handle_node/1)
```
benwilson512: This is a great use of postwalk!

lib/absinthe/phase/document/complexity.ex
```diff
@@ -0,0 +1,58 @@
...
+  def handle_node(node) do
+    node
+  end
+
+  defp field_complexity(%{complexity: nil}, _, child_complexity) do
+    @default_complexity + child_complexity
+  end
+  defp field_complexity(%{complexity: complexity}, arg, child_complexity)
+       when is_function(complexity, 2) do
+    complexity.(arg, child_complexity)
```
benwilson512: `arg` here is really all the `args` to the function right? Do we want to pass in just the arguments or the entire field blueprint node? I'm not entirely sure what would be needed additionally out of the blueprint node other than the argument data but I'm not sure we want to limit ourselves.

lib/absinthe/pipeline.ex
benwilson512: Good question. If it's non optional I think a relatively high default should be picked. We also need to make sure to propagate the ability to set this option to things like Absinthe.Plug.

I'm inclined to include it by default, it's relatively easy to omit a phase from a pipeline if desired.

lib/absinthe/phase/document/complexity.ex
benwilson512: Not entirely sure what this is referring to.

lib/absinthe/phase/document/complexity.ex
benwilson512: This does bring up a good general question about how to weight this stuff. A field that hits the database might have a reasonable default complexity of 100. I wonder how sangria weights stuff.

lib/absinthe/phase/document/complexity.ex
fishcakez: Instead of FunctionClauseError, the resolve fun has a nice error IIRC. The resolver also does some other things with the function but I'm not sure what they are for.

lib/absinthe/blueprint/document/operation.ex
fishcakez: I didn't like to set the complexity in case someone wanted to make use of the complexity in another phase.

lib/absinthe/phase/document/complexity.ex
fishcakez: Unsure, I haven't played with the implementation enough. I considered adding an arity 3 version with more information.

lib/absinthe/phase/document/complexity.ex
fishcakez: Thats a very good point about `Map.get`

lib/absinthe/phase/document/complexity.ex
```diff
@@ -0,0 +1,58 @@
...
+  @doc """
+  Run complexity analysis.
+  """
+  @spec run(Blueprint.t, Keyword.t) :: Phase.result_t
+  def run(input, options \\ []) do
+    %{operations: ops} = result = Blueprint.postwalk(input, &handle_node/1)
+    complexity = sum_complexity(ops)
+    max = Keyword.get(options, :max_complexity, :infinity)
+    if complexity > max do
+      {:error, "complexity is #{complexity}, which is above maximum #{max}"}
```
bruce: It might be nice for us to generate an error with more metadata here, too, e.g.:

```
{:error, message: "complexity of 6 exceeds limit of 4", complexity: [calculated: 6, limit: 4]}
```

Actual message and metadata up for discussion -- I'd just like us to include more data for automated systems, generally, in addition to textual error messages when we can.

lib/absinthe/pipeline.ex
bruce: Can we just support a `nil` default, making the phase "installed" by default, but essentially no-op?

lib/absinthe/schema/notation.ex
```diff
@@ -423,6 +423,21 @@ defmodule Absinthe.Schema.Notation do
     :ok
   end
 
+  @placement {:complexity, [under: [:field]]}
+  defmacro complexity(func_ast) do
```
bruce: I am always impressed when someone manages to modify our insane Notation macros. Bravo 😉

lib/absinthe/pipeline.ex
jparise: The other consideration is the cost of running the complexity analysis itself.

lib/absinthe/phase/document/complexity.ex
jparise: Related to this, what should 0 represent (as a value returned by a complexity function)? Is there something "less" complex than `Map.get`? Would that be something that just returns a literal / constant value?

lib/absinthe/type/field.ex
```diff
@@ -111,6 +120,37 @@ defmodule Absinthe.Type.Field do
...
+
+  If no complexity function is given, the default complexity function is used,
+  which is equivalent to:
+
+      fn(_, child_complexity) -> 1 + child_complexity end
+
+  ### Custom Complexity
+
+  When accepting arguments, however, you probably need do use them for
+  something. Here's an example of definining a field that looks up a at most
```
jparise: I think the ` a ` might be unintentional here?

lib/absinthe/phase/document/complexity.ex
jparise: I also wonder how we can best help the client react to this error. For example, should there be some more specific context about the "heaviest" part of the query?

lib/absinthe/pipeline.ex
benwilson512: Well, the cost of doing the analysis is very low. It's basically just walking the tree and building an accumulator. It's certainly no worse than any other phase.

lib/absinthe/phase/document/complexity.ex
bruce: Excellent point. Perhaps a by-weight ordered list? A top 5? Hmmm...

lib/absinthe/phase/document/complexity.ex
benwilson512: @fishcakez Have you considered using the 2 arity postwalk by the way? I do like having each node tagged with its aggregate weight, but if you used 2 arity version the tree would remain unchanged and you'd still get a total.

lib/absinthe/phase/document/complexity.ex
fishcakez: Should we split the phase in two? The first phase to analyse complexity and the second to handle it? I agree that it is opaque at present.

lib/absinthe/type/field.ex
fishcakez: Thanks

lib/absinthe/phase/document/complexity.ex
fishcakez: My initial version did that but I thought complexity values might be useful in later phases. See https://github.com/absinthe-graphql/absinthe/pull/260#discussion_r99453556

bruce:
@fishcakez With regard to complexity functions having access to execution environment information, some scenarios...

Forgive me if some of these feel a bit contrived, but I've been consistently surprised with the ways people want to modify execution behavior, and access to execution environment information (the schema, the field definition, the context, the root value, etc) seems to be a common need that enables users. So far, I think the "just give them the current execution environment information so they can do whatever" has served us well.  One example: just recently, giving custom scalars' `parse` functions access to the context helped us build the `:upload` type that's now part of absinthe_plug.

Anyhow, some random ideas that just may be kooky enough for someone to want:

- A field that needs to load data from different sources (with differing performance characteristics) depending on whether the request is authenticated/authorized, and so is therefore less/more "complex" (or costly) based on the API user.
- Some system status (services down, caches up, load spiking, etc) that changes how fields needs to access data (and can be stored in the context once at the beginning of the request to then be retrieved by fields) needs to be factored into their complexity analysis.
- It would be handy to use a single complexity function that could be aware of what field it's being called for, and can change the way it's calculating dependency based on the field definition. 

I've actually built this context-aware capability into another branch, `complexity-context` -- you can see a working test case here: https://github.com/absinthe-graphql/absinthe/blob/complexity-info/test/lib/absinthe/phase/document/complexity_test.exs#L70

bruce:
BTW, I'm not advocating for dynamic complexity definitions -- but I don't think we should arbitrarily stop them from being possible.

lib/absinthe/phase/document/complexity.ex
benwilson512: Returning a literal is slightly faster than doing a Map.get but honestly both a literal and a Map.get are fast enough that the resolution complexity is less than the overhead of resolving the field in general. I feel like 1 is a good floor, I can't think of any reason to go less.

lib/absinthe/phase/document/complexity.ex
benwilson512: I would prefer to handle this as either a schema validation or at the call sight of the `complexity` macro. Then in this actual phase at runtime we can leave it as is.

lib/absinthe/blueprint/document/operation.ex
benwilson512: Makes sense.

fishcakez:
I have updated the PR based on feedback:

- Support passing more context to complexity analysers
- Support integer complexity instead of analyser function
- Split complexity analysis and complexity result handling (allows custom phases to alter or explain complexity)
- Add detailed errors
- Only run complexity analysis on current operation

lib/absinthe/pipeline.ex
fishcakez: Still always running the analysis but doing nothing with it. Thoughts?

lib/absinthe/phase/document/complexity.ex
fishcakez: I have allowed `0` because a field with no children will have a child complexity of 0 passed to it.

lib/absinthe/type/field.ex
```diff
@@ -119,10 +169,12 @@ defmodule Absinthe.Type.Field do
                default_value: any,
                args: %{(binary | atom) => Absinthe.Type.Argument.t} | nil,
                resolve: resolver_t | nil,
+               complexity: complexity_t | nil,
```
benwilson512: This should either include `non_neg_integer`, or `complexity_t` should.

lib/absinthe/phase/document/complexity/analysis.ex
```diff
@@ -0,0 +1,76 @@
...
+    complexity.(arg, child_complexity)
+  end
+  defp field_complexity(%{complexity: complexity}, arg, child_complexity, info_data, node)
+       when is_function(complexity, 3) do
+    info = struct(Complexity, Map.put(info_data, :definition, node))
+    complexity.(arg, child_complexity, info)
+  end
+  defp field_complexity(%{complexity: {mod, fun}}, arg, child_complexity, info_data, _) do
+    info = struct(Complexity, Map.put(info_data, :definition, node))
+    apply(mod, fun, [arg, child_complexity, info])
```
benwilson512: Should we `max(result_of_function, 0)` this and the other complexity function invocations? I normally wouldn't code so defensively but I'm concerned that someone who does their logic badly may end up with weird errors.

fishcakez: Is it better to error?

benwilson512: An error feels a bit severe, maybe a `Logger.warn` ?

lib/absinthe/type/field.ex
fishcakez: Fixed

lib/absinthe/phase/document/complexity/analysis.ex
benwilson512: Example situation where `max` may be useful: `{foo(limit: -1000) {y], x { ...ThisCanBeArbitrarilyComplexNow}`. If the logic for computing the complexity of `foo` is not careful to restrict the values of `limit` the `foo` field would have substantial negative complexity, which gives wide leeway for the rest of the document to be as complex as you want.

While obviously complexity functions ought to take precautions, I feel like preventing this kind of attack is within the purview of what this phase should do.

fishcakez: I agree we need to prevent that but does setting to 0 prevent it?

lib/absinthe/type/field.ex
```diff
@@ -111,6 +121,47 @@ defmodule Absinthe.Type.Field do
...
+      arg :limit, :integer
+
+      complexity fn %{limit: limit}, child_complexity ->
+        10 + limit * child_complexity
+      end
+    end
+  end
+  ```
+
+  Custom complexity functions are passed two arguments:
```
jparise: Update this documentation to also include the 3-arity version of the complexity function?

lib/absinthe/phase/document/complexity/analysis.ex
bruce: I think it would be better to error if negative complexity is returned for a field, similar to what we do when, eg, there is a non_null constraint on a field when a null is provided, or a bad resolver result. Warning and/or silently setting to 0. I think we _should_ be harsh.

benwilson512: Yeah that's fair, maxing to 0 only mitigates the issue rather than preventing it.

lib/absinthe/type/field.ex
fishcakez: Updated, thank you.

lib/absinthe/phase/document/complexity/analysis.ex
fishcakez: I've added a pretty error for this (it will also catch bad complexity analysers) and it uncovered a bug. The branch is walking fields and selections which means we walk over children multiple times.

benwilson512:
I'd prefer to solve this more generically as this isn't the first such
phase to have this issue.

One option is to set a flag on the node in the flatten phase, which is used
by the prewalk code to choose which keys to walk

fishcakez:
I've added a flag as suggested. Also disabled the complexity analyse by default, and it is enabled by an option but it is not clear where to document the phase specific options.

benwilson512:
Thank you!

PR #265, opened 2017-02-13
fix adapter config option not being used

rschmukler:
This PR fixes the new pipeline changes in 1.2 not respecting the adapter configured by `:absinthe, :adapter`.

Cheers to @russmatney for discovering the issue in our code base.

### DIFF
--- a/lib/absinthe/phase/schema.ex
+++ b/lib/absinthe/phase/schema.ex
@@ -12,6 +12,8 @@ defmodule Absinthe.Phase.Schema do
 
   alias Absinthe.{Blueprint, Type, Schema}
 
+  @default_adapter Application.get_env(:absinthe, :adapter, Absinthe.Adapter.LanguageConventions)
+
   # The approach here is pretty simple.
   # We start at the top blueprint node and set the appropriate schema node on operations
   # directives and so forth.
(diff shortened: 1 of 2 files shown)


benwilson512:
Thank you!

lib/absinthe/phase/schema.ex
```diff
@@ -24,7 +26,7 @@ defmodule Absinthe.Phase.Schema do
   @spec run(Blueprint.t, Keyword.t) :: {:ok, Blueprint.t}
   def run(input, options \\ []) do
     schema = Keyword.fetch!(options, :schema)
-    adapter = Keyword.get(options, :adapter, Absinthe.Adapter.LanguageConventions)
+    adapter = Keyword.get(options, :adapter, @default_adapter)
```
benwilson512: Hey actually I'd rather if this was configured in the `init` part of the  plug.

benwilson512:
Ah shoot I got my github repos backwards one sec.

rschmukler:
@benwilson512 any update on this? Need any changes for it to be merged?

benwilson512:
I'm honestly trying to figure out if this config based approach is a feature we want to keep. It seems like you'd be better off just specifying the adapter on the plug.

benwilson512:
Hey!

I appreciate that you noticed this, but I think with 1.3 coming up we're actually going to deprecate this option. It is better to configure this via whatever mechanism you're using to expose Absinthe.

PR #271, opened 2017-02-21
Update documentation type in Interface module

simonprev:
I was following the doc and was presented this error:

```elixir
(CaseClauseError) no case clause matching: {:ok, :confirmed_collaborator}
 (absinthe) lib/absinthe/type/interface.ex:96: Absinthe.Type.Interface.resolve_type/3
```

Changing this to the same signature as the `union`’s `resolve_type` function made it work.

🎉

### DIFF
--- a/lib/absinthe/type/interface.ex
+++ b/lib/absinthe/type/interface.ex
@@ -23,7 +23,7 @@ defmodule Absinthe.Type.Interface do
     resolve_type fn
       %{age: _}, _ -> :person
       %{employee_count: _}, _ -> :business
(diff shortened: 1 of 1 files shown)


bruce:
Thanks, @simonprev -- very good catch. This is an artifact from an old implementation, thanks for finding it.

I think the final case, `_ -> :error` should probably be changed to `_ -> nil` as well, as otherwise Absinthe will assume there's an `:error` type.

simonprev:
@bruce Done!

PR #273, opened 2017-02-22
Support atoms for error messages

olafura:
Sometimes you get atoms in {:error, message} and instead of
people handling the converation themselves it's better just
returning the string equivilant of the atom inline with what
json libraries like Poison do.

I'm not sure how to test this because I can't find any other tests
for this.

Also I could have an implementation that handles everything
either through to_string or inspect.


benwilson512:
Can you elaborate on use cases for this?

When are people returning `{:error, message: :foo}`

olafura:
@benwilson512 We are using guardian and it's returning: {:error, :token_expired}

olafura:
#274 is the new PR

PR #274, opened 2017-02-22
Use to_string for error messages

olafura:
Sometimes you get other types in {:error, value}, like atom and integers,
and instead of people handling the converation themselves it's better just
returning the string equivalent of the atom inline with what json libraries
like Poison do or any other type that can be a string.

An example would be if you are doing `with` in you body you get the error
message from the library you are using. In guardian case you could get:
`{:error, :token_expired}`

### DIFF
--- a/lib/absinthe/phase/document/execution/resolution.ex
+++ b/lib/absinthe/phase/document/execution/resolution.ex
@@ -58,6 +58,11 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
 
       {:error, ["Simple message", [message: "A keyword list error", code: 1], %{message: "A map error"}]}
 
+  Generic handler for interoperability with errors from other libraries:
+      {:error, :foo}
+      {:error, 1.0}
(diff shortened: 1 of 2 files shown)


bruce:
This change seems fine, but:

- [ ] Related specs need to be updated
- [ ] The text of the large resolution error needs to note it as a valid rerun value

olafura:
@bruce I'm fine with completing those if you point me in the right direction ;)

bruce:
- The `Absinthe.Type.Field.error_value` type should probably include `serializable` as an option (preferably at the end)
- `@error_detail` in `Absinthe.Phase.Document.Execution.Resolution` needs to document this form with an example

olafura:
@bruce Hope the wording is ok, I think generally people are going to use it for the interoperability rather than just dumping random things into `{:error, value}`

bruce:
Superseded by #286, to be included in v1.3 release.

PR #276, opened 2017-02-23
fix: when scalar parsing failed it return 500

### LINKED ISSUE #275: When scalar parsing failed, it returns a server error
Say I have a scalar type:
```elixir
...
@desc """
 format: {YYYY}-{M}-{D}
"""
scalar :date do
   parse &Timex.parse(&1.value, "{YYYY}-{M}-{D}")
   serialize &Timex.format!(&1, "{YYYY}-{M}-{D}")
end
...
```

and use it in my input object

```elixir
 input_object :designer_input do
   ....
   field :birthdate, :date
   ....
end
```

following  mutation will get a status 500
```graphql
mutation {
  createDesigner(designer: {birthdate:"212231213" }){
    name
  }
}
```

the log looks like:
```
[error] #PID<0.495.0> running Hall.Endpoint terminated
Server: localhost:4000 (http)
Request: POST /api
    ** (CaseClauseError) no case clause matching: {:error, "Expected `1-4 digit year` at line 1, column 1."}
        (absinthe) lib/absinthe/phase/document/arguments/parse.ex:35: Absinthe.Phase.Document.Arguments.Parse.build_value/3
        (absinthe) lib/absinthe/phase/document/arguments/parse.ex:23: Absinthe.Phase.Document.Arguments.Parse.handle_node/2
.....
```

It should returns 400, right?

xingxing:
should be 400

close #275

### DIFF
--- a/lib/absinthe/phase/document/arguments/parse.ex
+++ b/lib/absinthe/phase/document/arguments/parse.ex
@@ -33,10 +33,12 @@ defmodule Absinthe.Phase.Document.Arguments.Parse do
 
   defp build_value(normalized, %Type.Scalar{} = schema_node, context) do
     case Type.Scalar.parse(schema_node, normalized, context) do
-      :error ->
-        {:error, :bad_parse}
       {:ok, val} ->
         {:ok, val}
+      {:error, _reason} ->
+        {:error, :bad_parse}
+      :error ->
+        {:error, :bad_parse}
     end
   end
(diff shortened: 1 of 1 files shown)


benwilson512:
Hey there! I appreciate that you're immediately trying to fix the issue!

However, this is intentional. Parse functions should return `{:ok, value}` or `:error`. Nothing else. If a parse function is returning invalid responses then it will quite reasonably crash the request. It is the same as a resolution function returning an invalid response.

Issue #211, opened 2016-11-18
With multiple interfaces, only the first interface is enforced


sorentwo:
Whether using the `interface` macro, or `interfaces`, only the first interface is considered. For example,

```elixir
interface :node do
  field :id, :id
end

interface :timed do
  field :time, :time
end

object :foo do
  interfaces [:node, :timed]

  field :id, :id
end
```

The unsatisfied interface error for missing `time` won't be raised at all. The results are the same with multiple `interface` calls instead.

I've taken a shot at a PR to fix this, but tracking through the error injection code gets a bit hairy. If you have any pointers I'd like to see the fix through to the end, otherwise I'll push up the test case I've worked through so far.

benwilson512:
Hey thanks for the issue! Just to make sure you're on the right track, this should be implemented as a schema validation rule, not anything to do with blueprints. https://github.com/absinthe-graphql/absinthe/blob/master/lib/absinthe/schema/rule/interfaces_must_resolve_types.ex is an example

bruce:
Notably the schema validation code will change for v1.3, when it will be converted to a phase (since the schema building will become phase-based as well). Just so you're aware; it'll be a month or so before we kick off that effort.

bruce:
Slating this for the v1.3 schema definition changes release.

sorentwo:
@bruce are you suggesting that I leave this alone, or hold off until the schema validation code has been re-worked?

benwilson512:
Fixed on master.

TimFung93:
To be clear this was fixed on the 1.3 release of Absinthe?

benwilson512:
It was fixed in whatever the next release was, you'll want to consult the changelog or tag history to figure out what exact version the fix shows up in.

PR #277, opened 2017-02-26
Don't clear accumulator when checking interfaces

tjfryan:
Fixing a small mistake I noticed.  I think this fixes the bug in #211.


benwilson512:
Thank you!

PR #280, opened 2017-03-05
Add logging, with support for filtering

bruce:
Adds `Absinthe.Logger`

### DIFF
--- a/lib/absinthe/logger.ex
+++ b/lib/absinthe/logger.ex
@@ -1,0 +1,128 @@
+defmodule Absinthe.Logger do
+  @moduledoc """
+  Handles logging of various instrumentation events.
+
+  ## Variable filtering
+
+  Absinthe can filter out sensitive information like tokens and passwords
+  during logging. Variables can be filtered via the `:filter_variables`
+  configuration setting for this module. For example:
+
+      config :absinthe, Absinthe.Logger,
+        filter_variables: ["token", "password", "secret"]
+
+  With the configuration above, Absinthe will filter any variable whose name
+  includes the terms `token`, `password`, or `secret`. The match is case
+  sensitive.
+
+  The default is `["token", "password"]`.
+
+  ## Pipeline display
+
+  Absinthe can optionally display the list of pipeline phases for each processed
+  document when logging. To enable this feature, set the `:pipeline`
+  configuration option for this module:
+
+      config :absinthe, Absinthe.Logger,
+        pipeline: true
+
+  The default is equivalent to `false`.
+  """
+  require Logger
+
+  @doc """
+  Log a document being processed.
+  """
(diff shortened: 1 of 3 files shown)


lib/absinthe/logger.ex
```diff
@@ -0,0 +1,128 @@
...
+
+  @default_filtering ~w(token password)
+
+  @spec variables_to_filter() :: [String.t]
+  defp variables_to_filter do
+    Application.get_env(:absinthe, __MODULE__, [])
+    |> Keyword.get(:filter_variables, @default_filtering)
+  end
+
+  @spec filter_variables(map) :: map
```
jparise: Shouldn't this be `filter_variables(map, [String.t]) :: map`?

lib/absinthe/logger.ex
```diff
@@ -0,0 +1,128 @@
+defmodule Absinthe.Logger do
+  @moduledoc """
+  Handles logging of various instrumentation events.
+
+  ## Variable filtering
+
+  Absinthe can filter out sensitive information like tokens and passwords
```
jparise: Perhaps note that the definition of "filter out" is "replace with the string `[FILTERED]`"?

lib/absinthe/logger.ex
```diff
@@ -0,0 +1,128 @@
...
+  defp pipeline_section(pipeline) do
+    case Application.get_env(:absinthe, __MODULE__, [])[:pipeline] do
+      true ->
+        do_pipeline_section(pipeline)
+      _ ->
+        ?\n
+    end
+  end
+
+  def do_pipeline_section(pipeline) do
```
jparise: `defp`?

test/lib/absinthe/logger_test.exs
```diff
@@ -0,0 +1,71 @@
...
+    @filtered "[FILTERED]"
+
+    test "it filters the set values, with defaults" do
+      assert %{
+        "token" => @filtered,
+        "password" => @filtered,
+        "alsoUnsafe" => @value
+      } = Absinthe.Logger.filter_variables(@variables)
+      # Change the config
+      Application.put_env(:absinthe, Absinthe.Logger, [filter_variables: ~w(token alsoUnsafe)])
```
jparise: I feel like I should know this, but does this line alter global state such that other test scopes would be affected?

bruce: It will, which is annoying. I need to find a better pattern for testing against application configuration.

lib/absinthe/logger.ex
bruce: Sure (this is copied almost verbatim from Phoenix, btw, to include that wording.)

lib/absinthe/logger.ex
bruce: Yeah, probably should be.

bruce:
Feedback addressed in #282. Thanks!

PR #283, opened 2017-03-08
Next

benwilson512:


lib/absinthe/schema.ex
```diff
@@ -145,17 +145,76 @@ defmodule Absinthe.Schema do
...
+            end)
+
+            %{object | fields: fields}
+          type ->
+            type
+        end
+      end
+
+      @doc false
+      def plugins do
```
benwilson512: @bruce one of the last things to decide is what we want to call this. This is the list of middleware that will have before_resolution and after_resolution callbacks run.

bruce: I thought we agreed that a "plugin" was now exactly this -- middleware that requires this type of wiring-in?

benwilson512: Yup ok.

PR #284, opened 2017-03-08
v1.3: Middleware, Logger, Included Custom Types, etc

benwilson512:


bruce:
- [ ] The "Ensure Query Type" rule needs an entry in the CHANGELOG. It's in the spec, but it's a new restriction that may break existing (e.g., mutation-only) schemas.

bruce:
- [ ] Note new Elixir version requirement in CHANGELOG

bruce:
- [ ] README update, to include new features

PR #288, opened 2017-03-09
Add failing test for interface subtype

aughr:
This shows that an implementor of an interface with a more specific field than expected will not compile. I've validated that `graphql-java` accepts this schema (and the documentation that this failure dumps asserts that it should be valid as well).

I don't know how to fix this, but I hope this failing test case will be helpful.


benwilson512:
Gonna merge into a branch to work on it, will update here.

PR #292, opened 2017-03-14
Tuning

benwilson512:


benwilson512:
@bruce I'm gonna merge and push this because it solves some pretty significant performance issues with abstract types, but I'm gonna put some comments on here for further review as well.

test/lib/absinthe/phase/document/flatten_test.exs
```diff
@@ -31,13 +31,9 @@ defmodule Absinthe.Phase.Document.FlattenTest do
       foo(id: $id) {
         gender
         ... FooFields
-        ... NotFooFields
         ... on Foo {
           age
         }
-        ... on NotFoo {
-          age
-        }
```
benwilson512: These had to be removed because it isn't valid to query for not foo fields on foo. However we could add them back if we make `foo` more abstract, and the `notFoo` fields simply additions instead of NOT foo. I don't remember enough about what these are testing.

test/lib/absinthe_test.exs
```diff
@@ -400,8 +400,9 @@ defmodule AbsintheTest do
       assert Enum.sort_by(input_fields, sort) == Enum.sort_by(correct, sort)
     end
 
-    it "ignores fragments that can't be applied" do
-      assert {:ok, %{data: %{"person" => %{"name" => "Bruce"}}}} == run(@unapplied_fragment, ContactSchema)
+    it "Object spreads in object scope should return an error" do
+      # https://facebook.github.io/graphql/#sec-Object-Spreads-In-Object-Scope
+      assert {:ok, %{errors: [%{locations: [%{column: 0, line: 4}], message: "Fragment spread has no type overlap with parent.\nParent possible types: [\"Person\"]\nSpread possible types: [\"Business\"]\n"}]}} == run(@unapplied_fragment, ContactSchema)
```
benwilson512: How do you feel about my error message?

PR #296, opened 2017-03-23
move scalar and enum serialization to the result phase

benwilson512:


lib/absinthe/phase/document/result.ex
```diff
@@ -46,11 +46,20 @@ defmodule Absinthe.Phase.Document.Result do
     %{errors: [format_error(error)]}
   end
 
-  defp data(%{errors: [], value: value}, errors), do: {value, errors}
```
benwilson512: @bruce this was the source of the test failures, it was never falling through to the `defp data` case later on that was theoretically handling leaf nodes. Probably something dialyzer would have caught.

PR #297, opened 2017-03-23
Switched HandleError to HandleAuth

cas27:
If I'm reading the example right I believe you meant the HandleAuth middleware would run after the Auth middleware.


benwilson512:
Hey thanks for the PR! I actually think that the typo is lower down where it should say `middleware HandleError`. We already have an Auth middleware set prior to resolution, the idea with HandleError is that it would let you return ecto changeset errors and so on from the resolution function.

The docs are still a bit of a work in progress.

cas27:
Gotcha.  Updated the middleware to HandleError.

PR #299, opened 2017-03-24
Add fix for merging fragments on a list field returning more than 1 element

howleysv:
The current implementation for fragment merging in the resolution step only matches on `[%{} = left], [%{} = right]` when lists of objects are found.

The PR correctly handles merging when multiple fragments are made over a list of objects field.

### DIFF
--- a/lib/absinthe/phase/document/result.ex
+++ b/lib/absinthe/phase/document/result.ex
@@ -81,11 +81,12 @@ defmodule Absinthe.Phase.Document.Result do
     Map.merge(left, right, &do_deep_resolve/3)
   end
 
-  defp do_deep_resolve(_key, [%{} = left], [%{} = right]) do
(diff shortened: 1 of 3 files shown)


benwilson512:
Hey @howleysv I really appreciate the PR. However at this point https://github.com/absinthe-graphql/absinthe/pull/309 is able to properly merge the fragment fields ahead of time, meaning we no longer need to merge in results at all.

The only failing test case on that branch at the moment is from the complexity analysis, which needs to be adjusted a bit. This includes the tests from this branch which I have copied in.

howleysv:
@benwilson512 Oh cool, it sounds like #309 will be a nice improvement, great job!

Hope the test cases were at least useful...

benwilson512:
They have been incredibly useful, your persistence here has been essential, thank you!

PR #302, opened 2017-03-28
Batch resolution phase

benwilson512:


benwilson512:
@skosch I'm moving the batch resolution ability to Absinthe itself as none of its concerns are really Plug specific, and it seems I'm likely to need them in Absinthe itself anyway.

As I'm doing this though it seems clear that would I would REALLY like is essentially

```
Absinthe.Pipeline.Batch.run(prep_phases, resolution_phases, result_phases, blueprints)
```

This became particularly clear when trying to write a test for just the resolution logic I had so far. There was easy way to get a set of blueprints that were up to the correct point.

The overall logic here is non trivial because errors in the prep phases needs to require jumping over the resolution phases for that doc and moving to the result phases. Fortunately we've already done 98% of this in the Absinthe.Plug PR, but I'm wondering if it really belongs here.

@bruce I think we could make the API even simpler to just:
```
Absinthe.Pipeline.Batch.run(phases, blueprints)
```

With the introduction of pipeline tags that we've talked about. @skosch pipeline tags are an idea where in a pipeline you could have:
```
[
  {:section, :prep}
  SomePhase,
  SomeOtherPhase,
  {:section, :resolution}
  ResolutionPhase
  {:section, :result}
  ResultPhase
]
```

Thoughts?

lib/absinthe/phase/document/execution/before_resolution.ex
```diff
@@ -1,17 +0,0 @@
...
-
-  @spec run(Blueprint.t, Keyword.t) :: Phase.result_t
-  def run(%{resolution: resolution} = bp_root, _opts \\ []) do
-    acc = Enum.reduce(bp_root.schema.plugins(), resolution.acc, fn plugin, acc ->
-      plugin.before_resolution(acc)
-    end)
-
-    {:ok, %{bp_root | resolution: %{resolution | acc: acc}}}
-  end
-end
```
benwilson512: @bruce note I've decided to go back on my `Before`/`After` resolution phase stuff. It kept producing a lot boilerplate everywhere. Instead i've simply got an option you can pass to the `Resolution` phase that turns on / off running the callbacks.

skosch:
@benwilson512 thank you for looking into this!

I obviously don't know the Absinthe codebase well enough to comment on how to minimize the pain of passing phases and results around. Do you see an advantage in the pipeline tags idea over, say, having an alternate `run` definition?
```elixir
def run(%{prep, resolution, result}, blueprints) do ...
def run(phases, blueprints) do ...
```

tlvenn:
Any chance to merge this soon if it's ready to get us started with transport batching ?

benwilson512:
So in the end most of this is going to go inside of Absinthe.Plug itself. There's a number of things that are a bit odd about handling document batches in Absinthe at the moment, so I think for 1.3 it all goes in Absinthe.Plug and we extract it to Absinthe itself for 1.4 when we look at subscriptions.

PR #306, opened 2017-04-10
Extensions

benwilson512:


test/lib/absinthe/extensions_test.exs
```diff
@@ -0,0 +1,47 @@
+defmodule Absinthe.ExtensionsTest do
+  use Absinthe.Case, async: true
```
benwilson512: @fishcakez this module serves as a very minimal sketch of what setting extensions looks like.

test/lib/absinthe/extensions_test.exs
```diff
@@ -0,0 +1,47 @@
...
+      field :foo, :string do
+        middleware :resolve_foo
+      end
+    end
+
+    def resolve_foo(res, _opts) do
+      %{res |
+        value: "hello world",
+        state: :resolved,
+        extensions: %{foo: 1},
```
benwilson512: arbitrary values can be placed here by middleware. All middleware share a single map, so it may be wise to namespace it by the name of the middleware.

test/lib/absinthe/extensions_test.exs
```diff
@@ -0,0 +1,47 @@
...
+      }
+    end
+  end
+
+  defmodule MyPhase do
+    # rolls up the extensions data into a top level result
+    def run(blueprint, _) do
+      extensions = get_ext(blueprint.resolution.result.fields)
+      result = Map.put(blueprint.result, :extensions, extensions)
```
benwilson512: The extension data is not included in the result by default because it is not possible to know how to roll that data up. Thus a phase must be provided to do so.

PR #309, opened 2017-04-13
Proper Fragment Projection

benwilson512:
Note: there are test failures, but they're all related to complexity analysis at this point which needs to be more nuanced.


benwilson512:
@howleysv hey btw are you able to try out this branch and see if it works with the scenarios you were running into previously?

howleysv:
I can try run our integration test suite against it - we have a few custom things in our 1.2 fork that we need to port to 1.3 but I wouldn't expect too many failures related to them being absent...

I'll try look at it over the weekend, but next week will mostly be spent at Elixirconf :sunglasses:

benwilson512:
Interesting, what did you fork for 1.2?

howleysv:
Right now there's not a lot, we've been slowly removing our temporary features as they become possible in absinthe core and I think when this is merged we can probably move toward abandoning our fork.

We're currently using our fork to handle:
- correct merging of list of fragments (solved by this)
- updating context inside mutation resolvers (solvable with 1.3 middleware)
- some custom logging around diagnosing batch timeouts

howleysv:
@benwilson512 looks good 👍
It seems to correctly resolve the insanely fragmented queries generated by our Relay (Classic) client

benwilson512:
Excellent, glad to hear!

PR #313, opened 2017-04-27
Topsort Fragments

benwilson512:
The idea here is that if you walk through `blueprint.fragments` in order, each fragment will only depend on fragments found prior in the list. This is particularly useful for complexity analysis, because you can simply go through the list summing complexity and if you find a fragment spread, you're guaranteed to have already seen it and computed its complexity.

### DIFF
--- a/lib/absinthe/phase/document/validation/no_fragment_cycles.ex
+++ b/lib/absinthe/phase/document/validation/no_fragment_cycles.ex
@@ -22,7 +22,7 @@ defmodule Absinthe.Phase.Document.Validation.NoFragmentCycles do
   @spec do_run(Blueprint.t, %{validation_result_phase: Phase.t}) :: Phase.result_t
   def do_run(input, %{validation_result_phase: abort_phase}) do
     {fragments, error_count} = check(input.fragments)
-    result = put_in(input.fragments, fragments)
+    result = %{input | fragments: fragments}
     if error_count > 0 do
       {:jump, result, abort_phase}
     else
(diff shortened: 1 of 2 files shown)


lib/absinthe/phase/document/validation/no_fragment_cycles.ex
```diff
@@ -35,7 +35,17 @@ defmodule Absinthe.Phase.Document.Validation.NoFragmentCycles do
   defp check(fragments) do
     graph = :digraph.new([:cyclic])
     try do
-      check(fragments, graph)
+      with {fragments, 0} <- check(fragments, graph) do
+        fragments = Map.new(fragments, &{&1.name, &1})
+
+        fragments =
+          graph
+          |> :digraph_utils.topsort
```
bruce: Nice!

bruce: We already create a digraph for fragment cycles validation. Should we keep around ets refs in the blueprint for reuse?

benwilson512: This is in that phase, `topsort/1` is just a function that operates on the graph we already created to return a list of verticies.

I've considered keeping the `:digraph` table around for the rest of the request but it complicates resource management a bit. More to the point, once you've sorted the fragments list there's not much you need the graph for, since you've got a handy sorted tree list on hand already that doesn't require any fancy pointer / vertex tracking.

PR #318, opened 2017-05-06
Late Binding Variables

benwilson512:
Right now variables are inlined pretty early in the main execution pipeline. This significantly limits how much work can be re-used across different executions of the same document.

The goal of this PR is to bind variables as late as possible. This will enable:
- Document wide validation without requiring any variables.
- Significant increase in caching / prepared query wins.

Status: Nearly all tests involving valid variables / arguments pass, nearly all tests that return errors about invalid variables are still failing.

### DIFF
--- a/lib/absinthe/phase/document/variables.ex
+++ b/lib/absinthe/phase/document/variables.ex
@@ -41,36 +41,39 @@ defmodule Absinthe.Phase.Document.Variables do
   # Note that no validation occurs in this phase.
 
   use Absinthe.Phase
-  alias Absinthe.Blueprint
+  alias Absinthe.{Blueprint, Pipeline}
 
   @spec run(Blueprint.t, Keyword.t) :: {:ok, Blueprint.t}
-  def run(input, options \\ []) do
+  def run(blueprint, options \\ []) do
     variables = options[:variables] || %{}
-    {:ok, update_operations(input, variables) |> IO.inspect}
-  end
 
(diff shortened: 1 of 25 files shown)


lib/absinthe/phase/document/variables.ex
```diff
@@ -41,36 +41,39 @@ defmodule Absinthe.Phase.Document.Variables do
...
-      variable_definitions: variable_definitions,
-      provided_values: provided_values
-    }
+    %{operation | variable_definitions: variable_definitions}
+  end
+
+  def validate_variable(variable_def, root, options) do
+    pipeline = Pipeline.for_variables(root.schema, options)
+
+    {:ok, variable_def, _} = Pipeline.run(variable_def, pipeline)
```
benwilson512: Running a separate pipeline within this phase is not how I want to do this. The motivation for doing so is that I can 100% re-use existing phases, but I want them to only apply to the variables portion of the tree.

My plan is to do something like

```
Phase.Parse
# ... validation stuff
Phase.Schema
# ... no variables yet
Phase.Validation.Result
{Phase.Schema, scope: :variables}
{Phase.Validation.Result, scope: :variables}
```

The idea with `scope` is that I could focus what the phase operates on. This MIGHT even be something worth extending to the transform walker itself, but we'll see.

Ultimately we may just need some custom variable phases.

benwilson512:
Yeah, killing this for now.

Issue #156, opened 2016-09-13
GraphQL Subscriptions


brentjanderson:
Using some pub sub mechanism (potentially built with GenStage?) it would be amazing to have Subscriptions that publish back to clients after a mutation, enabling realtime functionality. See https://medium.com/apollo-stack/graphql-subscriptions-in-apollo-client-9a2457f015fb for the Apollo stack approach to realtime GraphQL. If combined with web sockets/Phoenix channels, this would be a killer stack.

brentjanderson:
Refs #87 - perhaps there is enough direction and spec in place to support this now?

dpehrson:
After reading that article, this seems like it would be very cool.

benwilson512:
We are absolutely in the process of working on this. This is more of a roadmap issue than an immediate feature request. It is still unclear precisely the best way to implement this, and even the clientside and javascript reference implementation is in a lot of flux. The discussion is ongoing, and we're excited to see what will happen here.

dpehrson:
@benwilson512 It may be worth following what GitHub does with this. In their [engineering blog post about launching a GraphQL endpoing=t](http://githubengineering.com/the-github-graphql-api/) they specifically mentioned that a subscription model was one of their hopes/requirements.

Maybe they'll help move the ball forward for us.

bruce:
Well, to be clear by "we're excited to see what will happen here," we're not going to _wait_ to see what happens, or who will drive it forward. Phoenix (and, to speak more widely, OTP) is a particularly good fit for subscriptions (I'd argue it's a much better fit than, ie, Node.js), and we've been busy laying some groundwork to really start building this.

churcho:
Any traction on this or an update on the roadmap perhaps?

bruce:
@benwilson512 is currently digging into this, given we have what we need to build on top of in v1.2. Expect a report on progress by the end of this next week.

stemcc:
>  It is still unclear precisely the best way to implement this, and even the clientside and javascript reference implementation is in a lot of flux.

@benwilson512 As you consider this, it might be a good idea to keep in mind popular graphql clients that already ship with subscriptions support, like [Apollo ](https://github.com/apollostack/apollo-client), for potential support out-of-the-box. Relatively frictionless interop with subscriptions would be wonderful.

tlvenn:
@stemcc that should happen naturally given they are using apollo client as a GraphQL client already. So I am pretty sure the subscriptions will target apollo client.

[...]

PR #322, opened 2017-05-11
Subscriptions

benwilson512:
- [x] Local node subscriptions
- [x] Distributed subscriptions
- [x] Local broadcast optimizations
- [ ] testing


lib/absinthe/schema/notation.ex
```diff
@@ -33,6 +33,18 @@ defmodule Absinthe.Schema.Notation do
...
   end
 
+  defmacro topic(topic_fun) do
+    __CALLER__.module
+    |> Scope.put_attribute(:topic, topic_fun)
+    :ok
+  end
+
+  defmacro trigger(mutations, attrs) do
+    env = __CALLER__
```
jparise: This is super minor, but these two new macros use two different syntax approaches to get the module argument into `Scope.put_attribute`. Would look nicer if they were the same.

benwilson512: Yup. You may want to wait on reviewing until it's closer to being done, this is very WIP, there will undoubtedly by non idiomatic code from time to time until it's closer to completion.

bruce:
Just a note (so it's linked), this at least in part addresses #156.

PR #323, opened 2017-05-12
Added test for batch failing to query sub-fields correctly

howleysv:
I've encountered an issue where batching doesn't play nice with, I suspect, the new fragment merging behaviour on absinthe master. When you batch fields from different points in the schema, which may have different data requirements, they all use the data requirements from one arbitrarily chosen field from the batch.

This test exposes the incorrect behaviour


benwilson512:
It does seem related to the field merging changes, as the test passes without issue on `v1.3.0`. Investigating.

benwilson512:
Specifically, it's a problem with the projection cache. If you remove caching from the projector, it passes appropriately. The cache key is clearly missing some important piece of information here to distinguish where we are in the document appropriately.

On the plus side, the fragment merging logic itself seems OK still.

PR #325, opened 2017-05-18
Expose Meta fields to Introspection Query

michaelkschmidt:
Fields can have extra meta data attached to them:

```
query do
  field :my_field, :string do
    meta :default, "my default value"
  end
end
```
Currently there is no way to access this from the client.  This PR allows that information to be accessed via the Introspection Query
```
query {
      __type(name: "RootQueryType") {
        fields {
          name
          __default
        }
      }
    }
```

Since the meta data can by any type, there is a helper macro for defining the correct hook:
```
MetaUtil.expose_meta(:default, :string, "Field Default Value")
```

### DIFF
--- a/lib/absinthe/introspection/field.ex
+++ b/lib/absinthe/introspection/field.ex
@@ -67,4 +67,10 @@ defmodule Absinthe.Introspection.Field do
     }
   end
 
+  def meta(custom_meta) do
+    module = "Elixir.Absinthe.Introspection.Field.Meta.#{String.capitalize(custom_meta)}"
+              |> String.to_existing_atom()
+    apply(module,:meta,[])
+  end
+
 end
--- a/lib/absinthe/introspection/field/meta_util.ex
+++ b/lib/absinthe/introspection/field/meta_util.ex
@@ -1,0 +1,82 @@
+defmodule Absinthe.Introspection.Field.MetaUtil do
+  @moduledoc """
+  Helper macros for exposing custom meta data
+
+  example:
+    defmodule MyMetaMod do
+      require Absinthe.Introspection.Field.MetaUtil
+      alias Absinthe.Introspection.Field.MetaUtil
+      
+      MetaUtil.expose_meta(:default, :string, "Field Default Value")
+    end
+
+    defmodule TestSchema do
+      use Absinthe.Schema
+      use Absinthe.Schema.Notation
+
+      query do
+        field :my_field, :string do
+          meta :default, "my default value"
+        end
+      end
+    end
+
+    \"""
+    query {
+      __type(name: "RootQueryType") {
+        fields {
+          name
+          __default
+        }
+      }
+    }
+    \"""
+    |> Absinthe.run!(TestSchema)
+
+  output:
+    %{data: %{"__type" => 
+              %{"fields" => [
+                  %{"__default" => "my default value",
+                    "name" => "myField"}]}}}
+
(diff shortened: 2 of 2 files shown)


bruce:
`__*` prefixed fields are reserved according to the GraphQL specification. Sadly extension of introspection isn't particularly easy, but we need a different approach.

michaelkschmidt:
How would you prefer to expose it?  The __* was the most non-invasive hook I could find.  I don't mind doing a more complex implementation if you know how you want the query to look

bruce:
Yeah, I'm not denying `__` would be nice to use (we just can't, to stick to spec).

Ultimately I think, for now, this should be implemented _outside_ of the introspection types; there is [still a lot of discussion around how/when/exactly how introspection types should be extended](https://github.com/facebook/graphql/issues/300), and it's possible custom extensions could break clients with static client-side introspection types defined (eg, Elm).

I don't think it's a good idea for Absinthe to make an early decision and ship with meta introspection code at this point. Things may change with v1.4; a lot of schema definition code will be refactored in that release and the IDL will be fully supported, and hopefully we'll have a clearer direction on introspection extension.

If you do create a package, however, that adds the functionality (eg, as a module that can be pulled in with `import_types` and `import_fields`), we'd certainly link it in a future guide.

michaelkschmidt:
@bruce can you clarify how using `import_types` and `import_fields` would work in this case?  The whole point of the PR is to add a hook somehow into the introspection query.

bruce:
@michaelkschmidt What I mean is, instead of trying to hook into introspection types, just make a normal object who fields can be imported.

For example:

```elixir
defmodule MyApp.Web.Schema.Meta do
  use Absinthe.Schema.Notation

  object :meta_queries do

    @desc "Get a single meta value, for a type or field."
    field :meta, :string do
      arg :name, non_null(:string)
      arg :type, non_null(:string)
      arg :field, :string
      resolve # ... use Absinthe.Schema.lookup_type/2 with the schema, find data with get_in/2, etc.
    end
  end

   # You could also return all meta for an object, for all its fields, etc

end
```

Then, in the schema:

```elixir
defmodule MyApp.Web.Schema do
  use Absinthe.Schema

  import_types MyApp.Web.Schema.Meta
  
  query do
    import_fields :meta_queries
    # ... other stuff
  end

end
```

Then, from GraphQL:

```graphql
{
  myFieldDefault: meta(type: "RootQueryType", field: "myField", name: "default")
}
```

michaelkschmidt:
@bruce that makes sense.  Thanks for the full example.

One more question:  Can custom directives modify the value in Absinthe?  I've seen in other places people suggest exposing the meta data via the "description" field.  If I could write a directive to expose this, it might solve my problem

bruce:
Directives that are defined with an `expand` function can modify the tree (the Absinthe.Blueprint tree, that is), so can change the result however they like. Some places to look:
- The [Absinthe.Phase.Document.Directive](https://github.com/absinthe-graphql/absinthe/blob/master/lib/absinthe/phase/document/directives.ex#L22) phase, when the expansion occurs.
- The [expand](https://github.com/absinthe-graphql/absinthe/blob/master/lib/absinthe/type/directive.ex#L30) function to be set. (It occurs to me that some directive documentation needs to be updated, btw, as it's not documented).

michaelkschmidt:
@bruce Sweet!! Thanks again for all the help

PR #330, opened 2017-05-29
Phase.Document.Validation.SelectedCurrentOperation edge case

vderyagin:
I added a test with a query that looks like it should fail the validation, but it passes. Not sure if it's a bug or somehow intended behavior, so I did not attempt to fix it until clarified.


benwilson512:
Good catch, thank you!

PR #334, opened 2017-06-06
Improve generic example in docs of batch_fun

jonathanlemuel:
This brings the generic example of using `batch_fun` in line with the Ecto Best Practices guide. It had a typo and was missing the map step.

:)


benwilson512:
Thank you!

PR #356, opened 2017-08-02
Add No Introspection Validation

trinode:
Hello,

This is not ready to be merged (expecting to add tweak, tests, reformat etc), but for discussion.

I've a use case whereby disabling GraphQL introspection is desirable similar to this validation rule for the JS GraphQL implementation:-
https://github.com/helfer/graphql-disable-introspection

Currently you can choose to include this validation into your pipeline and it will reject __schema and __type by default, but this is overridable.

### DIFF
--- a/lib/absinthe/phase/document/validation/no_introspection.ex
+++ b/lib/absinthe/phase/document/validation/no_introspection.ex
@@ -1,0 +1,50 @@
+defmodule Absinthe.Phase.Document.Validation.NoIntrospection do
+  @moduledoc false
+
+  # Ensure that document doesn't have any of the specified types
+  # defaulting to __schema and __type
+  #
+  
+  use Absinthe.Phase
+  use Absinthe.Phase.Validation
+
+  @default_types [:__schema, :__type]
+  
+  @doc """
+  Run the validation.
+  """
+  @spec run(Blueprint.t, Keyword.t) :: Phase.result_t
+  def run(input, opts \\ []) do
+    excluded_types = Keyword.get(opts, :exclude, @default_types)
+
+    result = Blueprint.prewalk(input, &handle_node(&1, input.schema, excluded_types))
+    {:ok, result}
+  end
+
+  # Check a node for introspection types
+  defp handle_node(%{schema_node: nil} = node, _schema, _), do: {:halt, node}
+  defp handle_node(%{schema_node: %{type: type}} = node, _, excluded_types) do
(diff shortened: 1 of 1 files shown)


bruce:
While I think think there's a compelling use case for this, in my opinion (as it is for the JS implementation), this probably belongs in a separate package.

Now, if there are changes that need to be made to Absinthe itself because of missing mechanisms, to make integration easier, etc, those I can see being made here as changes.

trinode:
I guess my original thought was that introspection was a feature of the server, and I was expecting it to be possible to disable via configuration, so this is the closest I could get without delving too deep.

I think something like this that looks at internal structures (discussed with @benwilson512 on slack re Blueprint being considered kinda internal) could be quite fragile if outside the core app, unless I am really strict with version pinning, then that's painful.

This is, in my use case, to prevent leaking details of upcoming features which are enabled in non-public beta apps, some might consider it a security feature, if Blueprint changed, this would fall through to the default case and once again allow introspection even when enabled (which would be an unpleasant surprise for someone wanting to protect their schema details).

Basically I like the idea of tying this to the same test suite as the absinthe core app so that compatibility is guaranteed without strict version pinning of dependencies or panic updates when a breaking change occurs.

bruce:
@trinode Absinthe's intention is to provide GraphQL specification-compatible features; while I think thing there is a compelling use case for this for some, it sits squarely in the separate package bucket precisely because it's such a non-standard approach. (I've discussed the issue with @benwilson512 in detail and we're in agreement.)

As for Blueprint being internal, sure, it is: but version locking Absinthe from a separate package is still possible, and then a test suite ensuring regression doesn't occur can live with the separately well-documented, well-supported package -- vs something that's maintained by the core team. This isn't the first feature we've urged be built as a separate dependency (we've done a fair number ourselves), and its certainly not a judgement about the value of the feature to you or an unwillingness to work with you as changes occur to Absinthe. I would just rather we grow the Absinthe ecosystem of packages vs expand the core package by default -- we'll certainly work with package authors to keep their features working.

For what it's worth, BTW: There are no plans at the moment to change the Blueprint trees for GraphQL documents -- what's in the works is additions to Blueprint to support schema generation from GraphQL schema definitions, which shouldn't create issues with this usage.

trinode:
OK, feel free to close this PR, I'll rely on version pinning for now.

bruce:
@trinode Sounds good; if you create a package for this, let us know!

PR #359, opened 2017-08-07
Improve documentation for additional data types that ship with absinthe

johannesE:


lib/absinthe/type/custom.ex
```diff
@@ -2,7 +2,13 @@ defmodule Absinthe.Type.Custom do
   use Absinthe.Schema.Notation
 
   @moduledoc """
-  This module contains additional data types.
+  This module contains the following additional data types:
+  - datetime
```
benwilson512: Can you change this to `- datetime (UTC)`

johannesE: Of course, done. :) I hope you don't mind the force push. This way the history looks cleaner.

benwilson512:
Totally fine, thank you!

PR #362, opened 2017-08-12
Add support for Decimal values

lauraannwilliams:
As requested, this PR adds :decimal as a custom type in Absinthe.

Decimal values are return in JSON as strings to maintain precision, and are parsed into Decimal structs using the [Decimal](https://hexdocs.pm/decimal/Decimal.html) library

### DIFF
--- a/lib/absinthe/type/custom.ex
+++ b/lib/absinthe/type/custom.ex
@@ -7,7 +7,7 @@ defmodule Absinthe.Type.Custom do
   - naive_datetime
   - date
   - time
-  - decimal (only if Decimal is available)
+  - decimal (only if [Decimal](https://hex.pm/packages/decimal) is available)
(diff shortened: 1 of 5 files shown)


lib/absinthe/type/custom.ex
```diff
@@ -7,6 +7,7 @@ defmodule Absinthe.Type.Custom do
   - naive_datetime
   - date
   - time
+  - decimal (only if Decimal is available)
```
jparise: Would be nice to link to the Decimal package here. For example:

```
- decimal (only if [Decimal](https://hex.pm/packages/decimal) is available)
```

lauraannwilliams:
Tweak made :)

benwilson512:
Thanks!

PR #372, opened 2017-08-21
Use phase_config_t type instead of Phase.t for insert_{before,after}

vderyagin:
Tuples in form `{SomePhase, option: value}` can be inserted into pipeline using `insert_{before,after}` functions, which is not reflected in their `@spec`s.


benwilson512:
Thanks!

PR #380, opened 2017-08-26
Subscription Topic Improvements

### LINKED ISSUE #365: Subscription, pointers how to do authorization?
I'm trying to rewrite some of our custom websocket code to subscription. I had success in putting in the context. But I am running in an issue, when I would like to do authorization on who is able to subscribe to what. So not anyone can subscribe to every `content_item_id`. 

I tried adding a resolve block (but that doesn't seem to work). Is there another way to do authorization?

```elixir
  subscription do
    field :on_message_added, :message do
      arg :content_item_id, non_null(:id)

      topic fn args ->
        args.content_item_id
      end

      trigger :send_message, topic: fn message ->
        message.content_item_id
      end
    end
  end
```

benwilson512:
Fixes #365

### DIFF
--- a/lib/absinthe/schema/notation.ex
+++ b/lib/absinthe/schema/notation.ex
@@ -56,7 +56,7 @@ defmodule Absinthe.Schema.Notation do
   ```elixir
   config fn args, %{context: context} ->
     if authorized?(context) do
-      {:ok, args.client_id}
+      {:ok, topic: args.client_id}
     else
       {:error, "unauthorized"}
     end
--- a/lib/absinthe/schema/notation.ex
+++ b/lib/absinthe/schema/notation.ex
@@ -34,12 +34,44 @@ defmodule Absinthe.Schema.Notation do
   end
 
   @doc """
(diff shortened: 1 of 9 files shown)


benwilson512:
- [ ] more error formatting.

lib/absinthe/schema/notation.ex
```diff
@@ -34,15 +34,45 @@ defmodule Absinthe.Schema.Notation do
...
+  ## Example
+
+  ```elixir
+  config fn args, %{context: context} ->
+    if authorized?(context) do
+      {:ok, topic: args.client_id}
+    else
+      {:error, "unauthorized"}
+    end
+  end
```
benwilson512: @freshtonic ^

Note that we're doing away with the `topic` macro entirely in favor of this `config` macro. It gives us the possibility of setting up more than just the topic, which is valuable for a variety of other features we're looking at.

jfrolich: I like this!

PR #382, opened 2017-08-31
Allow parsing Input.Null for custom types

nicolas-dular:
`Absinthe.Blueprint.Input.Null` got introduced in https://github.com/absinthe-graphql/absinthe/pull/198 but parsing fails for
custom types, because they didn't handle it.

This PR fixes this by matching for `Absinthe.Blueprint.Input.Null` as well.

Are there any other types that would need to handle this case? If so, I can update my PR.

### DIFF
--- a/lib/absinthe/type/custom.ex
+++ b/lib/absinthe/type/custom.ex
@@ -68,7 +68,7 @@ defmodule Absinthe.Type.Custom do
       parse &parse_decimal/1
     end
 
-    @spec parse_decimal(any) :: {:ok, Decimal.t} | :error
+    @spec parse_decimal(any) :: {:ok, Decimal.t} | {:ok, nil} | :error
     defp parse_decimal(%Absinthe.Blueprint.Input.String{value: value}) do
       case Decimal.parse(value) do
         {:ok, decimal} -> {:ok, decimal}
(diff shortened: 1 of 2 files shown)


lib/absinthe/type/custom.ex
```diff
@@ -83,52 +83,67 @@ defmodule Absinthe.Type.Custom do
...
+    defp parse_decimal(%Absinthe.Blueprint.Input.Null{}) do
+      {:ok, nil}
+    end
     defp parse_decimal(_) do
       :error
     end
   end
 
-  @spec parse_datetime(Absinthe.Blueprint.Input.String.t) :: {:ok, DateTime.t} | :error
+  @spec parse_datetime(Absinthe.Blueprint.Input.String.t) :: {:ok, DateTime.t} | {:ok, nil} | :error
```
bruce: For example, change this to:

```elixir
@spec parse_datetime(Absinthe.Blueprint.Input.String.t) :: {:ok, DateTime.t} | :error
@spec parse_datetime(Absinthe.Blueprint.Input.Null.t) :: {:ok, nil}
```

nicolas-dular:
@bruce thanks, fixed it - haven't used `@spec` definitions in elixir before. Hope it's correct!

bruce:
Danke, Nicolas!

PR #386, opened 2017-09-05
Batch Subscription Resolution

benwilson512:
TODO
- [x] exception handling inside resolvers
- [x] exception handling inside result building


benwilson512:
🎉

PR #388, opened 2017-09-13
Fix `no function clause matching in sum_complexity` error

xingxing:
By default each field in a query will increase the complexity by 1

### DIFF
--- a/lib/absinthe/phase/document/complexity/analysis.ex
+++ b/lib/absinthe/phase/document/complexity/analysis.ex
@@ -119,6 +119,10 @@ defmodule Absinthe.Phase.Document.Complexity.Analysis do
     Enum.reduce(fields, 0, &sum_complexity/2)
   end
 
+  defp sum_complexity(%{complexity: complexity}, acc) when is_nil(complexity) do
+    @default_complexity + acc
+  end
+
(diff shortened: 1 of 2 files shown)


lib/absinthe/phase/document/complexity/analysis.ex
```diff
@@ -119,6 +119,10 @@ defmodule Absinthe.Phase.Document.Complexity.Analysis do
     Enum.reduce(fields, 0, &sum_complexity/2)
   end
 
+  defp sum_complexity(%{complexity: complexity}, acc) when is_nil(complexity) do
+    1 + acc
```
benwilson512: Can you change this to `@default_complexity + acc`

xingxing:
I addressed your comment, place have a look when you are free

benwilson512:
Awesome. Is there a test case you could add to the complexity test file that shows this code at work?

xingxing:
Sure will do that ASAP

xingxing:
It seems the issue will happen when using union object. I add the test case.

xingxing:
Hey @benwilson512 is that good to go?

benwilson512:
Thanks!

PR #390, opened 2017-09-14
Failing test for complexity analysis of GraphiQL introspection query

vderyagin:
[GraphiQL introspection query](https://github.com/graphql/graphiql/blob/dc4a561f796258c6e6773612279abccb76f10ee0/src/utility/introspectionQueries.js#L14) triggers error in Absinthe complexity analysis phase. That query is quite big, I cut some stuff out of it, and ended up with smaller one that still triggers the same error. Both queries run fine and return expected results with complexity analysis disabled.


benwilson512:
Handled in master, thanks!

PR #395, opened 2017-09-23
Fix doc for absinthe.schema.json task

nitingupta910:
Filename is expected as the first argument. Fix doc to reflect the same.


benwilson512:
Thanks!

Issue #393, opened 2017-09-19
Subscription name collision


binaryseed:
Hi there, just playing around with subscriptions, and ran into a bug.. 

We have an object named `subscription` in our schema, and we get a strange error in `1.4.0-rc.0`

Changing the object name clears it up, so there's some kind of name collision happening!

```
** (exit) an exception was raised:
    ** (Protocol.UndefinedError) protocol String.Chars not implemented for %{account_id: 1, ...}. This protocol is implemented for: Atom, BitString, Date, DateTime, Float, Integer, List, NaiveDateTime, Time, URI, Version, Version.Requirement
        (elixir) /private/tmp/elixir-20170801-32483-1rf8an1/elixir-1.5.1/lib/elixir/lib/string/chars.ex:3: String.Chars.impl_for!/1
        (elixir) /private/tmp/elixir-20170801-32483-1rf8an1/elixir-1.5.1/lib/elixir/lib/string/chars.ex:22: String.Chars.to_string/1
        (absinthe) lib/absinthe/phase/document/result.ex:53: Absinthe.Phase.Document.Result.data/2
        (absinthe) lib/absinthe/phase/document/result.ex:79: Absinthe.Phase.Document.Result.field_data/3
        (absinthe) lib/absinthe/phase/document/result.ex:69: Absinthe.Phase.Document.Result.list_data/3
        (absinthe) lib/absinthe/phase/document/result.ex:79: Absinthe.Phase.Document.Result.field_data/3
        (absinthe) lib/absinthe/phase/document/result.ex:20: Absinthe.Phase.Document.Result.process/1
        (absinthe) lib/absinthe/phase/document/result.ex:12: Absinthe.Phase.Document.Result.run/2
        (absinthe) lib/absinthe/pipeline.ex:249: Absinthe.Pipeline.run_phase/3
        (absinthe_plug) lib/absinthe/plug.ex:291: Absinthe.Plug.run_query/3
        (absinthe_plug) lib/absinthe/plug.ex:219: Absinthe.Plug.execute/2
        (absinthe_plug) lib/absinthe/plug.ex:155: Absinthe.Plug.call/2
```

Thanks!

benwilson512:
@binaryseed `:subscription` is a reserved object name, unfortunately if you have an object by that name you'll need to change it.

binaryseed:
I made an attempt to raise a better error, but I can't properly differentiate the case of `object :subscription` and `subscription do`...

[...]

PR #396, opened 2017-09-24
Don't allow Reserved names as object identifiers

binaryseed:
Re: https://github.com/absinthe-graphql/absinthe/issues/393

This PR raises an error when someone tries to define an object using a reserved `:identifier`.. (`:query`, `:mutation`, `:subscription`).

This currently leads to a fairly opaque error at runtime :(

### DIFF
--- a/lib/absinthe/schema/notation.ex
+++ b/lib/absinthe/schema/notation.ex
@@ -176,7 +176,12 @@ defmodule Absinthe.Schema.Notation do
   end
   ```
   """
-  defmacro object(identifier, attrs \\ [], [do: block]) do
+  @reserved_identifiers ~w(query mutation subscription)a
+  defmacro object(identifier, attrs \\ [], block)
+  defmacro object(identifier, _attrs, _block) when identifier in @reserved_identifiers do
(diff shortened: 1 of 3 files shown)


benwilson512:
I'm fine doing a raise

bruce:
Agree; we could do this with a schema rule, but it's hard to differentiate, and doesn't add much. (Plus schema rules are more about adherence to spec, not using the wrong macro.)

binaryseed:
Yeah, I can take another poke at it this afternoon..

binaryseed:
FYI, I think this one is ready when you are :)

benwilson512:
Thank you!

greenboxal:
Is this a reserved name in GraphQL spec? Or just an absinthe limitation? Not being able to create a `:subscription` object is terrible for me.

bruce:
`:subscription` is reserved by Absinthe as an _identifier_, but you should still be able to use `"Subscription"` as the canonical GraphQL object type name, eg:

```elixir
object :product_subscription, name: "Subscription" do
  # ...
end

object :user do
  field :subscription, :product_subscription
end
```

PR #400, opened 2017-09-27
don't say in field foo for resolution errors

benwilson512:


lib/absinthe/phase/document/execution/resolution.ex
```diff
@@ -249,7 +249,6 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
       {[], _} ->
         raise Absinthe.Resolution.result_error(original_value, bp_field, source)
       {[message: message], extra} ->
-        message = ~s(In field "#{bp_field.name}": #{message})
```
benwilson512: @bruce one line change in the end.

benwilson512:
@giulianok

PR #402, opened 2017-09-28
always stringify topics

### LINKED ISSUE #401: Manual publish does not work w/ ecto structs
The example shown here:
https://hexdocs.pm/absinthe/1.4.0-rc.2/Absinthe.Schema.html#subscription/2
(specifically the line `Absinthe.Subscription.publish(pubsub, user, [new_users: user.account_id])`)

doesn't actually work since there is a type mismatch between the subscribed topic and the published one.

Basically, in the config function, `args.id` is a string since the absinthe `ID` type comes in as a string, but since ecto structs get typecasted to numbers and so the `publish/3` call ends up publishing to topics clients are not listening to.

_WORKAROUND_: To get this to work you can either turn your topic key in your publish call into a string to match the topic subscribed like: `Absinthe.Subscription.publish(pubsub, user, [new_users: to_string(user.account_id)]) # Note the to_string at the end here` **OR** convert the topic in the config function to an integer.

Since all phoenix topics are strings, I don't think restricting the topic to strings only is of any major limitation as users will probably expect this is happening under the hood (as I did). Either the docs should be updated to note this, or that key should be converted here: https://github.com/absinthe-graphql/absinthe/blob/master/lib/absinthe/subscription/local.ex#L49 (and above on 46).

benwilson512:
Fixes #401

### DIFF
--- a/lib/absinthe/subscription/local.ex
+++ b/lib/absinthe/subscription/local.ex
@@ -46,7 +46,7 @@ defmodule Absinthe.Subscription.Local do
     do_get_docs(pubsub, field, topic_fun.(mutation_result))
   end
   defp get_docs(pubsub, field, _mutation_result, key) do
-    do_get_docs(pubsub, field, key)
+    do_get_docs(pubsub, field, to_string(key))
   end
 
   defp do_get_docs(pubsub, field, keys) do
--- a/lib/absinthe/phase/subscription/subscribe_self.ex
+++ b/lib/absinthe/phase/subscription/subscribe_self.ex
@@ -66,9 +66,9 @@ defmodule Absinthe.Phase.Subscription.SubscribeSelf do
         {:error, error}
       val ->
         raise """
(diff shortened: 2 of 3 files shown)


lib/absinthe/subscription/local.ex
```diff
@@ -46,7 +46,7 @@ defmodule Absinthe.Subscription.Local do
     do_get_docs(pubsub, field, topic_fun.(mutation_result))
```
mgwidmann: do you need to `to_string` this or is this for mutation publishes?

benwilson512: It happens on line 49.

benwilson512: oh wait no I'm dumb, hold that thought

benwilson512: fixed via 7b059d0

PR #404, opened 2017-09-29
Setup Dialyxir

BenMorganIO:
Welcome to a broken build!

This sets up dialyxir to run on in Travis. You'll see a lot of errors. Some of them are actual errors.

I also had to modify Travis. We now cache the `_build` and `deps` folders so that we don't have to always redownload and rebuild everything.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,30 @@
 language: elixir
+
 elixir:
   - 1.4.5
   - 1.5.1
+
 notifications:
   recipients:
     - bruce.williams@cargosense.com
     - ben.wilson@cargosense.com
+
 otp_release:
   - 19.2
   - 20.0
-script: "MIX_ENV=test mix local.hex --force && MIX_ENV=test mix do deps.get, test"
+
+cache:
+  directories:
+    - _build
+    - deps
+
+env:
+  - MIX_ENV=test
+
+install:
+  - mix local.hex --force
+  - mix local.rebar --force
+  - mix do deps.get, compile
+
+script:
+  - mix test
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,3 +28,4 @@ install:
 
 script:
   - mix test
+  - mix dialyzer
--- a/mix.exs
+++ b/mix.exs
@@ -37,7 +37,7 @@ defmodule Absinthe.Mixfile do
       {:ex_spec, "~> 2.0.0", only: :test},
       {:ex_doc, "~> 0.14", only: :dev},
       {:benchfella, "~> 0.3.0", only: :dev},
-      {:dialyze, "~> 0.2", only: :dev},
+      {:dialyxir, "~> 0.5", only: :dev},
       {:decimal, "~> 1.0", optional: :true},
       {:phoenix_pubsub, ">= 0.0.0", only: :test},
       {:mix_test_watch, "~> 0.4.1", only: [:test, :dev]}
--- a/mix.lock
+++ b/mix.lock
@@ -1,6 +1,6 @@
 %{"benchfella": {:hex, :benchfella, "0.3.4", "41d2c017b361ece5225b5ba2e3b30ae53578c57c6ebc434417b4f1c2c94cf4f3", [:mix], [], "hexpm"},
   "decimal": {:hex, :decimal, "1.4.0", "fac965ce71a46aab53d3a6ce45662806bdd708a4a95a65cde8a12eb0124a1333", [], [], "hexpm"},
-  "dialyze": {:hex, :dialyze, "0.2.1", "9fb71767f96649020d769db7cbd7290059daff23707d6e851e206b1fdfa92f9d", [], [], "hexpm"},
+  "dialyxir": {:hex, :dialyxir, "0.5.1", "b331b091720fd93e878137add264bac4f644e1ddae07a70bf7062c7862c4b952", [], [], "hexpm"},
   "earmark": {:hex, :earmark, "1.2.2", "f718159d6b65068e8daeef709ccddae5f7fdc770707d82e7d126f584cd925b74", [:mix], [], "hexpm"},
   "ex_doc": {:hex, :ex_doc, "0.16.2", "3b3e210ebcd85a7c76b4e73f85c5640c011d2a0b2f06dcdf5acdb2ae904e5084", [:mix], [{:earmark, "~> 1.1", [hex: :earmark, repo: "hexpm", optional: false]}], "hexpm"},
   "ex_spec": {:hex, :ex_spec, "2.0.1", "8bdbd6fa85995fbf836ed799571d44be6f9ebbcace075209fd0ad06372c111cf", [:mix], [], "hexpm"},


lib/absinthe/type/union.ex
```diff
@@ -75,16 +75,11 @@ defmodule Absinthe.Type.Union do
...
   def resolve_type(%{resolve_type: nil, types: types}, obj, %{schema: schema}, opts) do
     type_name = Enum.find(types, fn
-      %{is_type_of: nil} ->
-        false
+      %{is_type_of: nil} -> false
       type ->
         case Schema.lookup_type(schema, type) do
-          nil ->
-            false
-          %{is_type_of: nil} ->
```
benwilson512: `is_type_of` will never be nil?

BenMorganIO: Yep, Dialyzer issued a large error for it.

BenMorganIO:
```erlang
lib/absinthe/type/union.ex:84: The pattern #{'is_type_of':='nil'} can never match the type #{'__struct__':='Elixir.Absinthe.Type.Enum' | 'Elixir.Absinthe.Type.Field' | 'Elixir.Absinthe.Type.InputObject' | 'Elixir.Absinthe.Type.Interface' | 'Elixir.Absinthe.Type.List' | 'Elixir.Absinthe.Type.NonNull' | 'Elixir.Absinthe.Type.Object' | 'Elixir.Absinthe.Type.Scalar' | 'Elixir.Absinthe.Type.Union', '__private__'=>[{_,_}], '__reference__'=>#{'__struct__':='Elixir.Absinthe.Type.Reference', 'identifier':=atom(), 'module':=atom(), 'name':=binary()}, 'args'=>'nil' | #{atom() | binary()=>map()}, 'complexity'=>'nil' | fun((...) -> non_neg_integer()) | non_neg_integer() | {atom(),atom()}, 'config'=>_, 'default_value'=>_, 'deprecation'=>'nil' | #{'reason':=binary()}, 'description'=>'nil' | binary(), 'field_imports'=>_, 'fields'=>fun(() -> map()) | map(), 'identifier'=>atom(), 'interfaces'=>[map()], 'is_type_of'=>fun((_) -> boolean()), 'middleware'=>[], 'name'=>binary(), 'of_type'=>_, 'parse'=>fun((_) -> 'error' | {_,_}), 'resolve_type'=>fun((_,map()) -> atom()), 'serialize'=>fun((_) -> any()), 'triggers'=>_, 'type'=>atom(), 'types'=>[atom()], 'values'=>#{binary()=>map()}, 'values_by_internal_value'=>_, 'values_by_name'=>_}
```

benwilson512: Given this schema:
```
defmodule Foo do
  use Absinthe.Schema

  query do

  end

  union :foo do
    types [:user, :dog]
  end

  object :user do
  end

  object :dog do
  end

end
```

```
iex(2)> Absinthe.Schema.lookup_type(Foo, :foo)
%Absinthe.Type.Union{__private__: [],
 __reference__: %{identifier: :foo, location: %{file: "iex", line: 8},
   module: Foo}, description: nil, name: "Foo", resolve_type: nil,
 types: [:user, :dog]}
iex(3)> Absinthe.Schema.lookup_type(Foo, :user)
%Absinthe.Type.Object{__private__: [],
 __reference__: %{identifier: :user, location: %{file: "iex", line: 12},
   module: Foo}, description: nil, field_imports: [], fields: %{},
 identifier: :user, interfaces: [], is_type_of: nil, name: "User"}
```

resolve_type is nil on the union, and is_type_of is nil on the object type.

BenMorganIO:
Getting this error on Travis, which is funny because it is actually setup:

```shell
$ mix dialyzer
** (Mix) The task "dialyzer" could not be found

The command "mix dialyzer" exited with 1.
```

lib/absinthe/type/union.ex
BenMorganIO: Sounds like we need to update a spec then....

BenMorganIO: I've removed this change entirely since it sounds more like it's the wrong solution to the problem.

benwilson512: yeah it looks like the real issue is that the output of `lookup_types` should really be one of the type structs, not `Type.t` which merely refers (at the moment at least) to one of the type modules, not to their structs.

benwilson512: actually no that isn't true.

benwilson512: ah, here may be the issue: https://github.com/absinthe-graphql/absinthe/blob/v1.3.2/lib/absinthe/type/object.ex#L93 should be `| nil`

BenMorganIO: Yeah, that could very well be the issue.

jessedijkstra:
@BenMorganIO I believe it's because `only: "dev"`. Should be `only: [:test, :dev]`

benwilson512:
Hey @BenMorganIO, so I spent a good bit of time this morning trying to get dialyzer to be useful, and it's been quite the up hill battle. Perhaps you can offer some insight here.

The primary initial output of dialyzer is a very large number of missing function and missing type errors that lack line numbers:

```
:0: Unknown function 'Elixir.Absinthe.Blueprint.Draft.BitString':'__impl__'/1
:0: Unknown function 'Elixir.Absinthe.Blueprint.Draft.Float':'__impl__'/1
:0: Unknown function 'Elixir.Absinthe.Blueprint.Draft.Function':'__impl__'/1
:0: Unknown function 'Elixir.Absinthe.Blueprint.Draft.Integer':'__impl__'/1
:0: Unknown function 'Elixir.Absinthe.Blueprint.Draft.Map':'__impl__'/1
:0: Unknown function 'Elixir.Absinthe.Blueprint.Draft.PID':'__impl__'/1
:0: Unknown function 'Elixir.Absinthe.Blueprint.Draft.Port':'__impl__'/1
:0: Unknown function 'Elixir.Absinthe.Blueprint.Draft.Reference':'__impl__'/1
:0: Unknown function 'Elixir.Absinthe.Blueprint.Draft.Tuple':'__impl__'/1
:0: Unknown function 'Elixir.Absinthe.Traversal.Node.BitString':'__impl__'/1
:0: Unknown function 'Elixir.Absinthe.Traversal.Node.Float':'__impl__'/1
:0: Unknown function 'Elixir.Absinthe.Traversal.Node.Function':'__impl__'/1
:0: Unknown function 'Elixir.Absinthe.Traversal.Node.Integer':'__impl__'/1
:0: Unknown function 'Elixir.Absinthe.Traversal.Node.List':'__impl__'/1
:0: Unknown function 'Elixir.Absinthe.Traversal.Node.Map':'__impl__'/1
:0: Unknown function 'Elixir.Absinthe.Traversal.Node.PID':'__impl__'/1
:0: Unknown function 'Elixir.Absinthe.Traversal.Node.Port':'__impl__'/1
:0: Unknown function 'Elixir.Absinthe.Traversal.Node.Reference':'__impl__'/1
:0: Unknown function 'Elixir.Absinthe.Traversal.Node.Tuple':'__impl__'/1
:0: Unknown function 'Elixir.Dataloader':get/4
:0: Unknown function 'Elixir.Dataloader':load/4
:0: Unknown function 'Elixir.Dataloader':'pending_batches?'/1
:0: Unknown function 'Elixir.Dataloader':run/1
:0: Unknown function 'Elixir.Decimal':'nan?'/1
:0: Unknown function 'Elixir.Decimal':new/1
:0: Unknown function 'Elixir.Decimal':parse/1
:0: Unknown function 'Elixir.Decimal':to_string/1
:0: Unknown function 'Elixir.Mix.Generator':create_directory/1
:0: Unknown function 'Elixir.Mix.Generator':create_file/3
:0: Unknown function 'Elixir.Mix.Task':run/2
:0: Unknown function httpd_util:hexlist_to_integer/1
:0: Unknown type 'Elixir.Absinthe.Blueprint.Result':node_t/0
:0: Unknown type 'Elixir.Absinthe.Resolution.Result.Object':t/0
:0: Unknown type 'Elixir.Absinthe.Type.Schema':t/0
:0: Unknown type 'Elixir.Blueprint':flags_t/0
:0: Unknown type 'Elixir.Blueprint.Document.SourceLocation':t/0
:0: Unknown type 'Elixir.Decimal':t/0
:0: Unknown type 'Elixir.Document.Resolution':acc/0
```

While I'm sure some of those latter ones are in fact real errors, the lack of line number makes tracking it down wildly obnoxious.

1) Any idea how I can get line numbers
2) Any idea why using the Mix, Decimal, or DataLoader functions has issues?
3) What to do about the protocol errors?

There are of course other errors, but if the tool won't even give me line numbers I confess I'm skeptical about its general usefulness.

BenMorganIO:
1. Any idea if there are any "no local return" messages? This usually pops up because an error is prohibiting a local return, because, well... An error just doesn't let anything return.

2. For mix, decimal, and data_loader, they may not be in your applications for dialyxir.

benwilson512:
@ericmj summarizes my feelings on dialyzer here rather well: https://elixirforum.com/t/phoenix-project-doesnt-use-dialyzer/10361/5

I am more than happy to accept PRs to fix typespec problems, but the gigantic pain in the butt factor created by placing this as a CI step just isn't worth it for me.

PR #409, opened 2017-10-09
Rename Blueprint.Document.Resolution to Blueprint.Execution

benwilson512:
This solves some long standing confusion caused by overloading the term "resolution". By switching this name we can talk clearly about "executing a document" and "resolving a field". This execution struct tracks the progress of the execution progress, and the Resolution struct is used on each field to resolve it.

The resolution struct shares some important values in common with the execution struct, and this PR also makes sure that all three of those: `acc`, `context`, and `fields_cache` are readily available within plugins.

### DIFF
--- a/lib/absinthe/blueprint/result/object.ex
+++ b/lib/absinthe/blueprint/result/object.ex
@@ -1,13 +1,14 @@
-defmodule Absinthe.Blueprint.Document.Resolution.List do
+defmodule Absinthe.Blueprint.Result.Object do
 
   @moduledoc false
 
   alias Absinthe.{Blueprint, Phase}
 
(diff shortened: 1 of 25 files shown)


lib/absinthe/blueprint/result/object.ex
```diff
@@ -1,4 +1,4 @@
-defmodule Absinthe.Blueprint.Document.Resolution.Object do
+defmodule Absinthe.Blueprint.Result.Object do
```
benwilson512: @bruce moved these into the result namespace. It's both more descriptive of what they are, and also more technically correct. Resolution is not necessary to end up with these.

Issue #412, opened 2017-10-12
arguments for enum with default_value are lowercase for introspection


Gazler:
Consider the following schema:

```elixir
defmodule SomeSchema do
  use Absinthe.Schema.Notation

  enum :my_enum do
    value :foo
    value :bar
    value :baz
  end

    field :my_thing, :string do
      arg :my_arg, :my_enum, default_value: :foo
      resolve fn _, _, _ ->
        "return value"
      end
    end
end
```

When introspecting, this will return "FOO", "BAR" and "BAZ" as the enum values. However the default value for "my_arg" will be "foo". This is not a valid value for the argument, and will cause breakages in certain clients (the ruby graphql client in my case.)

Changing:

```elixir
      arg :my_arg, :my_enum, default_value: :foo
```

To:

```elixir
      arg :my_arg, :my_enum, default_value: :FOO
```

Fixes the issue.

benwilson512:
It also shows the wrong value when simply looking up the enum type in GraphiQL.

fsabe:
Hello. I'm learning Absinthe and this issue seemed simple enough to tackle. Do you mind taking a look at my solution?

bruce:
I believe this is now resolved due to @fsabe's contribution in #413.

PR #413, opened 2017-10-16
Use enum name instead of value in default value when instrospecting

fsabe:
https://github.com/absinthe-graphql/absinthe/issues/412

This changes the `__InputValue` to use the enum name instead of its value when the `default value` is an enum type.

### DIFF
--- a/lib/absinthe/type/built_ins/introspection.ex
+++ b/lib/absinthe/type/built_ins/introspection.ex
@@ -279,6 +279,8 @@ defmodule Absinthe.Type.BuiltIns.Introspection do
 
         _, %{schema: schema, source: %{default_value: value, type: type}} ->
           case Absinthe.Schema.lookup_type(schema, type, unwrap: true) do
+            %Absinthe.Type.Enum{values_by_internal_value: values} ->
+              {:ok, values[value].name}
             %{serialize: serializer} ->
               {:ok, inspect serializer.(value)}
             _ ->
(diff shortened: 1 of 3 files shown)


test/lib/absinthe/introspection_test.exs
```diff
@@ -139,6 +139,50 @@ defmodule Absinthe.IntrospectionTest do
...
+              }
+            }
+          }
+        }
+      }
+      """
+      |> run(ColorSchema)
+      assert {:ok, %{data: %{"__schema" => %{"queryType" => %{"fields" => fields}}}}} = result
+      assert [
+        %{"name" => "info", "args" => [%{"name" => "channel", "defaultValue" => "RED"}]}
```
benwilson512: @bruce this seems right to me

lib/absinthe/type/built_ins/introspection.ex
```diff
@@ -279,6 +279,8 @@ defmodule Absinthe.Type.BuiltIns.Introspection do
 
         _, %{schema: schema, source: %{default_value: value, type: type}} ->
           case Absinthe.Schema.lookup_type(schema, type, unwrap: true) do
+            %Absinthe.Type.Enum{values_by_internal_value: values} ->
+              {:ok, values[value].name}
```
fsabe: I'm not sure how to handle when the default value does not map to any enum value.

benwilson512: @fsabe we should probably try to add a schema validation rule that ensures all default values are valid. Let's handle that separately.

bruce:
Merging for inclusion in next release. Thank you, @fsabe.

PR #416, opened 2017-10-23
Integrate guides

bruce:
We're integrating and updating the guides for the v1.4
release. They'll be in `guides/` and available in hexdocs. The hope
is that this will make it easier to:

- Tie the guides to specific releases
- Make it easier to contribute
- Make it easier to keep them up-to-date

Much of the credit goes to the Phoenix project for the idea and
example ExDoc configuration.


bruce:
@benwilson512 Before we merge this I'd like to finish up absinthe-graphql/absinhe_example#17.

If we can hold the rc release until afternoon ET, that would be helpful.

bruce:
Upgrade Guide...

- [ ] Any content from absinthe_relay?
- [ ] Any content from absinthe_ecto?
- [ ] Any content from absinthe_phoenix?

tlvenn:
Ya we probably want to update the Ecto guide which is very outdated and should probably illustrate how to do batching manually quickly and how to leverage absinthe_ecto fully.

bruce:
@tlvenn There is a notice in the Ecto guide and a stubbed Batching guide for exactly those reasons. After we merge this (we don't want a lack of certain content to hold back the rc or stable, as we can push docs after release anyway), my plan is to create separate `Type: Documentation` issues for missing content.

tlvenn:
My bad, I somehow missed the notice which was in plain sight at the very beginning...

benwilson512:
@tlvenn I actually plan on deprecating Absinthe.Ecto. The `Batch` plugin is still useful for loading stuff where it doesn't make sense to talk about a "source", or if there's some very custom SQL query you want to run.

Dataloader does everything Absinthe.Ecto does at the moment but with an improved API, without violating contexts. Absinthe.Ecto will stick around cause I know people's projects rely on it, but at this point the plan is to promote Dataloader as its general replacement.

bruce:
@benwilson512 Seems like the right thing, documentation-wise, is to have the "Ecto Best Practices" guide document the existing absinthe_ecto API, make the case you just mentioned, and then transition into the better way, using Dataloader -- and linking to the more general Dataloader guide.

benwilson512:
I'm fine having documentation for how to use `Absinthe.Ecto` somewhere, but I'm hesitant to put it under the title of "Ecto Best Practices" given that it promotes hand written SQL queries all over your schema, a practice no longer considered "best".

bruce:
@benwilson512 Feel free to change the title; that's just what we had before. I'd already changed the filename to just plain `ecto.md`.

bruce:
(We're going to release this _after_ rc.3—still later today—as we have to get a release out quickly for another beta release of [the book](https://pragprog.com/book/wwgraphql/craft-graphql-apis-in-elixir-with-absinthe).)

guides/context-and-authentication.md
```diff
@@ -104,18 +104,8 @@ defmodule MyApp.Web.Context do
...
-        conn
-        |> send_resp(403, reason)
-        |> halt()
-      _ ->
-        conn
-        |> send_resp(400, "Bad Request")
-        |> halt()
-    end
+    context = build_context(conn)
+    put_private(conn, :absinthe, %{context: context})
```
yordis: ❤️

bruce:
Going to go ahead and merge this!

PR #418, opened 2017-10-25
Complexity analysis fragments

benwilson512:


lib/absinthe/phase/document/validation/no_fragment_cycles.ex
```diff
@@ -67,11 +69,12 @@ defmodule Absinthe.Phase.Document.Validation.NoFragmentCycles do
   @spec vertex(Blueprint.Document.Fragment.Named.t, :digraph.graph) :: Blueprint.Document.Fragment.Named.t
   defp vertex(%Blueprint.Document.Fragment.Named{} = fragment, graph) do
     :digraph.add_vertex(graph, fragment.name)
-    Enum.each(fragment.selections, fn
+    Blueprint.prewalk(fragment, fn
```
benwilson512: This was the root issue. The fragment reference detector wasn't doing a deep walk on the fragments. This meant that the topological sort we do on fragments wasn't happening. Consequently, later iteration through the fragments was unsafe.

PR #421, opened 2017-10-28
More spec-compliant handling of non null fields that return null

benwilson512:
- [x] add error to results
- [x] trim up to the first nullable field
- [x] Nullify top level data key if necessary
- [x] Make sure it works with lists.

### DIFF
--- a/lib/absinthe/phase/document/execution/resolution.ex
+++ b/lib/absinthe/phase/document/execution/resolution.ex
@@ -48,7 +48,10 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
 
     exec = plugins |> run_callbacks(:before_resolution, exec, run_callbacks?)
 
-    {result, exec} = walk_result(exec.result, operation, operation.schema_node, exec, [operation])
+    {result, exec} =
+      exec.result
+      |> walk_result(operation, operation.schema_node, exec, [operation])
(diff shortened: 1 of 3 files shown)


lib/absinthe/phase/document/execution/resolution.ex
```diff
@@ -202,37 +202,74 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
...
+    end
+
+    errors = maybe_add_non_null_error(errors, value, full_type)
 
-    %{result | extensions: extensions}
+    value
+    |> to_result(bp_field, full_type, extensions)
+    |> add_errors(Enum.reverse(errors), &put_result_error_value(&1, &2, bp_field, source, path))
     |> walk_result(bp_field, full_type, exec, path)
+    |> propagate_null_trimming
```
benwilson512: @bruce In the end I decided this was best accomplished within the Resolution phase itself. If we don't handle it here then the phase produces a non-spec compliant result tree, which each variant of the Result phase has to handle for itself. It's relatively important that the Result phases be kept simple because they're a key part of stuff like the Phoenix controller integration, any extension work, and so forth.

There's also a handy performance perk to doing it in this phase since we just make it part of the general stack unwinding process that builds the result tree, instead of doing a dedicated pass afterward.

PR #424, opened 2017-10-30
Improve Dataloader helper

### LINKED ISSUE #422: Dataloader helper should accept arguments defined at compile time
As discussed with @benwilson512 over Slack, the current `Dataloader` helper makes the assumption that only args coming from the field will drive further filtering / ordering and so forth. I have a lot of fields in my schema which dont have any args and yet do filtering and preloading.

Example:

```
field :photo, :asset, resolve: assoc(:photo, fn query, _, _ ->
        query
        |> Asset.with_purpose(:user_photo)
        |> Asset.with_current_version
      end)
```

We should probably be able to pass some arguments defined at compile time such as:

```
field :photo, :asset, resolve: dataloader(Asset, {:photo, purpose: :user_photo})
```

As Ben suggested, this could be even more flexible using the following form:

```
dataloader(Asset, fn parent, args, info ->
  {:photo, %{purpose: :user_photo, foo: args.bar * 2}}
end
```

which allows using the `parent/args/info` in order to construct the key for example.

tlvenn:
Fixes #422

### DIFF
--- a/lib/absinthe/resolution.ex
+++ b/lib/absinthe/resolution.ex
@@ -22,6 +22,9 @@ defmodule Absinthe.Resolution do
 
   To access the schema type for this field, see the `definition.schema_node`.
   """
+  @type arguments :: %{optional(atom) => any}
+  @type source :: any
+
   @type t :: %__MODULE__{
     value: term,
     errors: [term],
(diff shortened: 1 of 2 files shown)


benwilson512:
Thank you!

Issue #410, opened 2017-10-09
Umbrella Issue for Schema Code


benwilson512:
This issue will serve as the umbrella issue for problems related to the fact that Absinthe Schemas will allow "invalid" code without throwing a compilation error. Examples include:
```
object :foo do
  field :name # missing type
  this_function_doesnt_exist() # should explode
end
```

This happens because within each `do` block all that Absinthe does is walk through and expand macros. Everything else is discarded. This produces unexpected behaviour, and is slated to be looked at in the planned 1.5 schema rework.

benwilson512:
This works now in 1.5 / master

PR #426, opened 2017-10-30
Rename input object test and fix test failures

joshprice:
The test file was missing the `_test` suffix and therefore was not included by `mix test` .

The rule was renamed at one point and never tested because of the naming issue, so fixed that too.

Just a note about the Rule tests, they are rather awkward to debug when they fail, so happy to raise a separate issue/PR for this. (I'm fixing something else related and this was just something I happened to stumble across as I dug into the code).


benwilson512:
Hey Josh! Thanks for the PR. I agree that the rule tests can be a pain to work with. Can you add a comment to that effect under https://github.com/absinthe-graphql/absinthe/issues/410 ?

We're waiting to create individualized issues about the schema until we begin our planned 1.5 rework of schema internals, and we are trying to consolidate various issues and concerns with the current approach.

PR #428, opened 2017-10-30
docs for dataloader functions

benwilson512:
@tlvenn


benwilson512:
The main thing to note here is that the possible forms of `dataloader/1,2,3` are:

```
dataloader(Ctx)
dataloader(Ctx, :foo_field)
dataloader(Ctx, &fun/1)
dataloader(Ctx, &fun/1, use_parent: false)
```

PR #429, opened 2017-10-30
Don't clobber the contents of the result

binaryseed:
When generating the final `result`, we don't want to clobber anything that was previously added. Specifically, we may have added an `extensions` key, which is to be included in the final response.

I tweaked the `extensions_test` to validate that this woks as expected...

@benwilson512

### DIFF
--- a/test/lib/absinthe/extensions_test.exs
+++ b/test/lib/absinthe/extensions_test.exs
@@ -44,4 +44,17 @@ defmodule Absinthe.ExtensionsTest do
 
     assert bp.result == %{data: %{"foo" => "hello world"}, extensions: %{foo: 1}}
   end
+
+  it "Result phase doesn't clober the extensions" do
+    doc = "{foo}"
+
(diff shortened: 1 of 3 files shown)


test/lib/absinthe/extensions_test.exs
```diff
@@ -38,7 +38,7 @@ defmodule Absinthe.ExtensionsTest do
     pipeline =
       Schema
       |> Absinthe.Pipeline.for_document()
-      |> Absinthe.Pipeline.insert_after(Absinthe.Phase.Document.Result, MyPhase)
+      |> Absinthe.Pipeline.insert_before(Absinthe.Phase.Document.Result, MyPhase)
```
benwilson512: Do you mind adding a second dedicated test case for this instead of changing this one? We want to test both directions really.

binaryseed: sure

PR #430, opened 2017-10-31
Return a better error message when the default type for an enum is no…

Gazler:
…t present

Previously:

> ** (UndefinedFunctionError) function nil.name/0 is undefined or private
>     nil.name()
>     (absinthe) lib/absinthe/type/built_ins/introspection.ex:283:
>     anonymous fn/2 in Absinthe.Type.BuiltIns.Introspection.__absinthe_type__/1

Now:

> ** (RuntimeError) could not use default value of "other" for relation_type.
>
> Valid types are:
>
>  * friend
>  * family
>  * following

### DIFF
--- a/lib/absinthe/schema/rule.ex
+++ b/lib/absinthe/schema/rule.ex
@@ -32,6 +32,7 @@ defmodule Absinthe.Schema.Rule do
     Rule.ObjectMustImplementInterfaces,
     Rule.InterfacesMustResolveTypes,
     Rule.InputOuputTypesCorrectlyPlaced,
+    Rule.DefaultEnumValuePresent,
   ]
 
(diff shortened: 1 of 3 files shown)


Gazler:
A couple things I'm not sure of here.

1) The location of the test.
2) The type of error. Currently I am just using raise, but perhaps it should be an `Absinthe.Schema.Error`

benwilson512:
Hey! I love the error message, really great. However it strikes me that this is really something we could do as a schema rule and validate at compile time.

By way of example, see: https://github.com/absinthe-graphql/absinthe/blob/master/lib/absinthe/schema/rule/query_type_must_be_object.ex

I think you'd be able to just go through all the objects in a schema, and make sure that if their fields are an enum type, that their default values are actually in the enum.

benwilson512:
Thanks!

PR #432, opened 2017-11-01
Fixes #431

### LINKED ISSUE #431: Regression: Introspection query is broken
I updated my project dependencies this morning to use the latest (master) from Absinthe and suddenly the introspection query ran by Graphiql is returning an error:

```
Error: Invalid or incomplete schema, unknown type: Currency. Ensure that a full introspection query is used in order to build a client schema.
```

`Currency` is a well defined enum on absinthe side.

I am wondering if that regression is not due to this commit: 63a2e616f57604d3b897997abfa3238e6f40bc5d 

@bruce that was your commit, you might have an idea of what is going on.

benwilson512:
- [x] tweak to use identifiers
- [x] probably also tweak to use mapsets.

### DIFF
--- a/lib/absinthe/schema/notation.ex
+++ b/lib/absinthe/schema/notation.ex
@@ -298,6 +298,7 @@ defmodule Absinthe.Schema.Notation do
   @doc false
   # Record an interface type
   def record_interface!(env, identifier, attrs, block) do
+    attrs = Keyword.put(attrs, :identifier, identifier)
     scope(env, :interface, identifier, attrs, block)
   end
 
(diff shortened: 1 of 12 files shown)


lib/absinthe/schema/notation.ex
```diff
@@ -298,6 +298,7 @@ defmodule Absinthe.Schema.Notation do
   @doc false
   # Record an interface type
   def record_interface!(env, identifier, attrs, block) do
+    attrs = Keyword.put(attrs, :identifier, identifier)
```
benwilson512: @bruce Note, I've made sure every top level entity now has a top level key `:identifier`. It ends up being reasonably important for user code, and doing `__reference__.identifier` felt dirty.

PR #435, opened 2017-11-03
Set default complexity to all introspection fields to 0

Gazler:
Previously, with a max introspection, it was possible to hit the
complexity limit from an introspection query (such as that from
graphiql). This commit fixes that by using a default complexity of 0 for
all the introspection types.

### DIFF
--- a/lib/absinthe/type/built_ins/introspection.ex
+++ b/lib/absinthe/type/built_ins/introspection.ex
@@ -7,6 +7,7 @@ defmodule Absinthe.Type.BuiltIns.Introspection do
     description "Represents a schema"
 
     field :types, list_of(:__type) do
+      complexity 0
       resolve fn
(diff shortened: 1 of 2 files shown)


benwilson512:
I'm not entirely sure that this is the right call. I think on the whole even Map.get fields should be weighted, they just don't weigh much. What was the complexity limit you were using?

Gazler:
@benwilson512 50, and ~140 was used for the introspection query.

benwilson512:
Ah, I should probably document this but I'm expecting that ordinary complexity values will be in the thousands, with large in the 10s of thousands. Fields that require DB hits weigh in at ~500 or so in such cases.

bruce:
What's the right next step here? Close this in favor of a documentation PR?

benwilson512:
I think that's the right call yea.

PR #441, opened 2017-11-13
Import batch/4

telnicky:
I noticed that I was enable to pass options to the batch function without importing batch/4 directly


telnicky:
Seeing the test failures and how `async/1` behaves as well it looks like the desired pattern is to import the function without the option arg. Closing for now

PR #442, opened 2017-11-14
Bump dataloader version in ecto guide

radar:
I noticed the ecto guide was using an older version of dataloader. This PR bumps it to 1.0.0.


bruce:
Thanks!

Issue #352, opened 2017-07-22
Wrong ordering of returned fields


phyrog:
The returned fields seem to be in descending order, no matter in which order they are specified in the query.

According to the [GraphQL spec](http://facebook.github.io/graphql/#sec-Serialization-Format), the returned data have the same order as the fields in the query:

> Serialization formats which can represent an ordered map should preserve the order of requested fields as defined by CollectFields() in the Execution section. Serialization formats which can only represent unordered maps should retain this order grammatically (such as JSON).
>
> Producing a response where fields are represented in the same order in which they appear in the request improves human readability during debugging and enables more efficient parsing of responses if the order of properties can be anticipated.

bruce:
The previous version of the spec specified "must" vs "should," and the change was made after lengthy feedback by myself and @benwilson512 -- read https://github.com/facebook/graphql/issues/168. Since JSON object keys are (according to the JSON spec) unordered, this isn't on our roadmap to "fix" anytime soon. That said, we'd certainly welcome a PR for a custom serializer that adds this feature, provided it did so in a way that didn't noticeably degrade performance.

bruce:
@benwilson512 Leaving open for you to make a comment if you'd like.

phyrog:
Thanks for the feedback and the link to the discussion. Since I'm still new to Elixir and not familiar with the code base it's unlikely that I can produce a PR for this, but I can certainly have a look.

It's also not something that is really critical to me, it just surprised me today when using GraphiQL

bruce:
Thanks, @phyrog, and if you could use any tips while looking through, feel free to join us in the #absinthe-graphql channel in the Elixir Slack; we're usually around and try to be helpful :-)

benwilson512:
Yeah just to chime in, this is purely a function of how the `Absinthe.Phase.Document.Result` builds a result map, which is then used as an input to Poison. Elixir maps (and consequently Poison) don't have any order. This is definitionally fine for JSON objects.

If the ordering thing is a sticking point it would be very easy to replace the Result phase (it's a very simple phase) with something that built tuple lists instead of maps, and used a JSON codec that used some specific order.

MartinKavik:
Hello guys! I would like to reopen this issue because it's a show-stopper for me and I'd like to fix it. And fix it right.

### Why reopen
[...]

PR #445, opened 2017-11-15
Ordered results: + Document.OrderedResult integration

MartinKavik:
Issue #352 
Well, this PR is quite big, but the most changes are new tests for ordered results.

- The main goal was to add config option `:ordered` which switches between `Document.Result` and `Document.OrderedResult`
- All tests are running on `mix test`

I can provide more info / write comments, etc. if you want. Thanks for review.


benwilson512:
Hey @MartinKavik I've created a PR on the absinthe_tutorial project that shows how you can accomplish this without any changes to Absinthe at all: https://github.com/absinthe-graphql/absinthe_tutorial/pull/1

As noted in that PR, we may well make a few changes to Absinthe so that folks can avoid copying the whole `Result` module over, but other than that things are pretty straight forward. Let's continue the discussion on that PR.

bruce:
@benwilson512 Should this be closed?

benwilson512:
Yea, this isn't the way we want to go about this and it isn't feasible to try to redirect a PR this size.

PR #446, opened 2017-11-16
InputObject Parse Context

maennchen:
This PR is in a very early stage. I created it to get the discussion of Slack organized here.

## Problem

Since there is no `InputUnion` in GraphQL yet, it is currently not possible to build a scalar that returns two possible types without guessing based on the input.

## Example
Based on the assumption that we take either a OpenPGP Key / SMIME Cert as input.

### Problem

```elixir
scalar :public_key, name: "PublicKey" do
  parse fn
    %Input.String{value: value}, _ ->
      # How should I know here what type the value is?
      # It is a binary and we should not look into it before parsing
    _, _ ->
      :error
  end
end
```

### My Preferred Solution

This is my personal preferred solution. This is a matter of discussion though.

```elixir
enum :public_key_enum, values: [:open_pgp, :smime]

input_object :public_key do
  field :type, :public_key_enum
  field :value, :public_key_value, dependent: [:type]
end

scalar :public_key_value, name: "PublicKeyValue" do
  parse fn
    %Input.String{value: value, flags: %{type: :open_pgp}}, _ ->
      {:ok, :here_would_be_the_parsed_gpg_key}
    %Input.String{value: value, flags: %{type: :smime}} ->
      {:ok, :here_would_be_the_parsed_smime_cert}
    _, _ ->
      :error
  end
end
```

## Implementation

Existing phases should skip the loading of a field that has the argument `dependent` given.

A new phase should set the dependent values into the flags and load those fields.


benwilson512:
Hey! So I was playing around with this as well as an alternative take I was working on that wouldn't require absinthe changes, and both run into a bit of an issue: There is no guaranteed order that input object fields will be parsed in.

What happens if the enum value is invalid? Who says the enum value will even be parsed first? If they're in a different order in the schema module will that break it?

These kinds of questions have me thinking that trying to achieve this at the _parse_ level is the wrong approach. You could easily imagine circular parse dependencies cropping up as well.

At the end of the day if you want to treat this as a GraphQL validation I think you're going to need dedicated scalar values, and discrete arguments, at least until GraphQL supports input unions. This may be unpalatable, so you could also look at treating them as a data validation with middleware. IE:

```
input_object :public_key do
  field :type, :public_key_enum
  field :value, :string
end

def middleware(middleware, field, _object) do
  # if the field arguments have a `:public_key` input in them
  # add middleware that will validate the decode process, and if it fails add
  # an error.
end
  middleware
end
```

maennchen:
@benwilson512 

> Who says the enum value will even be parsed first?
> What about a different order in the GraphQL document?

Can't we ensure that fields without dependencies will be parsed first (do this in two phases).

I'd skip the parsing of values with dependencies in the normal phase and parse them in a separate phase.

> What happens if the enum value is invalid?

This should be a normal error on the enum of the input object.

> You could easily imagine circular parse dependencies cropping up as well.

Since this is a detour to having real input unions, I think that we can restrict the usage of scalars depending on enums. This way we solve that issue.

> These kinds of questions have me thinking that trying to achieve this at the parse level is the wrong approach.

I think that is not an optimal solution since this will tell the client that the input is valid if it actually wasn't. One of the reasons I chose to use GraphQL is the static typing.

benwilson512:
> Can't we ensure that fields without dependencies will be parsed first (do this in two phases).

Ah, I somehow missed the `dependent` annotation.

> This should be a normal error on the enum of the input object.

Sure, but what do we do about the value field? Should it also have an error? There's certainly no standard error for "An field I'm dependent on couldn't be parsed".

> One of the reasons I chose to use GraphQL is the static typing.

Sure, but technically there's already a fully static typed solution, it's just verbose.

```
scalar :smime_value, do: ...
scalar :pgp_value, do: ...

field :decode, :foo do
  arg :smime_input, :smime_value
  arg :pgp_input, :pgp_value
end
```

And if you wanted to make htem `non_null` you'd just have 2 fields. Verbose yes, but also strict.

The problem is simply that the static typing for inputs is very simple, and lacks certain ADT capabilities that could lower verboseness.

Let me talk this over with Bruce. The problem with adding something like this to Absinthe itself is that we're now on the hook for promoting, documenting, and maintaining this approach. While Absinthe has plenty of capabilities not found in the spec, none of them to date have been validation related, they're all just tooling around resolvers, which are pretty much black boxes as far as the spec is concerned.

This features solves this very specific type of case, but there are a host of other desired features around more flexible input argument handling (anytime anyone looks at geolocation data for example) and we need to be incredibly careful that we don't ship several unrelated and possibly conflicting features around this space, only to later have the spec change and break everyone's code. I guarantee you folks would want to do this to arguments as well as input objects, and then there's a question of whether you could do this to a nested value INSIDE another input object, and so on.

I'll also take some more time to see if there's a phase you could add to _your_ pipelines to add this functionality without requiring a commitment or stance on the issue from Absinthe.

In short the options going forward are:

A) Push this in basically "as is" as an isolated slightly hacky spot feature
B) Work this into a major feature set surrounding custom handling of inputs
C) Use existing pipeline hooks to solve this for just you.

maennchen:
> Ah, I somehow missed the `dependent` annotation.

It's not implemented yet, so it's hard to miss...

> Sure, but what do we do about the value field? Should it also have an error? There's certainly no standard error for "A field I'm dependent on couldn't be parsed".

I'd throw an error on both fields. I'm however aware that this is no perfect solution.

> In short the options going forward are:
>
> A) Push this in basically "as is" as an isolated slightly hacky spot feature
> B) Work this into a major feature set surrounding custom handling of inputs
> C) Use existing pipeline hooks to solve this for just you.

I don't really like option A since it solves a problem for users but we do not actually promote it which makes it a little useless.

If option B with a more general approach would work, that would be really great. (What are the use-cases for the feature except input unions?)

If option C works we could also create a little 3rd party lib. We could link to the lib in the absinthe documentation. The 3rd library itself should write that this is just a hacky feature which will be replaced by input unions.

If there are other use cases than the input union, I'd go for option B, otherwise for option C. But this is just my two cents.

Either way I'm happy to contribute towards the solution.

bruce:
Closing as stale. I think that, in general, the ideas here would be better looked at after the extensibility features of v1.5 are in-place; there may be a pattern that will make things like this feel less invasive (as we don't want to extend the DSL macros, generally speaking).

PR #447, opened 2017-11-16
Feature/import types from many modules

ssomnoremac:
Ok, I'm taking a stab at importing types like so:

`import_types MyApp.Schema.Types.{TypesA, TypesB}`

This is my first attempt at dealing with the AST, so I'm not sure if this is even remotely what I should be doing. It seems to almost works. 

```
Compilation error in file lib/myapp_web/schema.ex ==
** (ArgumentError) `MyApp.Schema.Types.TypesA` is not a module
This macro must be given a literal module name or a macro which expands to a
literal module name. Variables are not supported at this time.
```

I'd appreciate any advice on how to debug in this situation and if I am actually close or way off on this. Thanks!

### DIFF
--- a/lib/absinthe/schema/notation.ex
+++ b/lib/absinthe/schema/notation.ex
@@ -1139,29 +1139,30 @@ defmodule Absinthe.Schema.Notation do
   """
   defmacro import_types(type_module_ast) do
     env = __CALLER__
-    {:ok, _} =
-      type_module_ast
-      |> Macro.expand(env)
-      |> do_import_types(env)
+    type_module_ast
+    |> Macro.expand(env)
+    |> do_import_types(env)
     :ok
   end
 
   defp do_import_types({{:., _, [root_ast, :{}]}, _, modules_ast_list}, env) do
-    # root_ast = {:__aliases__, [alias: false], [:MyApp, :Schema, :Types]}
-    # modules_ast_list = [
-    #    {:__aliases__, [alias: false], [:TypesA]},
-    #    {:__aliases__, [alias: false], [:TypesB]}
-    #  ]
-    {_, _, root} = root_ast
-    Enum.each(modules_ast_list, fn {_, _, leaf} ->
-      do_import_types( root ++ leaf |> Enum.join(".") |> String.to_atom , env)
-    end)
-  end
+    {:__aliases__, _, root} = root_ast
 
+    root_module = Module.concat(root)
+    root_module_with_alias = Keyword.get(env.aliases, root_module, root_module)
+
+    for {_, _, leaf} <- modules_ast_list do
+      type_module = Module.concat([root_module_with_alias | leaf])
+      if Code.ensure_loaded?(type_module) do
+        do_import_types(type_module, env)
+      else
+        raise ArgumentError, "module #{type_module} is not available"
+      end
+    end
+  end
   defp do_import_types(type_module, env) when is_atom(type_module) do
     imports = Module.get_attribute(env.module, :absinthe_imports) || []
     _ = Module.put_attribute(env.module, :absinthe_imports, [type_module | imports])
-
     types = for {ident, name} <- type_module.__absinthe_types__, ident in type_module.__absinthe_exports__ do
       put_definition(
         env.module,
@@ -1190,8 +1191,6 @@ defmodule Absinthe.Schema.Notation do
         }
       )
     end
-
-    {:ok, types: types, directives: directives}
   end
   defp do_import_types(type_module, _) do
     raise ArgumentError, """
--- a/lib/absinthe/schema/notation.ex
+++ b/lib/absinthe/schema/notation.ex
@@ -1133,6 +1133,8 @@ defmodule Absinthe.Schema.Notation do
   ## Examples
   ```
   import_types MyApp.Schema.Types
+
+  import_types MyApp.Schema.Types.{TypesA, TypesB}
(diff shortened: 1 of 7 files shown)


benwilson512:
Can you include a test case so that it's easy for me to pull down the branch and mess with it?

Thank you!

test/lib/absinthe/type/custom/type_b
```diff
@@ -0,0 +1,8 @@
+defmodule Absinthe.Type.Custom.TypeB
+  use Hyr.Command
```
benwilson512: What is this?

ssomnoremac:
@benwilson512 added test which gives the same error above
```mix test test/lib/absinthe/type/import_types_test.exs```

File: test/lib/absinthe/type/custom/type_b (hunk 1 above)
Comment (ssomnoremac) [reply test/lib/absinthe/type/custom/type_b]: my bad, copy and paste error

Comment (ssomnoremac) [reply test/lib/absinthe/type/custom/type_b]: and I don't use it as an input_object, but the import is what matters

Comment (ssomnoremac) [reply test/lib/absinthe/type/custom/type_b]: should be all set now

**Comment by ssomnoremac** [comment]
k, sorry, I f'd up the tests. Passing the module as an atom now to `do_import_types(type_module, env) when is_atom(type_module)` results in the following:

```(module :"Absinthe.Type.Custom.TypeA" is not available)```

I'm not seeing how this is different than when `do_import_types/2` is called for a single module

**Comment by bruce** [comment]
Hey @ssomnoremac, @benwilson512, I think I fixed this. Give it a review when you have a moment. 😄

**Comment by ssomnoremac** [comment]
hey yeah, sorry, I missed that mention last week. @bruce, just tried it with the 30 or so different module imports which correlate to my commanded commands. Everything works! Saves me 30 lines of code in my schema. Thanks! Merge away I say!

**Comment by ssomnoremac** [comment]
Spoke too soon. Seems not to compile sometimes, likely due to `Code.ensure_loaded` usage.
```
 Compilation error in file lib/hyr_web/schema.ex ==
** (ArgumentError) module Elixir.Hyr.Shifts.Opening.Commands.CreateOpening is not available
    lib/absinthe/schema/notation.ex:1159: anonymous fn/4 in Absinthe.Schema.Notation.do_import_types/2
    (elixir) lib/enum.ex:1826: Enum."-reduce/3-lists^foldl/2-0-"/3
    lib/absinthe/schema/notation.ex:1154: Absinthe.Schema.Notation.do_import_types/2
    expanding macro: Absinthe.Schema.Notation.import_types/1
    lib/hyr_web/schema.ex:113: HyrWeb.Schema (module)
    (elixir) lib/kernel/parallel_compiler.ex:121: anonymous fn/4 in Kernel.ParallelCompiler.spawn_compilers/1
```

**Comment by ssomnoremac** [comment]
Well, I think the problem may have been I was using an undefined variable higher in the schema, which probably speaks to the problem with `ensure_loaded` as it could result in misleading errors.

**Comment by benwilson512** [comment]
Unfortunately I'm sort of confused at this point. Is there a bug here?

**Comment by ssomnoremac** [comment]
I think so yeah but I'm sure. Maybe `ensure_loaded` or maybe something else somehow causes a misleading error message, in this case `CreateOpening is not available` when in fact it is available, but there happens to be some other error in that module that's not present in the trace. In my case, an undefined variable.

**Comment by ssomnoremac** [comment]
right so `Code.ensure_loaded?(type_module)` is returning false earlier than the undefined variable is caught

**Comment by ssomnoremac** [comment]
After doing some testing it seems that it isn't some other error that is causing `ensure_loaded?` to return false; instead, it seems to be random :(

File: lib/absinthe/schema/notation.ex [hunk 2]
```diff
@@ -1139,29 +1139,30 @@ defmodule Absinthe.Schema.Notation do
...
-    end)
-  end
+    {:__aliases__, _, root} = root_ast
 
+    root_module = Module.concat(root)
+    root_module_with_alias = Keyword.get(env.aliases, root_module, root_module)
+
+    for {_, _, leaf} <- modules_ast_list do
+      type_module = Module.concat([root_module_with_alias | leaf])
+      if Code.ensure_loaded?(type_module) do
```
bruce: @ssomnoremac What happens if you remove this `ensure_loaded?` check in the project where you're seeing errors?

ssomnoremac: @bruce I can't reliably reproduce the error unfortunately, but I'll try that and work with it all week and let you know if I get the error again, manually recompiling as much as possible.

bruce: Thanks; let me know if you find a clear reproducible case. I'll try to circle back when I have time later this week.

ssomnoremac: @bruce , been using this for well over a week now and haven't had any issues. Deleted the `_build` folder as well to check that. Hasn't failed on me yet.

bruce: Had you tried deleting `_build` before? I wonder if things got into a weird state somehow.

ssomnoremac: no, someone suggested I do that to test that removing `ensure_loaded?` wouldn't affect the initial build. But before, no, I wasn't deleting the folder.

benwilson512:
Ah. This feature is likely fine then, I'm gonna merge it in.

PR #449, opened 2017-11-17
Support for null values as nullable enum argument

terianil:
## Issue

Nullable enum arg throws an error for null value.

### schema and query definition
```elixir
enum :post_type do
  value :a
  value :b
end

(...)
field :get_posts_by_type, :posts_result do
  arg :type, :post_type

  resolve (...)
end
```

### query
```javascript
getPostsByType(type: null) (...)
```

### result
```javascript
{
  "errors": [
    {
      "message": "Argument \"type\" has invalid value null.",
      "locations": [
        {
          "line": 2,
          "column": 0
        }
      ]
    }
  ]
}
```


benwilson512:
Looks good, thanks!

PR #453, opened 2017-11-22
Update middleware guides

ashkan18:
We are missing guides for how to create and use Middlewares.

This PR updates existing `guides/middleware-and-plugins.md` to include some notes about what are Middlewares and how to use them with some examples. Hopefully this is helpful 😎


benwilson512:
Thank you!

PR #456, opened 2017-11-28
Absinthe.Blueprint.Input.Null: add alias for Absinthe.Blueprint

vderyagin:
This change adds an alias that is used in type definition, without it dialyzer complains about unknown types `Elixir.Blueprint.flags_t` and `elixir.Blueprint.Document.SourceLocation.t`.


bruce:
Thanks!

PR #463, opened 2017-12-04
Block Strings

bruce:
Supports `"""`-delimited block strings, as defined in facebook/graphql#327, using the implementation in graphql/graphql-js#926 as a reference (omitting SDL `description` features slated for v1.5).

- [x] Lexer
- Parser
  - [x] Base support
  - [x] Escape handling
  - [x] Indentation handling
- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v1.4.5
+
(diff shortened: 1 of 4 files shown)


bruce:
Blurgh, I guess `string:split/3` is an Erlang/OTP 20 thing. Going to switch that up for Erlang/OTP 19 compat.

bruce:
(Putting this back as WIP until I can address a warning and a replacement for Erlang/OTP's `string:prefix/2`.)

PR #465, opened 2017-12-05
Add support for defining meta with keyword lists

tlvenn:
I tried to add support to the field macro as well but could not make it work, not familiar enough with the inner working of the schema notation macro work.

### DIFF
--- a/lib/absinthe/schema/notation.ex
+++ b/lib/absinthe/schema/notation.ex
@@ -709,10 +709,31 @@ defmodule Absinthe.Schema.Notation do
     |> record_private!(:meta, key, value)
   end
 
+  @doc """
+  Defines list of metadata's key/value pair for a custom type.
+  """
+  defmacro meta(keyword_list) do
+    __CALLER__
+    |> recordable!(:meta, @placement[:meta])
+    |> record_private!(:meta, keyword_list)
+  end
+
+  @doc false
+  # Record private values
+  def record_private!(env, raw_owner, raw_keyword_list) when is_list(raw_keyword_list) do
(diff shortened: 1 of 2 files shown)


tlvenn:
@bruce could use some help to enhance the field macro to support the meta keyword list as well.

lib/absinthe/schema/notation.ex
```diff
@@ -709,10 +709,31 @@ defmodule Absinthe.Schema.Notation do
...
+  defmacro meta(keyword_list) do
+    __CALLER__
+    |> recordable!(:meta, @placement[:meta])
+    |> record_private!(:meta, keyword_list)
+  end
+
+  @doc false
+  # Record private values
+  def record_private!(env, raw_owner, raw_keyword_list) when is_list(raw_keyword_list) do
+    [owner, keyword_list] = Enum.map([raw_owner, raw_keyword_list], &Macro.expand(&1, env))
```
benwilson512: You actually gotta do deep expansion here, so I'd just use the `expand/2` helper function:

```
owner = expand(owner, env)
keyword_list = expand(raw_keyword_list, env)
```

tlvenn: But i am already using the expand macro ;)

tlvenn: For reference, it did not come up with this myself, this construct was already being used:

https://github.com/absinthe-graphql/absinthe/blob/master/lib/absinthe/schema/notation.ex#L715

tlvenn:
@benwilson512 any chance you can follow up on this PR ? Would be super useful to get support for inlined field meta. Right now I am forced to create `do...end` around my fields :(

Thanks in a lot in advance

benwilson512:
Merging into feature branch so I can poke at it.

PR #467, opened 2017-12-08
Add waring in docs about third param of batch function

astery:
Recently, I got bitten by passing whole resolution tree in this parameter, and received silently N+1 and drastically performance hit.


benwilson512:
Merging into feature branch for tweaks, thanks!

PR #468, opened 2017-12-09
Absinthe.Schema: Fix middleware/3 typespec

xtian:
This typespec causes a Dialyzer warning because `[]` is matched on in `__using__/1`. That match would never succeed if this callback always returned a non-empty list.


benwilson512:
Ah, the match for `[]` is from a time when an empty list used to be a valid result. It is no longer a valid result, so the typespecs are correct and the code is wrong.

PR #470, opened 2017-12-17
Inline Meta Keyword list

benwilson512:


test/lib/absinthe/schema_test.exs
```diff
@@ -342,8 +342,9 @@ defmodule Absinthe.SchemaTest do
       #Query type must exist
     end
 
-    object :foo do
+    object :foo, meta: [foo: "bar"] do
```
benwilson512: @tlvenn tada^

tlvenn: 🎉

PR #472, opened 2017-12-20
Stop requiring the app to start

lswith:
Our application has a large boot up process and this shouldn't be necessary to generate a schema.


lib/mix/tasks/absinthe.schema.json.ex
```diff
@@ -48,7 +48,6 @@ defmodule Mix.Tasks.Absinthe.Schema.Json do
   @introspection_graphql Path.join([:code.priv_dir(:absinthe), "graphql", "introspection.graphql"])
 
   def run(argv) do
-    Mix.Task.run("app.start", [])
```
benwilson512: I think we still probably want an `Application.ensure_all_started(:absinthe)` here.

lswith: this also works too 👍

PR #473, opened 2017-12-20
Fix URL in our-first-query.md

kevinastone:
The router in the tutorial embeds the GraphiQL in an `/api` scope.


benwilson512:
Thanks!

PR #476, opened 2017-12-31
Schema: Remove match against `[]` to fix Dialyzer error

xtian:
According to @benwilson512 in https://github.com/absinthe-graphql/absinthe/pull/468 this code is invalid.

Removing this `case` fixes the Dialyzer error:
```
lib/<app>/schema.ex:6: The pattern [] can never match the type [any(),...]
```

### DIFF
--- a/lib/absinthe/schema.ex
+++ b/lib/absinthe/schema.ex
@@ -137,14 +137,6 @@ defmodule Absinthe.Schema do
         middleware
         |> Absinthe.Schema.__ensure_middleware__(field, object)
         |> __MODULE__.middleware(field, object) # run field against user supplied function
-        |> case do
-          [] ->
-            raise """
-            Middleware callback must return a non empty list of middleware!
-            """
-          middleware ->
-            middleware
-        end
       end
 
       @doc false


benwilson512:
This is one of the issues with dialyzer. `middleware` is a user space function. Nothing stops the user from returning an empty list, so if they do, I want to give them a nice error. If they're using dialyzer then I suppose they it would provide them an error but, given that most aren't, we shouldn't remove this clause.

xtian:
@benwilson512 That's why I think the change in #468 is correct.

The return value of `middleware/3` could be an empty list so any code that calls it assuming the list will be non-empty may break. `__do_absinthe_middleware__/3` on the other hand will always return a non-empty list (or raise an exception).

benwilson512:
The return value of `middleware` could be literally anything, but if we do `:: term` then it fails to be useful documentation.

Perhaps we want `[Absinthe.Middleware.spec, ...] | term` as the return value?

xtian:
Actually, is the `[]` arm of this case reachable? Won't [`__ensure_middleware__/3`](https://github.com/xtian/absinthe/blob/282326a85a01b804492129fba910ed029e32154e/lib/absinthe/schema.ex#L179-L184) handle the empty list case?

benwilson512:
`__ensure_middleware__` is called prior to the user supplied middleware https://github.com/xtian/absinthe/blob/282326a85a01b804492129fba910ed029e32154e/lib/absinthe/schema.ex#L138-L139

xtian:
Ah yeah, whoops! Well I'm not sure what the best thing is here. I agree that the callback documentation should be as instructive as possible. Would it make sense to move this `case` out of the macro and into the library where the middleware is utilized?

benwilson512:
Yeah it's a bit tricky. I think the best route forward is to change the typespec to `[Absinthe.Middleware.spec, ...] | term` and then change the case match to do `_ ->`

Feel free to push another commit to this branch and I'll re-open the PR.

PR #478, opened 2018-01-03
Remove ExSpec

bruce:
This is the first set of clean-up PRs meant to prepare for work on v1.5; this cleans and modernize our test suite a bit. We used ExSpec early on in the project (before ExUnit had things we wanted), and the main purpose here is removing that old dependency and the hacks put in place to make it play nice with newer releases of ExUnit.

- Remove the `ex_spec` dependency
- Restructure test contents to use `describe` vs `context` and `test` vs `it` (at only one level of nesting)
- Do some mild reorganization:
  - Move type-specific tests to `execution/arguments/<type>_test.exs`
  - Move schemas only used by one test file from `support` to the test file
- Assertions:
  - Add `assert_data`, removing a lot of repetitive match code
  - Add `assert_error_message`, also removing a lot of match code
  - Add `assert_error_message_lines` for more complex (multi-newline error messages)
  - Always make assertions available; automatically imports the newly-created `Absinthe.Case.Assertions` when you use `Absinthe.Case`


bruce:
This also establishes the following convention for GraphQL tests:

```elixir
@schema TestSchemaUsed

@graphql """
query ($foo: String!) {
  bar(foo: $foo) {
    name
  }
}
"""
test "querying baz results in quux" do
  assert_data %{"bar" => %{"name" => "Quuz"}},
    run(@graphql, @schema,
      variables: %{"foo" => "Baz"})
end
```

Working rules:

- Always use the operation type in the GraphQL document (unless you're explicitly testing the simple query document behavior)
- Don't provide an operation name (unless you're explicitly testing something relating to operation name behavior)
- If a test module uses a single schema, set `@schema`
- In general, set a `@graphql` module attribute before the test with the GraphQL document; there may be cases you don't because you run multiple docs in a single test (but break them up)
- Prefer `assert_data`, `assert_error_message`, and `assert_error_message_lines` to the more noisy `assert_result`, when possible
- Consider shown here is more instructive rather than prescriptive; especially where it may differ with `mix format`

bruce:
Travis CI failure seemed to just be a spurious HTTP connection issue with hex.pm. Merging. 🚀

PR #482, opened 2018-01-08
Refactor and Reorganize Tests

bruce:
This pull request refactors and reorganizes Absinthe's test suite, and is part of our preparation to start working on v1.5.

## Motivation

Our test suite has grown rather organically over the last 2+ years. While it's fairly substantial (at ~760 tests), this has led to:

- A massive `absinthe_test.exs`—a kitchen sink of tests that didn't seem to fit anywhere else
- End-to-end GraphQL tests sprinkled (when not in `absinthe_test.exs`) in whatever location felt vaguely appropriate (eg, along with more focused unit tests for modules)
- Duplication and (very likely) missed edge cases due to the confusion caused by the above.

## Status

Note: This pull request is still a _work in progress_ and is estimated to be completed on 2018-01-08.

- [x] Remove unnecessary `lib/` subdirectory in `test/`
- [x] Pull common test schemas into a more organized set of fixtures
  -  ✍️  New convention established; put them under `test/support/fixtures` (eg, `Absinthe.Fixtures.ContactSchema`)
- [x] Make writing phase tests easier and less verbose
  - ✍️ Created `Absinthe.PhaseCase`, applied it
- [x] Make writing validation-specific phase tests easier and less verbose
  - ✍️ Created `Absinthe.ValidationPhaseCase`, applied it
- [ ] Extract end-to-end GraphQL integration tests into a file-based system so contributors can easily drop in a GraphQL file and an Elixir file describing scenarios (settings and expected responses).
  - ✍️  Added `Absinthe.IntegrationCase` (substantial `@moduledoc` in place), used by `test/absinthe/integration_test.exs`, extracted integration tests from `absinthe_test.exs` (now gone!) to `test/absinthe/integration/**/*.{graphql,exs}`. Still need to:
    - [ ] Extract from other `_test.exs` files
    - [ ] Finalize integration test directory structure (will probably do away with separate `execution` and `validation` directories, as it's handy to test success/failure together as discrete scenarios against the same GraphQL document)
- [ ] Do a pass renaming tests to explicitly call out (possibly with `describe`) the functions they are testing

## Ideas

- It might be nice to make it easy for projects that use Absinthe to use our test cases in their own tests.


benwilson512:
@bruce where do we stand on this? Is it possible to merge what's here then do the rest in another PR? This is going to start blocking #503 because I want to start switching existing test clauses to use the new schema builder.

Issue #439, opened 2017-11-10
Feature Discussion: Runtime-defined, dynamic, mutable schema


cgiffard:
Hey Absinthe Team! Hope you're all well. :)

### Context

I'm working on a large greenfields API development effort with my team members @nwinch and @indieisaconcept over at @autopilothq.

We were foolhardy enough to decide that the premise of this API should be the dynamic definition of business concerns — (i.e. — wanna track/store/CRUD/manipulate something? Create a type for it!) — and so we currently have a REST-only API which is both strongly and dynamically typed. It was always our intention to expose this internal schema to clients via GraphQL, and this whole time we've had our fingers tightly crossed that we wouldn't have to implement our own GraphQL library in order to do so.

### Technical Spike

In order to test the feasibility of using absinthe for this feature, we did a brief spike. We ran into some problems with some assumptions that Absinthe makes, but were surprised and delighted to discover that not /all/ that much lies in the way of a completely dynamic schema, at least from a purely "just-get-it-done" perspective.

A number of problems stood out:

* The focus on a compile time DSL significantly improves the ease of definition of static schemas, but hides implementation, making dynamic definition (expectedly) difficult. We had to do quite a bit of reading of code (I must say though — your inline docs are wonderful — really appreciated that!) and exploring of macros and defined methods in order to mimic the signatures and return values of all the `__absinthe_*__` methods an Absinthe schema contains.
  * Perhaps it's down to a lack of research on our part (though we did look around for this) but nice public helpers for generating and validating the various required structs would be enormously helpful.
* The use of a process dictionary to cache type information is obviously beneficial from a performance perspective, but defeats our ability to redefine the schema in a reliable way. It would be nice to be able to control this behaviour (we ultimately came up with a solution to defeat this cache, but it isn't wonderful — more below.)
* When absinthe calls the various type introspection methods on the schema (`__absinthe_type__`, etc.) it does so absent any context information from the request. This makes sense given a context-sensitive dynamic schema was never a requirement in the past, but makes it difficult for us to deliver a different schema for each authorised client of our API.

### Our nasty hack "solution"

After working out how to mimic the macro defined schema (straightforward enough) we needed to work out how to defeat the process cache. Originally we patched `absinthe` itself to remove this cache altogether, but this was eventually superseded by a different solution.

Rather than embark on a broad-scope refactor of absinthe to support passing contextual information about a request to the schema, in order to just get on with it and demonstrate feasibility we decided to
[...]

PR #483, opened 2018-01-09
type : permit `strings` to be used as identifiers for types

indieisaconcept:
Hi Absinthe Team,

Following on from the issue ( #439 ) raised by @cgiffard last year I'd like to propose the changes outlined below for inclusion.

Cheers.

--------------------

**Context**

For types which are dynamically generated at runtime the use of an `atom`
as an identifier is not suitable as it could eventually lead to exhausting
the `atom` table limit, resulting in an application crash.

**Changes**

- referenced_type lookups are no-longer restricted to `atoms`
- type expansions extended to support `strings` lookups for `lists`
- updated schema notation to support string identifiers

**Ref**

* #439

### DIFF
--- a/lib/absinthe/schema.ex
+++ b/lib/absinthe/schema.ex
@@ -525,13 +525,13 @@ defmodule Absinthe.Schema do
   List all implementors of an interface on a schema
   """
   @spec implementors(t, Type.identifier_t | Type.Interface.t) :: [Type.Object.t]
-  def implementors(schema, ident) when is_atom(ident) do
+  def implementors(schema, %Type.Interface{} = iface) do
+    implementors(schema, iface.__reference__.identifier)
+  end
+  def implementors(schema, ident) do
     schema.__absinthe_interface_implementors__
     |> Map.get(ident, [])
     |> Enum.map(&lookup_type(schema, &1))
-  end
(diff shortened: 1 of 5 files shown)


bruce:
Some things to check here:

- Ensure this works with `lookup_type`
- Ensure this works with absinthe_relay’s `ParseIDs` middleware, which IIRC uses an is_atom check to decide how to look up a type _name_

indieisaconcept:
@bruce thanks for the feedback. I'll review.

benwilson512:
Apologies for the delay on reviewing this. While the total changes are small, the implications are reasonably large, and so I need to make sure that the changes are going to fit with the approach we want to take in Absinthe 1.5's schema refactoring.

indieisaconcept:
@benwilson512 thanks for the update. We've now been using a forked version of Absinthe with minimal updates for about a month and a half now. It's working quite well.

We've since discovered some additional minor changes in relation to atoms vs strings. Would you prefer I hold off on pushing these changes until you have had a change for a more formal review?

benwilson512:
This will all get subsumed under the 1.5 schema refactoring.

indieisaconcept:
Thanks for the update.

PR #489, opened 2018-01-17
Formatter export

aquarhead:
Since Elixir 1.6.0 just released, exporting these macros should help all projects using Absinthe adopt a consistent format.

The export list contains all macros defined by `Absinthe.Schema.Notation` except `list_of` and `non_null` which feels better with `()`. The list also includes top-level macros `query`, `mutation` and `subscription`, although in my project the current (elixir-1.6.0) formatter doesn't add parens around these, I think explicitly indicating these is better.

With this configuration my current `schema.ex` stays mostly the same, except the formatter adds new lines between 1-liner `field` definitions and those with a `do`-block. [As seen in this commit](https://github.com/aquarhead/naglfar/commit/cad924532718c1cb3509ba4b54cb5423122449ef#diff-eea2debc9a860680cd202604ae6f0cbf)

Test and package build output:

```
◊ mix test | grep failures
759 tests, 0 failures

◊ mix hex.build | grep .formatter.exs
    .formatter.exs
```


benwilson512:
Looks good, thanks!

PR #493, opened 2018-01-31
Add value/1 to formatter locals_without_parens

binaryseed:
Adding `value/1` to the list, I think it's common to have Enum values w/o a description...


benwilson512:
Thanks!

PR #494, opened 2018-02-02
Add /.elixir_ls to .gitignore

irrigator:
[The Elixir Language Server](https://github.com/JakeBecker/elixir-ls) which powers great plugins for VS Code and Atom would generate some analysis files in ```.elixir_ls``` automatically. Would it be okay to add this directory to .gitignore file?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 /cover
 /deps
 /doc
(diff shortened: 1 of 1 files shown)


benwilson512:
Thanks!

PR #496, opened 2018-02-07
Flesh out the schema guide's "Importing Types" section

axelson:
I ran into trouble trying to follow that guide so I pulled in the bare minimum from the "Importing Types" guide and added a link to the full guide.

In my opinion I'd like to see the "Writing Schemas" guide fleshed out with a section about importing fields as well as a recommended approach or two to structuring a full-blown schema (especially in the file-system).


benwilson512:
Ah excellent, thank you for all the doc contributions!

axelson:
You're welcome! Glad to help out!

PR #497, opened 2018-02-08
Docs: recommend Absinthe.Plug.put_options/2

axelson:
Makes sense to recommend `Absinthe.Plug.put_options/2` over `Plug.Conn.put_private/3` to reduce coupling of client code to Absinthe internals.


benwilson512:
Thanks!

axelson:
Thank you!

PR #500, opened 2018-02-09
Docs: Improve the documentation of Absinthe.Resolution

axelson:
A majority of the @moduledoc appears to accidentally have been in the @typedoc for arguments instead.

Also added some basic info on the `:private` key and an example of updating the context of a Resolution.

Reference to slack discussion on 2018/02/08:
https://elixir-lang.slack.com/archives/C0PR49P4P/p1518144512000216


benwilson512:
Thanks!

PR #501, opened 2018-02-16
[Draft] Adding docs for using dataloader

smoak:
This guide did not exist. It focuses on using Dataloader.KV as that was
lacking in docs.

Currently a draft and looking for feedback as I continue to update it.


smoak:
This will eventually close #443

benwilson512:
Thank yoU!

Issue #222, opened 2016-12-07
Is it possible to import an external schema?


liveresume:
I was wondering if it's possible to import a preexisting GQL Schema for Absinthe to use in place of defining it again using elixir.

Does it even make sense to use Absinthe if this is the requirement?

bruce:
Importing a schema IDL (which is the GraphQL-based string representation) is planned for v1.3, whose development is underway; expect to see an alpha in January.

Keep in mind logic will still have to be implemented in Elixir to resolve fields. In the meantime it may make sense to give the Elixir DSL a shot.

liveresume:
Been good so far.  Waiting patiently.

rawkode:
@bruce This would be awesome. If you need some help, please point me to some issues and I can pick them up :+1:

benwilson512:
Update: @bruce has made significant progress on a blueprint based method of constructing schemas. This method would be able to import IDLs.

Unfortunately 1.3 is going to ship with out it, as there is a fair bit of work to be done there still and 1.3 includes features like middleware that have been a long time coming and are more overall more pressing.

rawkode:
Hi @bruce, are you still working on this? My offer to help is still valid, I'd be keen to see this integrated into Absinthe :+1:

bruce:
This just got pushed by subscriptions, which ended up being higher priority. This is part of a larger rework of the entire schema definition mechanism (supporting use of its own pluggable phases), so I think it's a bit bigger than we could expect anyone else to reasonably bite off. This rework _is_ my main priority after this beta goes to stable, so hopefully you'll see some progress soon.

rawkode:
Hey @bruce, is this still on the backlog? :smile:

benwilson512:
This is indeed at the center of the set of features planned for Absinthe 1.5, you can track progress here: https://github.com/absinthe-graphql/absinthe/pull/503

PR #503, opened 2018-02-17
SDL Support

benwilson512:
- [x] Absinthe.Schema.Notation => Blueprint Tree
- [x] SDL schema => Blueprint Tree
- [x] Blueprint Tree => compiled / usable schema

This branch is wildly experimental right now, do not try to use.


benwilson512:
Unfortunately, the current implementation has the compilation time issue we ran into with early versions of Absinthe. The experimental implementation does:

```
@absinthe_blueprint Notation.put_attrs(@absinthe_blueprint, ...)
```

For ~200 line schemas it takes ~2 seconds to compile. Not great, but workable. The problem is it's non linear. A schema twice the size at ~400 lines takes 12 seconds to compile. 700 lines takes 26 seconds, and 1000 lines takes 45 seconds.

Basically, we can't do stuff in the module body. This may be OK though, since we're doing things differently this time. If we ignore the problem of anonymous functions for a second (I've got an idea for those) I think we can simply push values into an ets table. Then we do a single `@absinthe_blueprint` with the blob at the bottom.

lib/absinthe/phase/schema/build.ex
```diff
@@ -0,0 +1,40 @@
...
+    body = [
+      types,
+      quote do
+        def __absinthe_type__(_) do
+          nil
+        end
+      end
+    ]
+
+    Module.create(module_name, body, Macro.Env.location(__ENV__))
```
benwilson512: @bruce the very beginning of blueprint => compiled schema work.

collegeimprovements:
Thanks a lot @benwilson512 for SDL features. What can we expect? Does it mean it will include directive like prisma in nodejs world ?

benwilson512:
It will yes.

bruce:
@benwilson512 Is it fair to say this should be closed in favor of `switch-to-new-schema`?

benwilson512:
The only remaining functionality item at this point I think is

- [ ] Subscription triggers

Then of course there's all the validation rules we need to recreate, and the error system.

benwilson512:
OK. I'm gonna merge this in at this point, everything is fully functional. The error system still needs to happen, but I'll handle that in another PR. The next major task is actually over in Absinthe.Relay to make sure that we can get it to work with all of these changes.

binaryseed:
closes #222

PR #505, opened 2018-02-21
Formatter is missing config for parenthesis

prikha:
Hello guys, 

there is an issue with current `.formatter.exs`. In our project we have: 

```
[
  import_deps: [:absinthe]
]
```

but constructions, as follows, are not properly formatted: 

```elixir
    import_fields :slug_queries
    import_fields :label_queries
```

and 

```elixir
    middleware &build_payload/2
```

arguments are wrapped into parenthesis. Looks like we should be consistent and add proper values to config.

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -18,6 +18,7 @@
       field: 3,
       field: 4,
       import_fields: 2,
+      import_fields: 1,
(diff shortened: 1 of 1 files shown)


benwilson512:
Thanks!

PR #509, opened 2018-02-23
Adding handling for exceptions in resolver error status tuples

bundacia:
If a resolver function returns an error status tuple where the second element is an elixir exception (eg. `{:error, %RuntimeError{}}`) instead of displaying this error in the graphql result you'll actually get the following type of error:

> ** (Protocol.UndefinedError) protocol Enumerable not implemented for %RuntimeError{message: "Error"}. This protocol is implemented for: Date.Range, File.Stream, Function, GenEvent.Stream, HashDict, HashSet, IO.Stream, List, Map, MapSet, Range, Stream

While this probably isn't a super-common case, it seems counter-intuitive that an exception in this context would not work, and seems like something that would make sense to support.

For more background [see this blog post](https://tech.decisiv.com/handling-elixir-exceptions-in-absinthe-9656a9108c2d)

### DIFF
--- a/lib/absinthe/phase/document/execution/resolution.ex
+++ b/lib/absinthe/phase/document/execution/resolution.ex
@@ -308,6 +308,9 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
     end
   end
 
+  defp split_error_value(%{__exception__: true} = exception) do
+    {[message: Exception.message(exception)], []}
+  end
   defp split_error_value(error_value) when is_list(error_value) or is_map(error_value) do
     Keyword.split(Enum.to_list(error_value), [:message])
   end
--- a/mix.lock
+++ b/mix.lock
@@ -1,10 +1,12 @@
-%{"benchfella": {:hex, :benchfella, "0.3.5", "b2122c234117b3f91ed7b43b6e915e19e1ab216971154acd0a80ce0e9b8c05f5", [:mix], [], "hexpm"},
+%{
+  "benchfella": {:hex, :benchfella, "0.3.5", "b2122c234117b3f91ed7b43b6e915e19e1ab216971154acd0a80ce0e9b8c05f5", [:mix], [], "hexpm"},
   "dataloader": {:hex, :dataloader, "1.0.0", "d461598ddaa822d32bd7aca59f4b592fb6c8b21d74723d9ce9a5b8e32c301a4d", [:mix], [{:ecto, ">= 0.0.0", [hex: :ecto, repo: "hexpm", optional: true]}], "hexpm"},
   "decimal": {:hex, :decimal, "1.4.1", "ad9e501edf7322f122f7fc151cce7c2a0c9ada96f2b0155b8a09a795c2029770", [:mix], [], "hexpm"},
-  "dialyze": {:hex, :dialyze, "0.2.1", "9fb71767f96649020d769db7cbd7290059daff23707d6e851e206b1fdfa92f9d", [], [], "hexpm"},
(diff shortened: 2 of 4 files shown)


benwilson512:
This is an interesting situation. There's definitely a convenience to being able to simply return the errors.

There's also a risk though. Exception messages can easily contain sensitive information, which we're now regurgitating back to the user. My expectation would be that if a field is unable to resolve and you wanted to catch those errors the user would just get `"Internal Error"`, not the exception message. The exception message is for the system developers / administrators, not the users.

@bruce thoughts?

bundacia:
That's a good point. I'm a little torn. I agree that in general you don't want to just expose exception messages to the user, but in this case a middleware is explicitly returning an error tuple, not raising an exception. Is it safe enough in that case to assume that the intent is that be treated the same way that a map or string, etc would be treated?

bruce:
Definitely interesting.

Being that exceptions can be fairly customizable, I wonder if this is better as a documented pattern (@benwilson512 has been organizing an effort to capture ideas in the wiki, for instance) or as a third-party middleware package that could provide a nice API for how errors are converted into resolution results. I'm not sure it's straightforward yet to be a built-in part of the existing resolution logic or included as a standard middleware (as we do with certain scalar types).

bundacia:
> ... or as a third-party middleware package that could provide a nice API for how errors are converted into resolution results.

We've got something like that that we're using internally. I've got approval to open-source it, just need to push it up to github. That might be a good place to start a discussion.

benwilson512:
I think I'm going to close this in favor of documenting the ability to do this approach if desired. I'm still unconvinced that this is a safe pattern, and I don't think first class support for it is particularly necessary even if you do want to do it.

garthk:
Just ran into this one. Did we ever document the ability, @benwilson512? Did you ever publish the middleware, @bundacia? If not, all hints appreciated.

benwilson512:
@garthk As far as hints go the general idea would be to do something like:

```
  Enum.flat_map(middleware, fn
    {{Absinthe.Resolution, :call}, _} = resolver ->
      [resolver, {{MyApp.Whatever, :exceptions_to_errors}, []}]
    other -> [other]
  end)
end
```

where `:exceptions_to_errors` is a function that looks through the errors key on the resolution struct for errors that are exceptions, and does the `{[message: Exception.message(exception)], []}` thing to them.

bundacia:
We did open source a package that adds some extra error handling for this type of situation. See https://github.com/Decisiv/blunder and https://github.com/Decisiv/blunder-absinthe

We're doing a little more with those packages than just handling runtime errors. Blunder introduces a new type for representing errors, and then `blunder-absinthe` is a middleware that normalizes all different types of error responses from resolvers (including returned error tuples and thrown exceptions). Might be more than what you're looking for @garthk, but let me know if it's helpful.

garthk:
@benwilson512 as well as `{{Absinthe.Resolution, :call}, fun}` where `fun` is a literal external function e.g. `&Schema.foo/3`, I've also seen `{Absinthe.Resolution, fun}` with inline functions e.g. as returned by `Absinthe.Resolution.Helpers.dataloader/3`. Any reason I should avoid trying to interfere with the latter?

garthk:
@bundacia I love the name `Blunder`. Might use it just because that. 

I'm amused by `Blunder.Absinthe` replacing all of the invocation machinery to rescue from resolvers before crashing out the whole blueprint. It's like my #757, but done from the outside and also replacing the sync calls. Given #757 is doomed, reckon we could team up to extract one set of call hooks to share for both recovery and trace propagation? I'd be more comfortable filing another Absinthe PR after demonstrating a stable API covering a few use cases and supporting a handful or more users.

Also noticed you'd hadn't done anything in the middleware to wrap calls to `Dataloader.run/1`, only to see you'd defined your own wrapper around `Dataloader.KV`. Don't blame you at all. I've been looking into wrapping that from the other end, and it's tricky.

PR #515, opened 2018-03-05
update doc block sample: make id a non-null argument

steffkes:
otherwise elixir will complain about `no function clause matching in anonymous fn/2` where the passed arguments look like: `fn(%{}, %{id: "foo"}, #Absinthe.Resolution<...>)`

### DIFF
--- a/lib/absinthe/schema.ex
+++ b/lib/absinthe/schema.ex
@@ -43,9 +43,9 @@ defmodule Absinthe.Schema do
     field :item, :item do
 
       @desc "The ID of the item"
-      arg :id, type: :id
+      arg :id, type: non_null(:id)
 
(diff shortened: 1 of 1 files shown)


benwilson512:
Both the 2 arity and 3 arity functions are still supported. The issue here is that the `id` arg is nullable, which means if it isn't passed in you'll have a no function clause matching. The change from 2 to 3 arity doesn't affect that. I think a better fix would be changing the argument to be non null.

steffkes:
your comment was spot-on, thanks! i was confusing myself while doing to many changes at once, leading to a bad PR, sorry for that. 

I've added another commit on top to revert the first change back and change `id` to a non-null argument as suggested.

benwilson512:
Thanks!

PR #516, opened 2018-03-05
WIP: supplemental PR related to dataloader PR

jfrolich:
see https://github.com/absinthe-graphql/dataloader/pull/21


jfrolich:
Btw this is a small change, much of the changes are due to the code-formatter. How do I set the correct preferences?

benwilson512:
@jfrolich the code base does not yet use the formatter, you'll want to ensure it isn't run on any PR you submit for the time being.

jfrolich:
Understood, will clean it up later. After some experimentation I'm pretty pleased with a solution, almost done. Will add some examples later.

binaryseed:
This PR is a bit stale, nearing 1 year since last touch. Closing for now..

jfrolich:
That's fine. Any ideas or progress around fixing the underlying issue with dataloader (that dataloader is usable in contexts instead of only in the resolvers)? I was playing with an idea of trying to fix this using erlang processes instead of lazy data structures.

PR #519, opened 2018-03-11
[WIP] Switch to new schema

benwilson512:
#503 is big enough as is, gonna break it into several PRs.


benwilson512:
@bruce note that I'm moving development back to blueprint-schema-notation since this has largely accomplished the switch and it didn't make sense to have two PRs up for the same thing at this point.

PR #522, opened 2018-03-15
Dataloader bug, regression test

jfrolich:
Hereby the regression test

*note:* this small change to Dataloader.KV needs to be merged to make it testable using Dataloader.KV: https://github.com/absinthe-graphql/dataloader/pull/23.


benwilson512:
@jfrolich can you merge in master? I think these commits are already there, not entirely sure why this PR diff isn't reflecting that.

jfrolich:
Yeah the merged commit was changed while committed to master. Rebased, now it's just the regression test.

benwilson512:
Ah, I think I have this in the test case right beneath. Thanks!

PR #524, opened 2018-03-19
Add log_mode config to allow for single-line request logging

seddy:
This adds a configuration which allows developers to switch the
(potentially verbose) full graphql query into a single line. The idea is
that this allows for more compact logging in production environments,
which is specifically useful for log aggregation tools

---

I've added a test which covers the current state of the logger output, but not one for the alternate configuration. The main reason is that I'm not sure about best way to do that when it depends on an `Application.get_env` call - if there's a nice way of doing that I'm all ears :smile:.  It was useful to write the code against (as I switched the config to `single_line` locally to prove it out) and provides a regression which is why I left it in.

I'm also not 100% convinced on my use of the default `:multi_line` atom, as it's not actually used... Would this be better if the config option was something like `multi_line` and was a boolean?

### DIFF
--- a/lib/absinthe/logger.ex
+++ b/lib/absinthe/logger.ex
@@ -2,6 +2,7 @@ defmodule Absinthe.Logger do
   @default_log true
   @default_filter_variables ~w(token password)
   @default_pipeline false
+  @default_log_mode :multi_line
 
   @moduledoc """
   Handles logging of Absinthe-specific events.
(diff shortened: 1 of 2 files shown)


benwilson512:
This would also remove all the new lines from the GraphQL document, which is not guaranteed to produce the same document. So for example
```
{
foo
bar
}
```
is a valid document and if you remove the new lines you get
```
{foobar}
```
which probably isn't.

Really though I don't think that this is something Absinthe needs to solve. If you want to guarantee that your log entries are on a single line you should use a logging backend that emits things how you want, possibly with an escaped string.

seddy:
Hmm, that's a fair point @benwilson512 - what would be the best way of going about that? Would a custom plug work for that? 

When you suggest an escaped string, do you mean that something like `{\nfoo\nbar\n}` (using your example)?

benwilson512:
Yeah that's the output. And no this isn't a plug issue either, this is a question of how you use logger. See https://hexdocs.pm/logger/Logger.html#module-custom-backends

seddy:
Ok, I'll check it out, thanks @benwilson512

PR #528, opened 2018-03-20
Fix documentation on parse/1

dmarkow:
Per #275/#276, parse/1 needs to return `:error`, not `{:error, reason}`, otherwise a `no case clause` error is raised.


benwilson512:
Thanks!

PR #530, opened 2018-03-24
fix input output checking

### LINKED ISSUE #507: Introspection compiles but returns invalid schema at runtime if you include a normal object in an input object
### Environment

* Elixir version (elixir -v): 1.6.1
* Absinthe version (mix deps | grep absinthe): 1.4.7 
* Client Framework and version (Relay, Apollo, etc): Any client

### Expected behavior

Invalid input objects raise a compile time, or at least a runtime error.

### Actual behavior

Schema with invalid input object does not raise any error, but the generated introspection schema is invalid. Different clients fail in different ways. `graphql-cli` fails like this:

`Introspection must provide input type for arguments.`

### Steps to reproduce

Create an input object like this:

```elixir

object :my_object do
  field :anything, :string
end

input_object :some_input do
  field :bad, :my_object
end

```

benwilson512:
Fixes #507

### DIFF
--- a/test/support/case/assertions/schema.ex
+++ b/test/support/case/assertions/schema.ex
@@ -20,6 +20,8 @@ defmodule Absinthe.Case.Assertions.Schema do
         load_schema(schema_name)
       end
 
+    err |> IO.inspect()
+
     patterns
     |> Enum.filter(fn pattern ->
       assert Enum.find(err.details, fn detail ->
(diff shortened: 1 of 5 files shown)


corck:
@benwilson512  Any plans to release the changes into a new version, just ran into the same problem.

test/support/fixtures/dynamic/invalid_input_types.exs
```diff
@@ -1,19 +1,16 @@
-defmodule Absinthe.Fixtures.InvalidInputTypesSchema do
```
jessedijkstra: Don't you have the file names the other way around now? InvalidOutputTypesSchema is in the invalid_input_types.exs file and the other way around.

test/support/case/assertions/schema.ex
```diff
@@ -20,6 +20,8 @@ defmodule Absinthe.Case.Assertions.Schema do
         load_schema(schema_name)
       end
 
+    err |> IO.inspect()
```
jessedijkstra: Forgot a debug statement?

PR #536, opened 2018-03-27
make sure import_type always create compile-time dependency

### LINKED ISSUE #535: Schema validation is not performed when changing child schema and multiple modules syntax is used to import it
### Environment

* Elixir version (elixir -v): Elixir 1.6.2 (compiled with OTP 20)
* Absinthe version (mix deps | grep absinthe): * absinthe 1.5.0-dev (../../absinthe) (mix) (current HEAD, a0e2bbf)
* Client Framework and version (Relay, Apollo, etc): -

### Expected behavior

Compilation fails with error when schema is invalid

### Actual behavior

Compilation can be successful even when schema is invalid.

lib/schema.ex:
```elixir
defmodule App.Schema do
  use Absinthe.Schema

  import_types(App.A)
  import_types(App.{B, C})

  query do
    import_fields :a_queries
    import_fields :b_queries
    import_fields :c_queries
  end
end
```

lib/a.ex:
```elixir
defmodule App.A do
  use Absinthe.Schema.Notation
  object :a_queries do
    field :a, :integer do
      resolve fn _, _ -> {:ok, 1} end
    end
  end
end
```

lib/b.ex:
```elixir
defmodule App.B do
  use Absinthe.Schema.Notation
  object :b_queries do
    field :b, :integer do
      resolve fn _, _ -> {:ok, 2} end
    end
  end
end
```

lib/c.ex:
```elixir
defmodule App.C do
  use Absinthe.Schema.Notation
  object :c_queries do
    field :c, :integer do
      resolve fn _, _ -> {:ok, 3} end
    end
  end
end
```

Run `mix compile`, it will compile the schema with no issues (expected behavior for now).

Now change lib/c.ex to:
```elixir
defmodule App.C do
  use Absinthe.Schema.Notation
  object :c_queries do
    field :c, :integer do
      resolve fn _, _ -> {:ok, 3} end
    end
  end

  # duplicate object!
  object :b_queries do
    field :b, :integer do
      resolve fn _, _ -> {:ok, 5} end
    end
  end  
end
```

Run `mix compile`, it will successfully recompile only `lib/c.ex` -  not expected behavior.
Running `mix compile --force` will obviously yield an error.

azhi:
do this by running Macro.expand on an alias AST node, which should be
equivalent of referencing module on subject module scope

compile-time dependency will always recompile root schema when
child schema is changed, and we need this to run schema validations on
every change

Fixes #535 
Again, I'm not sure how to test this in a suite, let me know if you have an idea.

UPD: added a small test to ensure scoped references are processed correctly inside {}

### DIFF
--- a/lib/absinthe/schema/notation.ex
+++ b/lib/absinthe/schema/notation.ex
@@ -1241,7 +1241,7 @@ defmodule Absinthe.Schema.Notation do
   ```
   import_types MyApp.Schema.Types
 
-  import_types MyApp.Schema.Types.{TypesA, TypesB}
+  import_types MyApp.Schema.Types.{TypesA, TypesB, SubTypes.TypesC}
   ```
   """
   defmacro import_types(type_module_ast) do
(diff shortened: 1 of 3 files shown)


benwilson512:
hey @azhi this is fantastic thank you. I'm gonna merge this in and then work on a test case.

PR #537, opened 2018-04-04
Fix complex default value InputType introspection

### LINKED ISSUE #490: Introspecting complex default values raises Protocol.UndefinedError
hi guys ~

i am using `default_value` in `input_object` ,  it *worked* , but there some error in console when i refresh the schema in graphql-playground ..

```elixir
input_object :article_filter do
    field(:first, :integer, default_value: 20)
    field(:order, :sort_order, default_value: :desc)
    field(:tag, :string, default_value: :all)
  end

field :posts, non_null(list_of(non_null(:post))) do
    arg(:filter, :article_filter, default_value: %{first: 20})
    resolve(&Resolvers.CMS.posts/3)
end
```

image

image

error detail:
```elixir
Request: POST /graphiql
    ** (Protocol.UndefinedError) protocol String.Chars not implemented for %{first: 20}. This protocol is implemented for: Atom, BitString, Date, DateTime, Decimal, Ecto.Date, Ecto.DateTime, Ecto.Time, Float, Integer, List, NaiveDateTime, Postgrex.Copy, Postgrex.Query, Postgrex.Stream, Time, URI, Version, Version.Requirement

```

binaryseed:
WIP

This PR is an attempt to fix the bug that exists when introspecting a "complex" input value, ie: one that doesn't respond nicely to `to_string`...

closes #490


binaryseed:
The code I have now does a hacky thing that gets close...

The problem is that I don't have a really solid way to turn the input `%{fancy_value: "qwerty"}` into the GraphQL representation: `{fancyValue: "qwerty"}`

binaryseed:
I added a naive encoder to get the example test passing, but I don't think this is the right solution as-is

binaryseed:
Any thoughts on this @bruce or @benwilson512 ?

bruce:
I’ll give this a look this weekend. Thanks, @binaryseed!

binaryseed:
I found a simple way to re-use most of the code that existed, just calling it recursively to render input objects!

There may be other cases like lists that I might need to check out

steffenix:
Any progress on that?

binaryseed:
I updated this, it was rendering all the things properly, so I validated that in a test & cleaned up the code.

At this point I suspect that master has moved quite far ahead into `1.5`, if we were to merge this would you make sure it wound up in the `1.4` branch @benwilson512 ?

binaryseed:
Sweet, I used `adapter.to_external_name`.

Also I found the `unwrap:` option and used that explicitly so we can ensure that we properly handle any type of value inside a list...

@benwilson512 any other possible `Absinthe.Type.*` structs that I might need to handle?

stef-otpyrc:
Is this for the next version?

steffenix:
Is this for the next version?

benwilson512:
This looks great, thank you!

PR #540, opened 2018-04-19
Introduce phase for adding document context

### LINKED ISSUE #539: Add context values after compile phases
### Environment

* Elixir version (elixir -v):
```
Erlang/OTP 20 [erts-9.3] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:10] [hipe] [kernel-poll:false]
Elixir 1.6.4 (compiled with OTP 19)
```
```
* absinthe (Hex package) (mix)
  locked at 1.4.10 (absinthe) 9f8d0c34
* absinthe_plug (Hex package) (mix)
  locked at 1.4.3 (absinthe_plug) 2d7f445b
```

plain React 16.3.1

### Expected behavior

Context values to be passed in to resolvers for pre-compiled query documents.

### Actual behavior

Context values are passed only when the default document provider is used. For pre-compiled docs, context is always %{}

The problem is due to the fact that context is passed on to Blueprint execution context in phase `Absinthe.Phase.Blueprint` as:
```
    context = Map.merge(blueprint.resolution.context, options[:context] || %{})
    blueprint = put_in(blueprint.resolution.context, context)
```

For pre-compiled documents, context values (like current logged-in userID) are not available at compile stage. So, we need another phase to be added after all compilation phases, to pass on the context to Blueprint to allow splitting compile and execution pipeline.

nitingupta910:
Some context values like the current logged in user-id are not available
during compilation phases of the document pipeline. So, add another
phase after all compilation phases to grab context values and put them in
blueprint execution context. This is needed when documents are
pre-compiled and are executed later when the context (like current
logged-in userID) values are actually available.

fixes #539

### DIFF
--- a/lib/absinthe/phase/blueprint.ex
+++ b/lib/absinthe/phase/blueprint.ex
@@ -6,15 +6,9 @@ defmodule Absinthe.Phase.Blueprint do
   alias Absinthe.Blueprint
 
   @spec run(any, Keyword.t()) :: {:ok, Blueprint.t()}
-  def run(blueprint, options \\ []) do
+  def run(blueprint, _options \\ []) do
     input = blueprint.input
     blueprint = Blueprint.Draft.convert(input, blueprint)
-
-    context = Map.merge(blueprint.execution.context, options[:context] || %{})
-    blueprint = put_in(blueprint.execution.context, context)
-
-    root_value = Map.merge(blueprint.execution.root_value, options[:root_value] || %{})
-    blueprint = put_in(blueprint.execution.root_value, root_value)
 
     {:ok, blueprint}
   end
--- a/lib/absinthe/phase/document/context.ex
+++ b/lib/absinthe/phase/document/context.ex
@@ -1,0 +1,17 @@
+defmodule Absinthe.Phase.Document.Context do
+  @moduledoc "Pass on context and root value to document."
+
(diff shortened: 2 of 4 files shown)


nitingupta910:
All tests are passing locally. Not sure why they are failing in travis.

Local output from `mix test`:
```
Compiling 93 files (.ex)
Excluding tags: [pending: true]

..............................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................

Finished in 4.0 seconds
718 tests, 0 failures

Randomized with seed 628370
```

benwilson512:
hey @nitingupta910 thanks for taking initiative on this! I definitely like the idea of having the context handling (and we should do root value too) in its own phase. Right now this stuff exists in the `Absinthe.Phase.Blueprint` module, which is a problem for compiled documents since it runs at compile time.

I would like to avoid duplicating the code inside `Absinthe.Phase.Blueprint` so can you modify your phase to handle both the context and the root value, and then remove those bits from `Absinthe.Phase.Blueprint` ?

nitingupta910:
Sure, would send updated pull request shortly.

nitingupta910:
The commit now has all the changes you suggested.

benwilson512:
Thank you!

PR #543, opened 2018-04-27
Fix bug in proxy shard numbering

bernardd:
There's a nasty little bug with subscriptions in distributed (ie multi-node) systems. `Absinthe.Subscription.ProxySupervisor` uses a sharding system which numbers the shard proxies 1..N. However when `Absinthe.Subscription.publish_remote/4` is selecting a shard, it uses `:erlang.phash2/2` which produces a value from 0..N-1. The result is that when a publication occurs on a node other than the one handling the subscription, one out of every N times it will fail to be sent.

I couldn't see any tests directly relating to this, but I can tell you that this change fixed up the issues we were having.


benwilson512:
Great catch, thank yoU!

PR #549, opened 2018-05-10
Showing off the bug with integer value in identifier

0nkery:
Hey, guys. Thanks for awesome work on Absinthe!

Recently I stumbled upon a stealthy bug when field with identifier like `field_1` (integer after underscore) is not allowed to be queried. This thing also prevents subscriptions from publishing any data if such a field is requested.

Note the second field `field2` which resolves fine or not throwing errors at least.

I tried to narrow down and fix the issue by myself but I don't have much time to really dig into Absinthe 
codebase. :(

### DIFF
--- a/test/absinthe/execution/default_resolver_test.exs
+++ b/test/absinthe/execution/default_resolver_test.exs
@@ -26,12 +26,13 @@ defmodule Absinthe.Execution.DefaultResolverTest do
 
       query do
         field :field_1, :integer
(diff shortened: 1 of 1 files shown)


benwilson512:
Hey @0nkery, field names need to be a valid result from `Macro.underscore`, which always results in `_` before a number. I believe we just merged in a warning into master to let folks know when their schema is breaking this rule.

PR #552, opened 2018-05-18
Subscription: Add child_spec/1 hook

xtian:
Allows `Absinthe.Subscription` to be started using the new child spec form from Elixir 1.5: https://hexdocs.pm/elixir/1.6.5/Supervisor.html#module-child_spec-1


benwilson512:
Thank you!

Issue #556, opened 2018-05-30
Dataloader helper, default value for use_parent is wrong


tlvenn:
The documentation states that the default value for `:use_parent` is `true` which should indeed be the default but it appears that the code actually default to `false` at the moment:

https://github.com/absinthe-graphql/absinthe/blob/master/lib/absinthe/resolution/helpers.ex#L239

PR #557, opened 2018-05-30
Fix default value for use_parent

tlvenn:
Closes: #556

### DIFF
--- a/lib/absinthe/resolution/helpers.ex
+++ b/lib/absinthe/resolution/helpers.ex
@@ -236,7 +236,7 @@ defmodule Absinthe.Resolution.Helpers do
     end
 
     defp use_parent(loader, source, resource, parent, args, opts) do
-      with true <- Keyword.get(opts, :use_parent, false),
+      with true <- Keyword.get(opts, :use_parent, true),
            {:ok, val} <- is_map(parent) && Map.fetch(parent, resource) do
         Dataloader.put(loader, source, {resource, args}, parent, val)
       else


tlvenn:
Interesting, some tests are failing and with reason, we don't / can't check that the value we get from the parent is a loaded association.

In this particular case, we got `nil` and we put it in the cache as the value.

It seems we need a `assoc_loaded?` fn at the dataloader source level and the default value of whether to use the parent should probably come from the source as well.

@benwilson512 thought ?

benwilson512:
@tlvenn `nil` is a perfectly valid value to put in the cache though. We handle this with the Ecto source by pattern matching on the Ecto association not loaded struct and just `no-op`ing if you try to `put` that value. https://github.com/absinthe-graphql/dataloader/blob/master/lib/dataloader/ecto.ex#L229

tlvenn:
Good catch regarding Ecto and put but then I dont understand why suddenly the tests are failing. Also do we agree that there is indeed an issue with the default value not matching the documented one ?

benwilson512:
I definitely agree that the docs / implementation have a mismatch.

benwilson512:
The `nil` thing may still be an issue for perhaps the KV store now that I think about it. Let me dig in.

archseer:
Note: it fails with a different error now:

```
  1) test can resolve a field using the normal dataloader helper (Absinthe.Middleware.DataloaderTest)
     test/absinthe/middleware/dataloader_test.exs:100
     ** (Dataloader.GetError) "Unable to find batch {:organization_id, %{pid: #PID<0.1931.0>}}"
     code: assert {:ok, %{data: data}} = Absinthe.run(doc, Schema)
     stacktrace:
       lib/dataloader.ex:193: Dataloader.do_get/2
       (absinthe) lib/absinthe/resolution/helpers.ex:294: anonymous fn/6 in Absinthe.Resolution.Helpers.do_dataloader/6
       (absinthe) lib/absinthe/middleware/dataloader.ex:33: Absinthe.Middleware.Dataloader.get_result/2
       (absinthe) lib/absinthe/phase/document/execution/resolution.ex:230: Absinthe.Phase.Document.Execution.Resolution.reduce_resolution/1
       (absinthe) lib/absinthe/phase/document/execution/resolution.ex:185: Absinthe.Phase.Document.Execution.Resolution.do_resolve_field/3
       (absinthe) lib/absinthe/phase/document/execution/resolution.ex:170: Absinthe.Phase.Document.Execution.Resolution.do_resolve_fields/6
       (absinthe) lib/absinthe/phase/document/execution/resolution.ex:88: Absinthe.Phase.Document.Execution.Resolution.walk_result/5
       (absinthe) lib/absinthe/phase/document/execution/resolution.ex:114: Absinthe.Phase.Document.Execution.Resolution.walk_results/6
       (absinthe) lib/absinthe/phase/document/execution/resolution.ex:103: Absinthe.Phase.Document.Execution.Resolution.walk_result/5
       (absinthe) lib/absinthe/phase/document/execution/resolution.ex:280: Absinthe.Phase.Document.Execution.Resolution.build_result/3
       (absinthe) lib/absinthe/phase/document/execution/resolution.ex:170: Absinthe.Phase.Document.Execution.Resolution.do_resolve_fields/6
       (absinthe) lib/absinthe/phase/document/execution/resolution.ex:88: Absinthe.Phase.Document.Execution.Resolution.walk_result/5
       (absinthe) lib/absinthe/phase/document/execution/resolution.ex:67: Absinthe.Phase.Document.Execution.Resolution.perform_resolution/3
       (absinthe) lib/absinthe/phase/document/execution/resolution.ex:24: Absinthe.Phase.Document.Execution.Resolution.resolve_current/3
       (absinthe) lib/absinthe/pipeline.ex:351: Absinthe.Pipeline.run_phase/3
       (absinthe) lib/absinthe.ex:103: Absinthe.run/3
       test/absinthe/middleware/dataloader_test.exs:119: (test)



  2) test can resolve batched fields cross-query that have different data requirements (Absinthe.Middleware.DataloaderTest)
     test/absinthe/middleware/dataloader_test.exs:126
     ** (Dataloader.GetError) "Unable to find batch {:organization_id, %{pid: #PID<0.1932.0>}}"
     code: assert {:ok, %{data: data}} = Absinthe.run(doc, Schema)
     stacktrace:
       lib/dataloader.ex:193: Dataloader.do_get/2
       (absinthe) lib/absinthe/resolution/helpers.ex:294: anonymous fn/6 in Absinthe.Resolution.Helpers.do_dataloader/6
       (absinthe) lib/absinthe/middleware/dataloader.ex:33: Absinthe.Middleware.Dataloader.get_result/2
       (absinthe) lib/absinthe/phase/document/execution/resolution.ex:230: Absinthe.Phase.Document.Execution.Resolution.reduce_resolution/1
       (absinthe) lib/absinthe/phase/document/execution/resolution.ex:185: Absinthe.Phase.Document.Execution.Resolution.do_resolve_field/3
       (absinthe) lib/absinthe/phase/document/execution/resolution.ex:170: Absinthe.Phase.Document.Execution.Resolution.do_resolve_fields/6
       (absinthe) lib/absinthe/phase/document/execution/resolution.ex:88: Absinthe.Phase.Document.Execution.Resolution.walk_result/5
       (absinthe) lib/absinthe/phase/document/execution/resolution.ex:114: Absinthe.Phase.Document.Execution.Resolution.walk_results/6
       (absinthe) lib/absinthe/phase/document/execution/resolution.ex:103: Absinthe.Phase.Document.Execution.Resolution.walk_result/5
       (absinthe) lib/absinthe/phase/document/execution/resolution.ex:280: Absinthe.Phase.Document.Execution.Resolution.build_result/3
       (absinthe) lib/absinthe/phase/document/execution/resolution.ex:170: Absinthe.Phase.Document.Execution.Resolution.do_resolve_fields/6
       (absinthe) lib/absinthe/phase/document/execution/resolution.ex:88: Absinthe.Phase.Document.Execution.Resolution.walk_result/5
       (absinthe) lib/absinthe/phase/document/execution/resolution.ex:67: Absinthe.Phase.Document.Execution.Resolution.perform_resolution/3
       (absinthe) lib/absinthe/phase/document/execution/resolution.ex:24: Absinthe.Phase.Document.Execution.Resolution.resolve_current/3
       (absinthe) lib/absinthe/pipeline.ex:351: Absinthe.Pipeline.run_phase/3
       (absinthe) lib/absinthe.ex:103: Absinthe.run/3
       test/absinthe/middleware/dataloader_test.exs:149: (test)
```

I'm currently looking into this because I'd really like it fixed. (I also tried overriding this on the `dataloader/2` call, but that leads to a set of different issues #720)

archseer:
@benwilson512 So after some digging, the issue seems to be the on_load callback:

Here the source will still contain batch data:
https://github.com/absinthe-graphql/absinthe/blob/9c8e97167efacd7da2d28ef59e8a7c3341330a8f/lib/absinthe/resolution/helpers.ex#L289

But not here:

https://github.com/absinthe-graphql/absinthe/blob/9c8e97167efacd7da2d28ef59e8a7c3341330a8f/lib/absinthe/resolution/helpers.ex#L294

That's because in this function, the first branch won't use the passed in context:

https://github.com/absinthe-graphql/absinthe/blob/9c8e97167efacd7da2d28ef59e8a7c3341330a8f/lib/absinthe/middleware/dataloader.ex#L15-L26

Changing it to this:

```elixir
    def call(%{state: :unresolved} = resolution, {loader, callback}) do
      if !Dataloader.pending_batches?(loader) do
        resolution = put_in(resolution.context.loader, loader)
        get_result(resolution, callback)
      else
        %{
          resolution
          | context: Map.put(resolution.context, :loader, loader),
            state: :suspended,
            middleware: [{__MODULE__, callback} | resolution.middleware]
        }
      end
    end
```

Made the code use the cache data, but now the issue is that it gets back `"1"` -- which is what was put into the cache by `use_parent` (it seems to incorrectly use `organization_id` instead of `organization` for the cache value).

bgentry:
I believe this can be closed now that #882 is merged, and #556 is also closed.

benwilson512:
That is correct. Dataloader's Ecto source will also properly ignore `%Ecto.AssociationNotLoaded{}` structs.

PR #558, opened 2018-05-31
Add dataloader callback

tlvenn:
This is particularly useful with computed fields which need an association to resolve.

### DIFF
--- a/lib/absinthe/resolution/helpers.ex
+++ b/lib/absinthe/resolution/helpers.ex
@@ -254,9 +254,14 @@ defmodule Absinthe.Resolution.Helpers do
       |> use_parent(source, resource, parent, args, opts)
       |> Dataloader.load(source, {resource, args}, parent)
       |> on_load(fn loader ->
-        result = Dataloader.get(loader, source, {resource, args}, parent)
-        {:ok, result}
+        callback = Keyword.get(opts, :callback, &default_callback/3)
+
+        loader
+        |> Dataloader.get(source, {resource, args}, parent)
+        |> callback.(parent, args)
       end)
     end
+
+    defp default_callback(result, _parent, _args), do: {:ok, result}
   end
 end


benwilson512:
Can you provide a test case or an example of use?

tlvenn:
Sure thing, I have a classical parent <--> childs assoc, a root query which returns a list of childs and there is a field on that child which is a computed value which needs the parent to in order to properly compute the value.

So for example:
```
field :effective_price, :integer, resolve: dataloader(Ecto, :activity, args: %{with_active_promotions: true}, callback: fn activity, bookable, _ ->
  {:ok, Bookable.effective_price(bookable, activity, activity.promotions)}
end)
```

This kinda falls to some degree into the same problem category as what @jfrolich is trying to solve.

Ideally, I should not need to do this here given this is not a concern of the graphql layer or what the client is even requesting but as far as I know, I cant really use the dataloader in my business logic in a way that it would be coordinated with the resolution of fields, batch things properly and would be able to inject back the loader once it's done with it.

So for now, as a lack of a better solution I end up pushing that loading concern up into my absinthe / graphql layer so that everything I need in my business layer is already efficiently loaded. Of course, I apply this pattern only when this kind of N+1 issue emerge in my business logic.

Does that make sense ? Is there a better way to solve this ?

tlvenn:
@benwilson512 any chance to merge this small PR ?

jfrolich:
Hi @tlvenn, you can do this in your code base by your own implementation of the dataloader helper function (but you are probably aware of this). Agree that there need to be a proper solution for this. And also agree this callback already solves some of these use-cases. I need this in almost any project (sometimes also solved using the batch helper).

tlvenn:
Hi @jfrolich, ya as you pointed out, this can be handled by a customised version of the dataloader helper but with this simple callback, I avoid duplicating and managing this helper which is great ;)

@benwilson512 any though on this subject ? Can we merge this simple PR plz ?

tlvenn:
@benwilson512 any update plz ? Would very much like to get rid off my fork...

benwilson512:
Sorry, been a bit myopic on the 1.5 branch!

PR #562, opened 2018-06-19
Support `import_fields` for `interface`s

binaryseed:
Howdy! I noticed that the docs say that `interface` supports `import_fields`, but it seems to be missing. I put this together to implement support for this mirroring how it's done with the other types..

Let me know if we need to do anything more than this!


binaryseed:
Awesome! It'd be great to get a little version bump for this :)

PR #564, opened 2018-06-20
Add correct syntax for pretty print

maartenvanvliet:
Without adding the `true` parameter the command to pretty print does not work for me. Adding it fixed it for me.

### DIFF
--- a/lib/mix/tasks/absinthe.schema.json.ex
+++ b/lib/mix/tasks/absinthe.schema.json.ex
@@ -36,7 +36,7 @@ defmodule Mix.Tasks.Absinthe.Schema.Json do
   Write to path `/path/to/schema.json` using the `MySchema` schema, using the
   default `#{@default_codec_name}` JSON codec, and pretty-printing:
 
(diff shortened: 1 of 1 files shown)


benwilson512:
`--foo` should evaluate to `foo: true` per the option parser docs. I just tried this locally and it works fine for me, what version of Elixir are on you?

maartenvanvliet:
I'm am running on elixir 1.6.5

Sorry, I was not complete in the description of what's happening. If I omit the `true` from the cli command it does not write the schema.json into the directory, but rather into the current directory.
```
mix absinthe.schema.json --schema Sketchql.Graphql.Schema --pretty priv/schema.json
* creating .
* creating ./schema.json
```
So, the file is written into `./schema.json` instead of `priv/schema.json`

When, the `true` is included it does write it to `priv/schema.json`
```
$ mix absinthe.schema.json --schema Todo.Graphql.Schema --pretty true priv/schema.json
* creating priv
* creating priv/schema.json
```

PR #566, opened 2018-06-27
Fix incorrectly placed ) in docs

maartenvanvliet:
In the docs the Dataloader.get/4 is called with 3 arguments, and a dangling fourth.


benwilson512:
Thanks!

PR #575, opened 2018-07-11
Update Adapter documentation to not refer to configuration

axelson:
The configuration does not appear to be read so instead just document how to set the adapter configuration.

Note: I think the [Absinthe.Plug](https://hexdocs.pm/absinthe_plug/Absinthe.Plug.html) documentation for Phoenix is incorrect. For me instead of:

```
forward "/graphiql",
  to: Absinthe.Plug.GraphiQL,
  init_opts: [
    schema: MyAppWeb.Schema,
    interface: :simple
  ]
```

What works for me is:

```
  forward "/graphiql", Absinthe.Plug.GraphiQL,
    [schema: MyAppWeb.Schema, adapter: YourApp.Adapter.TheAdapterName]
```

Should we use that style here as well? (I may be misunderstanding something about `forward` with plug vs Phoenix)

### DIFF
--- a/lib/absinthe/adapter.ex
+++ b/lib/absinthe/adapter.ex
@@ -17,14 +17,9 @@ defmodule Absinthe.Adapter do
     modifications. (Note at the current time this does not support introspection
     if you're using camelized conventions).
 
-  To set an adapter, you can set an application configuration value:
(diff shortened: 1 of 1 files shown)


benwilson512:
Thanks!

PR #576, opened 2018-07-11
Add an Absinthe.Adapter.Underscore adapter

axelson:
It is similar to the `Absinthe.Adapter.LanguageConventions` adapter but converts all incoming identifiers to underscores and does not modify outgoing identifiers (since those are already expected to be underscores). Unlike `Absinthe.Adapter.Passthrough` it does not break introspection.

If this has a chance of being merged than I gladly work on some tests for this adapter. If not maybe I could add this adapter to the wiki somewhere?


benwilson512:
Thanks!

PR #578, opened 2018-07-15
Documentation was moved

Razzeee:
Just updated the links to reference the new(er) documentation. Old links lead nowhere.


benwilson512:
Thanks!

PR #582, opened 2018-07-24
Default MapGet middleware should ignore already resolved fields

axelson:
Similar to the default resolve middleware `Absinthe.Middleware.MapGet` should ignore already resolved fields instead of over-writing their values.

### DIFF
--- a/lib/absinthe/middleware/map_get.ex
+++ b/lib/absinthe/middleware/map_get.ex
@@ -1,6 +1,11 @@
 defmodule Absinthe.Middleware.MapGet do
   @moduledoc """
-  This is the default middleware.
+  This is the default middleware. It assumes the the object it receives is a map
+  and uses `Map.get/2` to get the value for this field. If this field is already
+  marked as resolved, then this middleware does not touch it.
+
+  If you want to replace this middleware you should use
+  `Absinthe.Schema.replace_default/4`
   """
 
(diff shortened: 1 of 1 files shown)


axelson:
I'm having trouble creating a test for this because I get the following error:
```
Jasons-MBP-2% mix test
Compiling 1 file (.erl)
/Users/jason/.asdf/installs/erlang/20.2.2/lib/parsetools-2.1.6/include/leexinc.hrl:36: syntax error before:
/Users/jason/.asdf/installs/erlang/20.2.2/lib/parsetools-2.1.6/include/leexinc.hrl:10: function token/2 undefined
/Users/jason/.asdf/installs/erlang/20.2.2/lib/parsetools-2.1.6/include/leexinc.hrl:10: function token/3 undefined
/Users/jason/.asdf/installs/erlang/20.2.2/lib/parsetools-2.1.6/include/leexinc.hrl:10: function tokens/2 undefined
/Users/jason/.asdf/installs/erlang/20.2.2/lib/parsetools-2.1.6/include/leexinc.hrl:10: function tokens/3 undefined
/Users/jason/.asdf/installs/erlang/20.2.2/lib/parsetools-2.1.6/include/leexinc.hrl:21: function string/4 undefined
```

benwilson512:
Thanks!

axelson:
Wow, that was fast!

PR #584, opened 2018-07-27
Add missing definition of Absinthe.Blueprint.Input.RawValue.t()

vderyagin:
`Absinthe.Blueprint.Input.Value.t()`[ refers to](https://github.com/absinthe-graphql/absinthe/blob/5600d03a0c9daa493a869910a6e2be3689082d35/lib/absinthe/blueprint/input/value.ex#L30) `Absinthe.Blueprint.Input.RawValue.t()`, whis is not defined.

This change defines that missing type.


benwilson512:
Thanks!

PR #598, opened 2018-08-25
Document dataloader helper callback option

maartenvanvliet:
Documents the callback option for the dataloader helper. Also included an example to illustrate its use.

### DIFF
--- a/lib/absinthe/resolution/helpers.ex
+++ b/lib/absinthe/resolution/helpers.ex
@@ -165,6 +165,13 @@ defmodule Absinthe.Resolution.Helpers do
       field :organization, :organization do
         resolve dataloader(Accounts, :organization, use_parent: false)
       end
+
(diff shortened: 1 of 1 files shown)


lib/absinthe/resolution/helpers.ex
```diff
@@ -165,6 +165,13 @@ defmodule Absinthe.Resolution.Helpers do
       field :organization, :organization do
         resolve dataloader(Accounts, :organization, use_parent: false)
       end
+
+      field(:subscription_active, non_null(:boolean), resolve: dataloader(
```
benwilson512: Let's pick an example that doesn't involve the word "subscription", I'm concerned it will confuse people who will think this has something to do with GraphQL subscriptions.

maartenvanvliet: Good point

benwilson512:
Thanks!

PR #599, opened 2018-08-25
Update ex_doc

maartenvanvliet:
`mix docs` fails on elixir 1.7 on versions on the previous version
of ex_doc with `** (RuntimeError) module Absinthe.Adapter.LanguageConventions
was not compiled with flag --docs`

See https://github.com/elixir-lang/elixir/issues/8034

We can also expand the test matrix in travis to test on elixir 1.7? Would not have caught this bug but might catch others.


benwilson512:
Thanks!

benwilson512:
I'll update the travis definition too.

PR #602, opened 2018-08-29
v1.5: Schema definition decoration mechanism

bruce:
(This PR is part of the schema definition rework slated for v1.5 and targets the branch in #503.)

This PR adds support for a new schema hook, `decorations/2`. This function is used to modify the schema definition blueprint tree on-the-fly during schema compilation. Most notably, this is useful for applying dynamic features (such as resolvers) to schemas built using the [Schema Definition Language](https://www.apollographql.com/docs/resources/graphql-glossary.html#schema-definition-language) (SDL). People with a JavaScript + GraphQL background may recognize a parallel with [`makeExecutableSchema`](https://www.apollographql.com/docs/graphql-tools/generate-schema.html#makeExecutableSchema), although our approach is quite different.

The function is provided the blueprint schema definition node as the first argument and its ancestor nodes in a list as its second argument. Matching against these two arguments gives schema authors a straightforward way to apply decorations to the specific piece(s) of the blueprint they'd like.

## Examples

Assuming a simple schema built using `import_sdl`:

```elixir
defmodule MyApp.Schema do
  use Absinthe.Schema

  import_sdl """
  type Query {
    users: [User]!
  }

  type User {
    name: String!
  }
  """

end
```

If we wanted to wire in a resolver, `MyApp.Resolvers.Accounts.get_users/3` for the `:users` field, we could add a `decorations/2` function somewhere in the `MyApp.Schema` module:

```elixir
def decorations(%{identifier: :users}, [%{identifier: :query} | _rest]) do
  {:resolve, &MyApp.Resolvers.Accounts.get_users/3}
end
```

(Strictly speaking, we don't need to match against the `:query` object, as there's only one `:users` definition around, but we do so here to show you have access to the field's ancestor definitions. We _could_ have matched against the specific definition structs here,  namely `%Absinthe.Blueprint.Schema.FieldDefinition{}` and `%Absinthe.Blueprint.Schema.ObjectTypeDefinition{}`, but the above is a bit more succinct and readable.)

## Extensibility

The specific instructions (_decorations_) that the `decorations/2` function can return and can be applied to definitions is completely dependent on the decorator module that the `Absinthe.Phase.Schema.Decorate` is configured to use as part of the compilation pipeline.

We'll ship with support for a number of decorations, but (as with most things Absinthe), the mechanism will be customizable. Adding support for your own bespoke decorations will just be a matter of modifying the pipeline and writing a module that implements a behaviour.

## Limitations

You will _not_ be able to supply decorations that include anonymous functions (that is, using `fn`) due to AST-related limitations during compilation. We'll support:

- `&Simple.captures/1` like `&this/2`
- `{Module, :function}` and `{Module, :function, [args]}`

There will be more details in the docs.

## Verbosity

The reaction from developers with JS GraphQL experience may be that the `decorations/2` hook is much more verbose than `makeExecutableSchema` (_3 lines for each field resolver!_).

Here are some features that will help:

:one: We'll support a decoration for the top-level schema definition that will let you apply changes deeply, eg:

```elixir
def decorations(%Absinthe.Blueprint.SchemaDefinition{}, _) do
  {:resolve, [
    query: [
      users: &MyApp.Resolvers.Accounts.get_users/3
    ]
  ]}
end
```

:two: The function supports returning multiple decorations (as the name may suggest):

```elixir
def decorations(%{identifier: :users}, [%{identifier: :query} | _rest]) do
  [
    {:description, "A dynamically supplied description"},
    {:meta, [some: "special", metadata: :here]},
    {:resolve, &MyApp.Resolvers.Accounts.get_users/3}
  ]
end
```

This gives you a fair amount of condensed functionality.

:three:  Because you can match anything inside the definitions (including metadata, etc), you can _very_ flexibly address modifications to the schema.

:four: ... and, lastly, you can use your own crazy decorating scheme if you'd like to! Just write a decorator module and configure `Absinthe.Phase.Schema.Decorate` to use it instead!

## Special Note

There's nothing specifically SDL-related about `decorations/2`. While it's important that we provide a mechanism like this so that SDL schemas work, there's nothing stopping you from using it in schemas that are build using the conventional `Absinthe.Schema.Notation` schema definition macros (or a mix of those + `import_sdl`).


benwilson512:
Superseded by #606

PR #604, opened 2018-09-05
fix documentation typo

jeffweiss:
Prior to this commit, the explanatory text about the async middleware
mispelled `reason` as `reasoon`.


benwilson512:
Thanks!

PR #606, opened 2018-09-18
Sdl resolvers

benwilson512:
- [x] remove the duplicated `__ensure_middleware__` call
- [x] extract type building from compilation phase


kamilchm:
Great direction! I'm looking for something than can be easy used to replace prototypes written with Apollo and this can save a lot of tedious work of translating GraphQL SDL into Absinthe DSL. I'm now looking forward for 1.5 release :+1:

arathunku:
Same here! Cannot wait for the alpha release :) I've got a small admin panel that's based on Absinthe, Preact and `Absinthe.Phoenix.Controller` (@graphql, where it makes sense for server rendered view).

PR #607, opened 2018-09-19
require explicitly specifying node name in subscription pubsub

azhi:
modify pubsub behaviour to include `node_name/0` callback

### Justification for breaking change

Currently pubsub system (specifically pubsub proxy server) relies on `Kernel.node/0` to check whether message is local or not. `Kernel.node` is a function that was intended to work inside erlang cluster.

But pubsub system can be run in a "cluster" connected through a bunch of mechanisms, not only in erlang cluster. Phoenix pubsub recognizes this and includes a [node_name](https://hexdocs.pm/phoenix_pubsub/Phoenix.PubSub.html#node_name/1) function in it's specification, which is used in different pubsub adapters, such as [redis one](https://hexdocs.pm/phoenix_pubsub_redis/Phoenix.PubSub.Redis.html#module-options), where you could explicitly specify node name.

In fact, this is the background for this PR - I was using redis adapter, and couldn't make subscriptions work. Turns out it was because of this node names issue. It could be fixed by switching from using explicit `node_name` in `Phoenix.PubSub.Redis` config to using `--sname` flag, but I think it is worth fixing now when not a lot of people is dependent on current `Absinthe.Subscription.Pubsub`.


benwilson512:
Hey @azhi I appreciate the PR! We're gonna be reconfiguring the whole internals of PubSub for 1.5 probably, so I my hold off on merging this for a while so I can evaluate it in that context.

azhi:
Oh, ok, I guess I'll have to revisit all this stuff when upgrading to 1.5 :).

I was also gonna make some work in supporting [Direct usage](https://hexdocs.pm/phoenix_pubsub/Phoenix.PubSub.html#module-direct-usage) of phoenix pubsub in absinthe_phoenix. Do you want me to open PR for that, or this part will also be reworked in 1.5?

benwilson512:
Thanks!

benwilson512:
@azhi by all means do a PR for direct usage. I'm hoping we wouldn't have to change very much but please go for it!

PR #608, opened 2018-09-19
Update Subscriptions Tutorial with one extra line

ivosequeros:
After following the subscriptions tutorial, I realised that there is a required line missing.

### DIFF
--- a/guides/tutorial/subscriptions.md
+++ b/guides/tutorial/subscriptions.md
@@ -52,6 +52,7 @@ defmodule BlogWeb.Endpoint do
   use Phoenix.Endpoint, otp_app: :blog
   use Absinthe.Phoenix.Endpoint
 
+  socket "/websocket", BlogWeb.UserSocket
   << rest of the file>>
 ```
 


bruce:
@ivosequeros Thanks! You're a keen reader!

Should the path be `/socket`, though?

ivosequeros:
Hi Bruce!
I tried `/socket` but it was failing. By opening the Chrome Dev Tools, I saw that both Graphiql and Playground connect to `/websocket`, so I changed it and it worked.

screenshot 2018-09-20 at 00 56 20

benwilson512:
@ivosequeros can you show how you have your socket setup? By default the path is `"/socket"` but if you've changed it in your Endpoint.ex that may be why.

ivosequeros:
Sure. This is my Socket:
```
defmodule Server.UserSocket do
  use Phoenix.Socket
  use Absinthe.Phoenix.Socket, schema: Server.Schema

  transport :websocket, Phoenix.Transports.WebSocket

  def connect(%{"Authorization" => auth}, socket) do
    "Bearer " <> token = auth
    {:ok, resource} = Server.Authentication.Access.authenticate token
    socket = Absinthe.Phoenix.Socket.put_options socket, context: resource

    {:ok, socket}
  end

  def connect(_, socket) do
    {:error, :no_access}
  end

  def id(_socket), do: nil
end
```

And this is my Endpoint
```
defmodule Server.Endpoint do
  use Phoenix.Endpoint, otp_app: :lighthouse
  use Absinthe.Phoenix.Endpoint

  # The default was /socket, but it didn't work with Playground nor Graphiql.
  # It works with /websocket since Playground seems to connect any URL you give it + /websocket
  socket "/websocket", Server.UserSocket

  ...
end
```

benwilson512:
Right, so the default is `socket "/socket", Server.UserSocket`. The issue you're having is that the GraphiQL interface will cache whatever preferences you have. If you change the `Endpoint` back to the default, clear your cache in GraphiQL, it'll go back to using that default.

ivosequeros:
Hi @benwilson512, you're right. I cleaned the LocalStorage and it works with `/socket` 👍

PR #611, opened 2018-09-25
Add closing parenthesis for example function

maartenvanvliet:
Example code was missing closing parenthesis in anonymous function


benwilson512:
Thanks!

PR #620, opened 2018-10-19
Modify lexer and parser to support parsing kitchen-sink and schema-kitchen-sink fixtures

potatosalad:
This pull request adds tests for parsing the fixtures [`kitchen-sink.graphql`](https://github.com/graphql/graphql-js/blob/v14.0.2/src/language/__tests__/kitchen-sink.graphql) and [`schema-kitchen-sink.graphql`](https://github.com/graphql/graphql-js/blob/v14.0.2/src/language/__tests__/schema-kitchen-sink.graphql) as used in [graphql/graphql-js](https://github.com/graphql/graphql-js).

**Update 2018-12-04:** The tests now pass after rebasing against master.

## Overview of Changes

- `Absinthe.Lexer`
  - Ignore Ampersand (`&`) in the same way that Comma (`,`) is ignored.

     _Reasoning:_
     
     The [June 2018 GraphQL spec](https://facebook.github.io/graphql/June2018/#sec-Objects) specifies that multiple interfaces on an object type are separated by a `&`: `type O implements I1 & I2`
     
     This change came after many implementations out there have already started using a `,` to specify multiple interfaces: `type O implements I1, I2`
     
     By ignoring `&` and `,`, the lexer can handle both non-standard and standard forms.
     
  - Remove `BooleanValue` and instead treat booleans as if they were reserved words by traversing on `boolean_value_or_name_or_reserved_word`.
  
     _Reasoning:_
     
     Try parsing the following document using the old and new lexer:
     
     ```graphql
     query AnnoyingQuery($truetype: Font, $falsetto: Boolean = false, $nullify: Boolean = null) {
       annoying(truetype: $truetype, falsetto: $falsetto, nullify: $nullify)
     }
     ```

     You will find that the old lexer parses it correctly while the new one attempts to treat things like `$falsetto` as `false`.
     
     There may be a better way to handle this, but I noticed that `null` was being handled roughly the same way and didn't show the same symptoms.

- `:absinthe_parser`
  - Support "undefined style" definitions.

     _Example:_

     ```graphql
     enum UndefinedEnum
     enum UndefinedEnum @onType
     input UndefinedInputObject
     input UndefinedInputObject @onType
     interface UndefinedInterface
     interface UndefinedInterface @onType
     schema
     schema @onType
     type UndefinedObject
     type UndefinedObject @onType
     union UndefinedUnion
     union UndefinedUnion @onType
     ```

     _Reasoning:_

     All of that in order to support `extend` as seen below.

  - Support `extend` definitions.

     _Example:_

     ```graphql
     extend enum UndefinedEnum = VALUE
     extend input UndefinedInputObject { key: Int }
     extend interface UndefinedInterface { key: Int }
     extend schema { subscription: RootSubscriptionType }
     extend type UndefinedObject { key: Int }
     extend union UndefinedUnion = AnotherType
     ```
     
     _Reasoning:_
     
     Because almost everything is now extensible according to [the spec](https://facebook.github.io/graphql/June2018/#sec-Type-System-Extensions).  Directives aren't extensible, for now.

  - Support Pipe (`|`) prefixed members of directives and unions.

     _Example:_

     ```graphql
     directive @foo on FIELD | FRAGMENT_SPREAD
     directive @bar on
       | FIELD
       | FRAGMENT_SPREAD

     union Foo = A | B
     union Bar =
       | A
       | B
     ```

  - Added [`Expect 10`](https://github.com/potatosalad/absinthe/blob/3c6a844f27e6982f6e105e3cb904fe6126cdf9fb/src/absinthe_parser.yrl#L315) to the parser to suppress 10 shift/reduce conflicts.

     _Reasoning:_

     As a result of going from:

     ```erlang
     EnumTypeDefinition -> 'enum' Name '{' EnumValueDefinitionList '}' : '...'.
     EnumTypeDefinition -> 'enum' Name Directives '{' EnumValueDefinitionList '}' : '...'.
     ```

     To the more ambiguous:

     ```erlang
     EnumTypeDefinition -> 'enum' Name : '...'.
     EnumTypeDefinition -> 'enum' Name Directives : '...'.
     EnumTypeDefinition -> 'enum' Name '{' EnumValueDefinitionList '}' : '...'.
     EnumTypeDefinition -> 'enum' Name Directives '{' EnumValueDefinitionList '}' : '...'.
     ```

     On compilation, `yecc` will warn about the shift/reduce ambiguity (10 in total).  There may be some way to remove the warnings, but from [what I read](http://erlang.org/doc/man/yecc.html), you can simply tell `yecc` to ignore the known ambiguity and only warn if the number changes.

     From the docs:

     > The optional expect declaration can be placed anywhere before the last optional section with Erlang code. It is used for suppressing the warning about conflicts that is ordinarily given if the grammar is ambiguous. An example:
     >
     > ```erlang
     > Expect 2.
     > ```
     >
     > The warning is given if the number of shift/reduce conflicts differs from 2, or if there are reduce/reduce conflicts.

### DIFF
--- a/lib/absinthe/lexer.ex
+++ b/lib/absinthe/lexer.ex
@@ -52,19 +52,24 @@ defmodule Absinthe.Lexer do
   # Comma :: ,
   comma = ascii_char([?,])
 
+  # Ampersand :: &
+  ampersand = ascii_char([?&])
+
   # Ignored ::
   #   - UnicodeBOM
   #   - WhiteSpace
   #   - LineTerminator
   #   - Comment
   #   - Comma
+  #   - Ampersand
   ignored =
     choice([
       unicode_bom,
       whitespace,
       line_terminator,
       comment,
-      comma
+      comma,
+      ampersand
     ])
 
   # ## Lexical Tokens
@@ -95,10 +100,10 @@ defmodule Absinthe.Lexer do
     ])
     |> traverse({:atom_token, []})
 
-  name_or_reserved_word =
+  boolean_value_or_name_or_reserved_word =
     ascii_char([?_, ?A..?Z, ?a..?z])
     |> repeat(ascii_char([?_, ?0..?9, ?A..?Z, ?a..?z]))
-    |> traverse({:name_or_reserved_word_token, []})
+    |> traverse({:boolean_value_or_name_or_reserved_word, []})
 
   # NegativeSign :: -
   negative_sign = ascii_char([?-])
@@ -211,14 +216,6 @@ defmodule Absinthe.Lexer do
     |> ignore(string(~S(""")))
     |> traverse({:block_string_value_token, []})
 
-  # BooleanValue : one of `true` `false`
-  boolean_value =
-    choice([
-      string("true"),
-      string("false")
-    ])
-    |> traverse({:boolean_value_token, []})
-
   defp not_end_of_quote(<<?", _::binary>>, context, _, _) do
     {:halt, context}
   end
@@ -260,8 +257,7 @@ defmodule Absinthe.Lexer do
         string_value,
         float_value,
         int_value,
-        boolean_value,
-        name_or_reserved_word
+        boolean_value_or_name_or_reserved_word
       ])
     )
   )
@@ -274,6 +270,11 @@ defmodule Absinthe.Lexer do
     binary = :unicode.characters_to_binary([value])
     {[binary], context}
   end
+
+  @boolean_words ~w(
+    true
+    false
+  ) |> Enum.map(&String.to_charlist/1)
 
   @reserved_words ~w(
     directive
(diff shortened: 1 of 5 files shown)


potatosalad:
Another parsing problem I discovered while trying to parse [`github-schema.graphql`](https://github.com/graphql/graphql-js/blob/master/src/__fixtures__/github-schema.graphql) is that larger unicode codepoints (like emoji) are not handled correctly in either escaped or unescaped form.

### Examples

```graphql
"👍"
schema
```

```elixir
```

bruce:
Great work on this; I'm going to look at pulling this in after a few small SDL-related fixes around schema validation.

bruce:
Sorry about the delay on this. Excellent work, thank you!

PR #623, opened 2018-10-25
Clarifying what context is being talked about

S1r-Lanzelot:
I am new to Elixir, Phoenix, and obviously Absinthe - when reading this line `In a Phoenix application you'll generally have one
source per context...` I had troubles discerning if the docs were referring to phoenix contexts or absinthe contexts thus I am adding "absinthe" before "context" here - please let me know if this is actually in reference of Phoenix contexts...

### DIFF
--- a/guides/ecto.md
+++ b/guides/ecto.md
@@ -159,7 +159,7 @@ Latest install instructions found here: https://github.com/absinthe-graphql/data
 
 Let's start with a data source. Dataloader data sources are just structs that encode
 a way of retrieving data in batches. In a Phoenix application you'll generally have one
-source per context, so that each context can control how its data is loaded.
+source per absinthe context, so that each context can control how its data is loaded.
 
 Here is a hypothetical `Blog` context and a dataloader ecto source:
 


S1r-Lanzelot:
@benwilson512 Oh alright got it, thanks for the response. How about adding dataloader beo

benwilson512:
Feel free to edit the wording BTW, I'm happy to accept a PR to clarify!

S1r-Lanzelot:
@benwilson512 thanks for the response… I read it a couple times and here is my current confusion, it seems that the Phoenix documentation has some (perhaps) conflicting advice: 

> Sometimes it may be tricky to determine if two resources belong to the same context or not. In those cases, prefer distinct contexts per resource and refactor later if necessary. Otherwise you can easily end-up with large contexts of loosely related entities. In other words: if you are unsure, you should prefer explicit modules (contexts) between resources.
https://hexdocs.pm/phoenix/contexts.html#thinking-about-design

Reading this I think "ok, maybe I should be breaking up my context into smaller ones (per resource possibly - so multiple contexts per data source)." But then I read this in Absinthe docs:

> In a Phoenix application you’ll generally have one source per context, so that each context can control how its data is loaded.

Am I misunderstanding the meaning of source?

I am not trying to turn this into Q&A, just trying to nail down the correct wording so that I/others can understand - I might be mixing up a few things here. Perhaps I should move this to an "issue"... Please let me know if so. Thanks!

S1r-Lanzelot:
@benwilson512 its all good, I had to reread it a few times but initially I thought that the docs were saying generally you will have a 1:1 between source and context but after rereading I see its 1:m. I am not sure a change is needed.

hisapy:
Here's my attempt to clarify it for you and possibly others ... 

You can think of a context in Phoenix as a _module_ where you put functions to handle a subset of the data of the application, i.e.: the Users context, where you have functions functions like: `create_users`, `updated_users`, `delete_user` but also functions to read data like `get_user`, `find_users` and so on. This pattern is based on Domain Driven Design.

So based on the above, what Absinthe docs is trying to tell you, is that _generally_ if you follow the above pattern, you might be agree functions to load your data, in this case with `Dataloader` should also be in their corresponding context.

On the other hand, `Absinthe` supports `Dataloader` by requiring you to add your data sources. So if we have 2 _Phoenix contexts_ like users and news each with its own Dataloader and you want to get data from both via GraphQL, then you'll have to add on source per context.

Notice however, that you can have one big data source, like one _Phoenix context_ centralizing all Dataloader related logic. Absinthe does not interfere in the design of your _Phoenix contexts_

S1r-Lanzelot:
@hisapy thank you - that clears a lot up for me!

benwilson512:
Right. A Phoenix context will usually have several (tens even) of schemas within it. It represents a unit of interrelated business logic and data. You generally expose that data via dataloader with a single source that is provided by the context module.

PR #624, opened 2018-10-31
Correct typo in error message

rmoorman:
This change corrects a small typo within the error that is being raised when a subscription's config function does return an unexpected value.


benwilson512:
Thanks!

rmoorman:
You are very welcome! 😃

PR #629, opened 2018-11-06
Subscriptions - multiple topic support

darrenclark:
Adds support for specifying multiple topics when configuring a subscription field.

The function provided to `config` can now return either a single topic, or a list of topics, for example:

```elixir
subscription do
  field :some_field, :string do
    config fn _, _ ->
      {:ok, topic: ["topic_1", "topic_2", "topic_3"]}
    end
  end
end
```

Thanks to @benwilson512 for pointing me in the right direction!


benwilson512:
This rocks, thank you!

PR #635, opened 2018-11-15
Build type __reference__ values for SDL-generated types

bruce:
This PR calculates `__reference__` values for types created by `import_sdl/2`, with specialized support for the `:path` option (so schema validation errors will reference the source `.graphql` file & line). Various parts of Absinthe assume `__reference__` is set.

Aside: @benwilson512 and I have discussed reassessing how we use `__reference__` and `__private__` (and `source_location`, likely), considering some conflation in use.

### DIFF
--- a/lib/absinthe/schema/notation.ex
+++ b/lib/absinthe/schema/notation.ex
@@ -1486,7 +1486,7 @@ defmodule Absinthe.Schema.Notation do
             quote do
               File.read!(@__absinthe_import_sdl_path__)
             end,
-            Keyword.delete(opts, :path)
+            opts
           ),
           quote do
             @external_resource @__absinthe_import_sdl_path__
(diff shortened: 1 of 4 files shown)


lib/absinthe/schema/notation.ex
```diff
@@ -1150,12 +1150,14 @@ defmodule Absinthe.Schema.Notation do
...
-    do_import_sdl(nil, opts)
+    __CALLER__
+    |> do_import_sdl(nil, opts)
   end
 
   @spec import_sdl(String.t() | Macro.t(), [import_sdl_option()]) :: Macro.t()
   defmacro import_sdl(sdl, opts \\ []) do
-    do_import_sdl(sdl, opts)
+    __CALLER__
+    |> do_import_sdl(sdl, opts)
```
benwilson512: @bruce ah yeah this is definitely the right thing to do here.

PR #636, opened 2018-11-15
Reuse the resolution struct to avoid unnecessary allocations

benwilson512:
# Wide Query
```
Name                     ips        average  deviation         median         99th %
wide-query-new        803.28        1.24 ms    ±20.39%        1.13 ms        1.98 ms
wide-query-old        505.16        1.98 ms    ±11.62%        1.99 ms        2.54 ms

Comparison: 
wide-query-new        803.28
wide-query-old        505.16 - 1.59x slower

Memory usage statistics:

Name              Memory usage
wide-query-new         1.26 MB
wide-query-old         2.29 MB - 1.82x memory usage
```

# Deep Query
```
Name                     ips        average  deviation         median         99th %
deep-query-new         84.07       11.89 ms    ±13.62%       11.65 ms       16.01 ms
deep-query-old         67.71       14.77 ms    ±10.15%       14.56 ms       18.74 ms

Comparison: 
deep-query-new         84.07
deep-query-old         67.71 - 1.24x slower

Memory usage statistics:

Name              Memory usage
deep-query-new         9.00 MB
deep-query-old        14.25 MB - 1.58x memory usage
```

### DIFF
--- a/lib/absinthe/phase/document/execution/resolution.ex
+++ b/lib/absinthe/phase/document/execution/resolution.ex
@@ -84,48 +84,48 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
   This function walks through any existing results. If no results are found at a
   given node, it will call the requisite function to expand and build those results
   """
-  def walk_result(%{fields: nil} = result, bp_node, _schema_type, exec, path) do
-    {fields, exec} = resolve_fields(bp_node, exec, result.root_value, path)
-    {%{result | fields: fields}, exec}
+  def walk_result(%{fields: nil} = result, bp_node, _schema_type, res, path) do
+    {fields, res} = resolve_fields(bp_node, res, result.root_value, path)
+    {%{result | fields: fields}, res}
   end
 
-  def walk_result(%{fields: fields} = result, bp_node, schema_type, exec, path) do
-    {fields, exec} = walk_results(fields, bp_node, schema_type, exec, [0 | path], [])
+  def walk_result(%{fields: fields} = result, bp_node, schema_type, res, path) do
+    {fields, res} = walk_results(fields, bp_node, schema_type, res, [0 | path], [])
 
-    {%{result | fields: fields}, exec}
+    {%{result | fields: fields}, res}
   end
 
-  def walk_result(%Result.Leaf{} = result, _, _, exec, _) do
-    {result, exec}
+  def walk_result(%Result.Leaf{} = result, _, _, res, _) do
+    {result, res}
   end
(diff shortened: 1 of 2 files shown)


lib/absinthe/phase/document/execution/resolution.ex
```diff
@@ -48,11 +48,27 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
...
+    res =
+      %Absinthe.Resolution{
+        path: nil,
+        source: nil,
+        parent_type: nil,
+        middleware: nil,
+        definition: nil,
+        arguments: nil
+      }
+      |> Map.merge(common)
```
michalmuskala:
```elixir
%{adapter: adapter, context: context, acc: acc, root_value: root_value, schema: schema, fragments: fragments, fields_cache: fields_cache} = exec

%Absinthe.Resolution{
  path: nil,
  source: nil,
  parent_type: nil,
  middleware: nil,
  definition: nil,
  arguments: nil,
  adapter: adapter, 
  context: context, 
  acc: acc, 
  root_value: root_value, 
  schema: schema, 
  fragments: fragments, 
  fields_cache: fields_cache
}
```
explicit maps should be even slightly more performant, though quite verbose in here

benwilson512: Yeah, this allocation happens just once per request though, the biggest win was https://github.com/absinthe-graphql/absinthe/pull/636/files#diff-9e88b6c0fafa4e1d518da81f1c8f16a8L195 which is called for every field, potentially thousands of times.

atomkirk: so a resolution is still created on every request? I was noticing how big that struct is and wondered if it would slow things down to be created on every request when you are trying to handle dozens per second on an instance. googling led me here :)

benwilson512: Hi @atomkirk dozens of requests per second should be trivially achieved on a single instance, the presence of large maps is not going to be your bottleneck.

atomkirk: I think it probably also greatly reduces allocations when the vast majority of the keys are atoms. I would assume atoms are shared across all of the vm's processes.

benwilson512: That is correct, structs benefit from the VM's constant heap allocation. Not just the keys, but the entire "key structure" part of the map is allocated once when the application boots, and then referenced.

PR #645, opened 2018-12-04
Flatten the body for the schema module

michalmuskala:
This should speed up compilation and avoid building huge lists that
might blow up the compiler.


benwilson512:
Thanks!

PR #646, opened 2018-12-04
WIP: fragment cycle phase errors for unknown nested fragments

hauptbenutzer:
Running a query like 

```graphql
query {
  currentSession {
    ...knownFragment
  }
}

fragment knownFragment on Session {
  ...unkownFragment
}
```

currently results in `(KeyError) key "unknownFragment" not found in: [...]` at https://github.com/absinthe-graphql/absinthe/blob/master/lib/absinthe/phase/document/validation/no_fragment_cycles.ex#L48. The KnownFragmentName phase currently runs after the NoFragmentCycles phase. But even when flipped, this error will persist, as KnownFragmentName only flags the fragment spread, which is ignored in NoFragmentCycles.

As suggested, I've added a failing test, where I thought it would make the most sense.

### DIFF
--- a/lib/absinthe/phase/document/validation/no_fragment_cycles.ex
+++ b/lib/absinthe/phase/document/validation/no_fragment_cycles.ex
@@ -45,7 +45,8 @@ defmodule Absinthe.Phase.Document.Validation.NoFragmentCycles do
           graph
           |> :digraph_utils.topsort()
           |> Enum.reverse()
-          |> Enum.map(&Map.fetch!(fragments, &1))
+          |> Enum.map(&Map.get(fragments, &1))
+          |> Enum.reject(&is_nil/1)
 
         {fragments, 0}
       end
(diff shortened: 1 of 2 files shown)


lib/absinthe/phase/document/validation/no_fragment_cycles.ex
```diff
@@ -45,7 +45,8 @@ defmodule Absinthe.Phase.Document.Validation.NoFragmentCycles do
           graph
           |> :digraph_utils.topsort()
           |> Enum.reverse()
-          |> Enum.map(&Map.fetch!(fragments, &1))
+          |> Enum.map(&Map.get(fragments, &1))
+          |> Enum.reject(&is_nil/1)
```
hauptbenutzer: So this was the simplest thing I could think of. Not sure, if this is the right approach?

Also, we can do this in one pass, if necessary.

binaryseed:
This looks reasonable, thanks!

PR #647, opened 2018-12-04
Pass :strict option to OptionParser.parse/2

redrabbit:
Removes deprecation warning when calling `Mix.Tasks.Absinthe.Schema.Json.run/1` with Elixir >= 1.7.


benwilson512:
Thanks!

PR #649, opened 2018-12-05
Deep decoration

benwilson512:


benwilson512:
@bruce aight this is safe to merge now, we can refactor how decorators work later but there's too many other branches that expect some commits here to exist to not merge.

PR #652, opened 2018-12-14
Fix __typename prevent subscription deduplication

darrenclark:
This fixes an issue where identical subscription GraphQL queries would not be deduplicated, i.e. would end up with different doc IDs here:

https://github.com/absinthe-graphql/absinthe/blob/d34d5ab682a8800a0831cd2b991d1d5d9dede5d5/lib/absinthe/phase/subscription/subscribe_self.ex#L22-L23

The anonymous function used when resolving `__typename` ends being "different" when run on separate processes, preventing identical subscription queries that use `__typename` from being deduplicated (when run from separate processes)

---

The root cause of the issue:

```elixir
defmodule Example do
  def print(msg), do: IO.puts msg

  def make_print_func(), do: &Example.print/1

  def make_anon_print_func(), do: &print/1
end

a = Example.make_print_func()
b = Task.await(Task.async(&Example.make_print_func/0))

IO.puts "a == b?  #{a == b}"
IO.puts ":erlang.term_to_binary(a) == :erlang.term_to_binary(b)?  #{:erlang.term_to_binary(a) == :erlang.term_to_binary(b)}"

anon_a = Example.make_anon_print_func()
anon_b = Task.await(Task.async(&Example.make_anon_print_func/0))

IO.puts "anon_a == anon_b?  #{anon_a == anon_b}"
IO.puts ":erlang.term_to_binary(anon_a) == :erlang.term_to_binary(anon_b)?  #{:erlang.term_to_binary(anon_a) == :erlang.term_to_binary(anon_b)}"

```

ends up producing:

```
a == b?  true
:erlang.term_to_binary(a) == :erlang.term_to_binary(b)?  true
anon_a == anon_b?  true
:erlang.term_to_binary(anon_a) == :erlang.term_to_binary(anon_b)?  false
```

---

I noticed this issue also happens with `Dataloader` / `Dataloader.Ecto` because of these anonymous functions:

https://github.com/absinthe-graphql/dataloader/blob/a61f550a28820d2befcd94411f7b03163f30db30/lib/dataloader/ecto.ex#L190-L191

(was able to workaround that on my end though, as I just provided non-anonymous functions for `query` and `run_batch` in my code)

---

@benwilson512 How do you feel about:

- Replacing `:crypto.hash(:sha256, :erlang.term_to_binary(blueprint)) |> Base.encode16()`? (maybe to ignore anonymous functions, or simply use the query's string representation, not sure)

and/or 

- Allowing projects to generate their own doc IDs?  I believe the project I work on could do some optimizations specific to our schema

I'd be happy to take on both of those tasks, just wanted to get your thoughts before I started on them

### DIFF
--- a/lib/absinthe/introspection/field.ex
+++ b/lib/absinthe/introspection/field.ex
@@ -12,28 +12,7 @@ defmodule Absinthe.Introspection.Field do
       type: :string,
       description: "The name of the object type currently being queried.",
       middleware: [
-        Absinthe.Resolution.resolver_spec(fn
-          _, %{parent_type: %Type.Object{} = type} ->
-            {:ok, type.name}
-
-          _, %{source: source, parent_type: %Type.Interface{} = iface} = env ->
-            case Type.Interface.resolve_type(iface, source, env) do
-              nil ->
-                {:error, "Could not resolve type of concrete " <> iface.name}
-
-              type ->
-                {:ok, type.name}
-            end
-
-          _, %{source: source, parent_type: %Type.Union{} = union} = env ->
-            case Type.Union.resolve_type(union, source, env) do
-              nil ->
-                {:error, "Could not resolve type of concrete " <> union.name}
-
-              type ->
-                {:ok, type.name}
-            end
-        end)
+        Absinthe.Resolution.resolver_spec(&__MODULE__.typename_resolver/2)
       ]
     }
   end
@@ -70,5 +49,29 @@ defmodule Absinthe.Introspection.Field do
         end)
       ]
     }
+  end
+
+  def typename_resolver(_, %{parent_type: %Type.Object{} = type}) do
+    {:ok, type.name}
+  end
+
+  def typename_resolver(_, %{source: source, parent_type: %Type.Interface{} = iface} = env) do
+    case Type.Interface.resolve_type(iface, source, env) do
+      nil ->
+        {:error, "Could not resolve type of concrete " <> iface.name}
+
+      type ->
+        {:ok, type.name}
+    end
+  end
+
+  def typename_resolver(_, %{source: source, parent_type: %Type.Union{} = union} = env) do
+    case Type.Union.resolve_type(union, source, env) do
+      nil ->
+        {:error, "Could not resolve type of concrete " <> union.name}
+
+      type ->
+        {:ok, type.name}
+    end
   end
 end
--- a/test/absinthe/execution/subscription_test.exs
+++ b/test/absinthe/execution/subscription_test.exs
@@ -264,6 +264,36 @@ defmodule Absinthe.Execution.SubscriptionTest do
     assert String.contains?(error_log, "boom")
   end
 
+  @query """
+  subscription { user { group { name } id } }
+  """
+  test "identical subscription docs are deduplicated" do
+    opts = [context: %{test_pid: self()}]
+
+    assert {:ok, %{"subscribed" => doc1}} =
+             run(@query, Schema, opts)
+
+    assert {:ok, %{"subscribed" => doc2}} =
+             run(@query, Schema, opts)
+
+    assert doc1 == doc2
+  end
+
+  @query """
+  subscription { user { __typename group { name } id } }
+  """
+  test "identical subscription docs on different processes are deduplicated" do
+    opts = [context: %{test_pid: self()}]
+
+    task1 = Task.async(fn -> run(@query, Schema, opts) end)
+    task2 = Task.async(fn -> run(@query, Schema, opts) end)
+
+    assert {:ok, %{"subscribed" => doc1}} = Task.await(task1)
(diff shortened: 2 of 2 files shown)


benwilson512:
Hey @darrenclark thanks for taking a look at this!

In Absinthe 1.5 what I want to do is focus on letting people generate their own doc ids. The current system has a couple of flaws, one of which you've found.

My thought is that we'd probably eliminate "by default" deduplication. If you want to have things deduplicated then you need to provide some kind of identifier. Identifiers really have two parts:

1) The document itself. A sha256 of the original query string is probably OK here and I'm not really sure it needs to be configurable. I want to avoid giving people obvious foot guns.
2) The context. This is the part we would require from people. Right now I do the term to binary and sha but depending on the circumstance this is often too narrow. It's too narrow in the sense that term equality can fail at times when there's still a logical equality (two different loads of the user out of the database may not have term equality even though they're the same user). We also want to allow people to have control over deduplication. IE perhaps the context id can just be `"logged-in"` for any valid user instead of always using the current user id. This provides substantial performance improvements.

A PR that works on these two things would be incredible.

darrenclark:
@benwilson512 Perfect, we can definitely look into this.  It should open up some significant optimization opportunities for us.

Any thoughts on the interface into this?  Something similar to the document adapter interface (https://hexdocs.pm/absinthe/adapters.html)?  `Absinthe.Subscription.DocumentIdentification` or something?

With that in mind, should I close this PR?  Given the plans for 1.5, I'm not sure it adds anything of value.

benwilson512:
@darrenclark yeah I think we should drop this particular PR in favor of the above approach.

for the current user style id I was thinking it could be as simple as adding a key to the return value of the subscription `config` function IE:

```
config args, %{context: %{current_user: _}} ->
  {:ok, topic: args.whatever, id: "authenticated"}
end
```

Then the full subscription id would be `"authenticated:#{sha256(variables_json <> query_string)}"`. Do we need to make the document id be configurable? Are there situations where two documents that don't have the same query string sha should be treated as identical?

darrenclark:
I have a contrived example of what I was hoping to support:

```graphql
subscription PriceChange(productIds: [ID!]!) {
  priceDidChange(productIds: $ids) {
    product {
      __typename
      id
      name
      price
    }
  }
}
```

producing a stream of JSON objects something like:

```json
{
  "event": {
     "__typename": "Product",
     "id": "Product:3242",
     "price": 10.5
  }
}

{
  "event": {
     "__typename": "Product",
     "id": "Product:12",
     "price": 11.25
  }
}

{
  "event": {
     "__typename": "Product",
     "id": "Product:924",
     "price": 5.05
  }
}
```

I was hoping that we'd be able to give all these documents the same ID:

```graphql
# only whitespace differs
subscription PriceChange(productIds: [ID!]!) {
  priceDidChange(productIds: $ids) { product { __typename id name price } }
}

# same fields, different order
subscription PriceChange(productIds: [ID!]!) {
  priceDidChange(productIds: $ids) { product { id __typename price name } }
}

# different operation name / variables
subscription PriceDidChangeSubscription(ids: [ID!]!) {
  priceDidChange(productIds: $ids) { product { __typename id name price } }
}

# fragments
subscription PriceChange(productIds: [ID!]!) {
  priceDidChange(productIds: $ids) { product { ...ProductInfo } }
}
fragment ProductInfo on Product { id __typename price name }
``` 

It'd also be nice for us to be able to ignore the arguments for the subscription `priceDidChange` field (though maybe not all subscription fields in our schema), and rely on the provided `topic:`s to ensure each client got the data updates they were interested in.

Any thoughts on:

- no `id:` provided, uses existing functionality
- `id: "authenticated"` provided, does `"authenticated:#{sha256(variables_json <> query_string)}"`
- when value provided to `id:` is a function, simply calls it to get the document ID

darrenclark:
Or rather, it would be nice to have the option to have full control over the IDs being generated, so that we could account for those situations.

I can only really see us accounting for those situations if we see a lot of duplication due to minor changes like those.

darrenclark:
> And if you have a serializer that respects the order of the fields as queried then they really ARE different

Ah yes, good to know (I had no idea [they mentioned it in the spec](https://facebook.github.io/graphql/draft/#sec-Serialized-Map-Ordering) hah)

> However, Absinthe can sort of punt on these questions I suppose if we also allow a document_id: option that lets the user do whatever they wish:

The final ID would end up being `"#{id}:#{document_id}"` then?  Users would provide `id:` and `document_id:` options?

This would work well for our use case

benwilson512:
> The final ID would end up being "#{id}:#{document_id}" then? Users would provide id: and document_id: options?

Yes, which we'll refer to as the `subscription_id`. It may make sense to think of the subscription id as the composite of a context_id and a document_id. Having `id` simpliciter may be confusing.

benwilson512:
If no `document_id` is provided Absinthe will just hash the query string + variables. If no context_id is provided we could probably get away with `:erlang.unique_integer`.

PR #654, opened 2018-12-18
Adjust how subscription IDs are generated

darrenclark:
First pass at the changes described in https://github.com/absinthe-graphql/absinthe/pull/652

If things look good, I'll go update all documentation, fix conflicts, etc.

I also need to add the blueprint to the resolution struct (https://github.com/absinthe-graphql/absinthe/pull/652#issuecomment-447462852)

---

Subscription IDs are now derived via a `context_id` option returned by a subscriptions `config` function, combined with the GraphQL document and any provided variables.

This gives library users more control over which subscription queries get merged.

If `context_id` is not provided, a unique integer is generated instead (to ensure we don't merge any queries by default).  

- We could also do the `:erlang.term_to_binary` + SHA256 hash dance on the `context` by default instead.  It _should_ be safe to merge subscriptions whose `context` is the same.. **Thoughts?**

Additionally, if one wants to override the default behaviour of combining the GraphQL document & provided variables to determine the `document_id`, they can return a `document_id` from the subscription's `config` function

### DIFF
--- a/guides/subscriptions.md
+++ b/guides/subscriptions.md
@@ -196,4 +196,49 @@ If you want to subscribe to mutations from within your application, you can do:
 MyAppWeb.Endpoint.subscribe(topic)
 ```
 
+### De-duplicating Updates
+
+By default, Absinthe will resolve each outgoing publish once per individual subscription.  This ensures:
+
+- different GraphQL documents each receive the different fields they requested
+- user-specific updates are sent out, in case `context` contains user-specific data
+
+To improve the scale at which your subscriptions operate, you may tell Absinthe when it is safe to de-duplicate updates.  Simply return a `context_id` from your field's `config` function.
+
+```elixir
+subscription do
+  field :news_article_published, :article do
+    config fn _, _ ->
+      {:ok, topic: "*", context_id: "global"}
+    end
+  end
+end
+```
+
+Here we return a constant (`"global"`) because our `:article` type doesn't contain any user-specific fields on it.
+
+Given these three active subscriptions:
+
+```graphql
+# user 1
+subscription {
+  newsArticlePublished { content }
+}
+
+# user 2
+subscription {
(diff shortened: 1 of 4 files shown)


arkgil:
@darrenclark Hi 👋 

> We could also do the :erlang.term_to_binary + SHA256 hash dance on the context by default instead. It should be safe to merge subscriptions whose context is the same.. Thoughts?

I'm wondering if you could simply rely on the `:topic` returned by the `config` function? I would expect that retuning the same topic would tap me into the same stream of events.

arkgil:
@benwilson512 right, I am aware of it, I wasn’t really clear with my comment. What I meant is, with the approach presented in this PR Subscriptions attached to the same topic are duplicated by default, and context_id needs to be explicitly set if you want them deduplicated. I was wondering whether having them deduplicated given the same topic would make sense.

Or I’m just talking nonsense 😄 Anyway, thanks for clarification 🙂

benwilson512:
@arkadiyk Imagine you have 10 documents submitted by 10 different users, each with the user set in the context. If you dedup only on the bodies, then any data loading that happens which takes into account the current user becomes horribly broken. Only the first document is run, which loads stuff that may be exclusive to just that user, and then it gets broadcasted to everyone.

darrenclark:
Regarding the status of this PR:  Been busy with other things, but hoping to loop back to this in the next week or two

darrenclark:
@benwilson512 I (finally) got this updated, it's ready for review now.

I did not document (outside of the tests for it) the `document_id` option.  Figured it'd be best, so as to avoid confusion.  Thoughts?   I figure if anyone goes looking for this functionality, they will end up digging into Absinthe's source code and discover it there.

@arkgil If you get a chance, can you let me know if this addresses #675

benwilson512:
@darrenclark This is really fantastic, thank you!

PR #657, opened 2018-12-27
Use Jason codec

maartenvanvliet:
Jason has quickly become the default Json-codec in Elixir. This PR uses it in the example docs for configuring absinthe_plug and sets it as the default codec for the cli tool


benwilson512:
Yeah I've been meaning to do this for a while, thanks!

PR #658, opened 2018-12-27
Add function docs for pipeline module

maartenvanvliet:
Added docs with examples for several pipeline manipulation functions. Tested the examples with a doctest so they should work.


benwilson512:
Thanks!

PR #659, opened 2018-12-27
 Add documentation for batch timeout option

maartenvanvliet:
The timeout option was missing in the docs for the batch helper. Also added the option to the typespecs


benwilson512:
Thanks!

PR #660, opened 2018-12-28
docs: Update references query_arguments

scott-ad-riley:
Updates the references in the `Query Arguments` tutorial/guide to the correct method name/arity.


benwilson512:
Thanks!

PR #663, opened 2019-01-05
Telemetry integration

### LINKED ISSUE #662: Telemetry integration & design
Hello! I recently released a basic New Relic integration with Absinthe:

* https://github.com/binaryseed/new_relic_absinthe/blob/master/lib/new_relic/absinthe/middleware.ex

It works fine as a `Middleware`, but it'd be really sweet to have a `Telemetry` integration that could be leveraged generically, and it'd make the instrumentation automatic instead of requiring a custom middleware stack. I'm working on a New Relic telemetry exporter and have been keeping a close eye on how the `Telemetry` project evolves. 

* https://github.com/beam-telemetry/telemetry

I'd like to offer to build into Absinthe a telemetry integration. To get started, we should sketch out a design of what the events might look like!

Possible events:
* `[:absinthe, :resolver, :call]` - Triggered for each resolver that calls a function (ie: not the default `Map.get`)

```elixir
metadata = %{
  type: :absinthe_resolver_call,
  schema: inspect(res.schema),
  type: Absinthe.Type.name(res.definition.schema_node.type, res.schema),
  field_name: res.definition.name,
  query_path: Absinthe.Resolution.path(res) |> Enum.join("."),
  parent_type: res.parent_type.name
}
```

* `[:absinthe, :resolution, :complete]` - Triggered at the end of a completed query resolution

```elixir
metadata = %{
  type: :absinthe_resolution_complete,
  schema: inspect(res.schema),
  query_complexity: operation.complexity,
  operation_type: operation.type,
  operation_name: operation.name
}
```

This is just pulled from what I have in the New Relic middleware, hoping it's a good starting off point to discuss...

-------

@benwilson512 @bruce 
cc/ @arkgil @josevalim

binaryseed:
This PR builds out [Telemetry](https://github.com/beam-telemetry/telemetry) integration for Absinthe.

This integration executes the following events:

* `[:absinthe, :execute, :operation]` Upon completion of the entire operation pipeline
* `[:absinthe, :resolve, :field]` Upon completion of each defined resolver function call
* `[:absinthe, :subscription, :publish]` Upon publication in a subscription

To aid in tracing, we also execute events at the start of each of these ^:

* `[:absinthe, :execute, :operation, :start]`
* `[:absinthe, :resolve, :field, :start]`
* `[:absinthe, :subscription, :publish, :start]`

Notes:

* The default `Map.get` resolver doesn't execute events

closes #662

### DIFF
--- a/lib/absinthe/middleware/telemetry.ex
+++ b/lib/absinthe/middleware/telemetry.ex
@@ -2,34 +2,51 @@ defmodule Absinthe.Middleware.Telemetry do
   @moduledoc """
   Gather and report telemetry about an individual field resolution
   """
+  @telemetry_event_start [:absinthe, :resolve, :field, :start]
   @telemetry_event [:absinthe, :resolve, :field]
 
   @behaviour Absinthe.Middleware
 
   @impl Absinthe.Middleware
   def call(resolution, _) do
-    on_complete =
-      {{__MODULE__, :on_complete},
-       [
-         start_time: System.system_time(),
-         start_time_mono: System.monotonic_time(),
-         middleware: resolution.middleware
-       ]}
+    id = :erlang.unique_integer()
+    start_time = System.system_time()
+    start_time_mono = System.monotonic_time()
 
-    %{resolution | middleware: resolution.middleware ++ [on_complete]}
+    :telemetry.execute(@telemetry_event_start, %{start_time: start_time}, %{id: id})
+
+    %{
+      resolution
+      | middleware:
+          resolution.middleware ++
+            [
+              {{__MODULE__, :on_complete},
+               %{
+                 id: id,
+                 start_time: start_time,
+                 start_time_mono: start_time_mono,
+                 middleware: resolution.middleware
+               }}
+            ]
+    }
   end
 
-  def on_complete(%{state: :resolved} = resolution,
-        start_time: start_time,
-        start_time_mono: start_time_mono,
-        middleware: middleware
+  def on_complete(
+        %{state: :resolved} = resolution,
+        %{
+          id: id,
+          start_time: start_time,
+          start_time_mono: start_time_mono,
+          middleware: middleware
+        }
       ) do
+    end_time_mono = System.monotonic_time()
+
     :telemetry.execute(
       @telemetry_event,
+      %{duration: end_time_mono - start_time_mono},
       %{
-        duration: System.monotonic_time() - start_time_mono
-      },
-      %{
+        id: id,
         start_time: start_time,
         middleware: middleware,
         resolution: resolution
--- a/lib/absinthe/middleware/telemetry.ex
+++ b/lib/absinthe/middleware/telemetry.ex
@@ -1,8 +1,8 @@
 defmodule Absinthe.Middleware.Telemetry do
   @moduledoc """
-  Gather and report telemetry about an individual resolver function
+  Gather and report telemetry about an individual field resolution
   """
-  @telemetry_event [:absinthe, :resolver]
+  @telemetry_event [:absinthe, :resolve, :field]
 
   @behaviour Absinthe.Middleware
 
--- a/lib/absinthe/middleware/telemetry.ex
+++ b/lib/absinthe/middleware/telemetry.ex
@@ -2,8 +2,8 @@ defmodule Absinthe.Middleware.Telemetry do
   @moduledoc """
   Gather and report telemetry about an individual field resolution
   """
-  @telemetry_event_start [:absinthe, :resolve, :field, :start]
-  @telemetry_event [:absinthe, :resolve, :field]
+  @field_start [:absinthe, :resolve, :field, :start]
+  @field [:absinthe, :resolve, :field]
 
   @behaviour Absinthe.Middleware
 
@@ -13,7 +13,7 @@ defmodule Absinthe.Middleware.Telemetry do
     start_time = System.system_time()
     start_time_mono = System.monotonic_time()
 
-    :telemetry.execute(@telemetry_event_start, %{start_time: start_time}, %{id: id})
+    :telemetry.execute(@field_start, %{start_time: start_time}, %{id: id})
 
     %{
       resolution
@@ -43,7 +43,7 @@ defmodule Absinthe.Middleware.Telemetry do
     end_time_mono = System.monotonic_time()
 
     :telemetry.execute(
-      @telemetry_event,
+      @field,
       %{duration: end_time_mono - start_time_mono},
       %{
         id: id,
--- a/lib/absinthe/middleware/telemetry.ex
+++ b/lib/absinthe/middleware/telemetry.ex
@@ -7,24 +7,22 @@ defmodule Absinthe.Middleware.Telemetry do
   @behaviour Absinthe.Middleware
 
   @impl Absinthe.Middleware
-  def call(%{middleware: [{{Absinthe.Resolution, :call}, resolver_fun} | _]} = resolution, _) do
+  def call(resolution, _) do
     on_complete =
       {{__MODULE__, :on_complete},
        [
          start_time: System.system_time(),
          start_time_mono: System.monotonic_time(),
-         resolver_fun: resolver_fun
+         middleware: resolution.middleware
        ]}
 
     %{resolution | middleware: resolution.middleware ++ [on_complete]}
   end
 
-  def call(resolution, _config), do: resolution
-
   def on_complete(%{state: :resolved} = resolution,
         start_time: start_time,
         start_time_mono: start_time_mono,
-        resolver_fun: resolver_fun
+        middleware: middleware
       ) do
     :telemetry.execute(
       @telemetry_event,
@@ -33,7 +31,7 @@ defmodule Absinthe.Middleware.Telemetry do
       },
       %{
         start_time: start_time,
-        resolver_fun: resolver_fun,
+        middleware: middleware,
         resolution: resolution
       }
     )
--- a/lib/absinthe/middleware/telemetry.ex
+++ b/lib/absinthe/middleware/telemetry.ex
@@ -1,0 +1,56 @@
+defmodule Absinthe.Middleware.Telemetry do
+  @moduledoc """
+  Gather and report telemetry about an individual resolver function
+  """
+  @telemetry_event [:absinthe, :resolver]
+
+  @behaviour Absinthe.Middleware
+
+  @impl Absinthe.Middleware
+  def call(%{middleware: [{{Absinthe.Resolution, :call}, resolver_fn} | _]} = res, _config) do
+    on_complete = [
+      {{__MODULE__, :on_complete},
+       [
+         start_time: System.system_time(),
+         start_time_mono: System.monotonic_time(),
+         resolver_fn: resolver_fn
+       ]}
+    ]
+
+    %{res | middleware: res.middleware ++ on_complete}
+  end
+
+  def call(res, _config), do: res
+
+  def on_complete(%{state: :resolved} = res,
+        start_time: start_time,
+        start_time_mono: start_time_mono,
+        resolver_fn: resolver_fn
+      ) do
+    :telemetry.execute(
+      @telemetry_event,
+      %{
+        start_time: start_time,
+        duration: System.monotonic_time() - start_time_mono
+      },
+      %{
+        schema: res.schema,
+        mfa: resolver_mfa(resolver_fn),
+        arguments: res.arguments,
+        path: Absinthe.Resolution.path(res),
+        field_name: res.definition.name,
+        field_type: Absinthe.Type.name(res.definition.schema_node.type, res.schema),
+        parent_type: res.parent_type.name
+      }
+    )
+
+    res
+  end
+
+  defp resolver_mfa({mod, fun}), do: {mod, fun, 3}
+
+  defp resolver_mfa(fun) when is_function(fun) do
+    info = :erlang.fun_info(fun)
+    {info[:module], info[:name], info[:arity]}
+  end
+end
--- a/lib/absinthe/middleware/telemetry.ex
+++ b/lib/absinthe/middleware/telemetry.ex
@@ -7,50 +7,37 @@ defmodule Absinthe.Middleware.Telemetry do
   @behaviour Absinthe.Middleware
 
   @impl Absinthe.Middleware
-  def call(%{middleware: [{{Absinthe.Resolution, :call}, resolver_fn} | _]} = res, _config) do
-    on_complete = [
+  def call(%{middleware: [{{Absinthe.Resolution, :call}, resolver_fun} | _]} = resolution, _) do
+    on_complete =
       {{__MODULE__, :on_complete},
        [
          start_time: System.system_time(),
          start_time_mono: System.monotonic_time(),
-         resolver_fn: resolver_fn
+         resolver_fun: resolver_fun
        ]}
-    ]
 
-    %{res | middleware: res.middleware ++ on_complete}
+    %{resolution | middleware: resolution.middleware ++ [on_complete]}
   end
 
-  def call(res, _config), do: res
+  def call(resolution, _config), do: resolution
 
-  def on_complete(%{state: :resolved} = res,
+  def on_complete(%{state: :resolved} = resolution,
         start_time: start_time,
         start_time_mono: start_time_mono,
-        resolver_fn: resolver_fn
+        resolver_fun: resolver_fun
       ) do
     :telemetry.execute(
       @telemetry_event,
       %{
-        start_time: start_time,
         duration: System.monotonic_time() - start_time_mono
       },
       %{
-        schema: res.schema,
-        mfa: resolver_mfa(resolver_fn),
-        arguments: res.arguments,
-        path: Absinthe.Resolution.path(res),
-        field_name: res.definition.name,
-        field_type: Absinthe.Type.name(res.definition.schema_node.type, res.schema),
-        parent_type: res.parent_type.name
+        start_time: start_time,
+        resolver_fun: resolver_fun,
+        resolution: resolution
       }
     )
 
-    res
-  end
-
-  defp resolver_mfa({mod, fun}), do: {mod, fun, 3}
-
-  defp resolver_mfa(fun) when is_function(fun) do
-    info = :erlang.fun_info(fun)
-    {info[:module], info[:name], info[:arity]}
+    resolution
   end
 end
--- a/lib/absinthe/phase/parse.ex
+++ b/lib/absinthe/phase/parse.ex
@@ -23,7 +23,13 @@ defmodule Absinthe.Phase.Parse do
   end
 
   def run(input, options) do
-    run(%Absinthe.Blueprint{input: input}, options)
+    telemetry = %{
+      start_time: System.system_time(),
+      start_time_mono: System.monotonic_time(),
+      source: input
+    }
+
+    run(%Absinthe.Blueprint{input: input, telemetry: telemetry}, options)
   end
 
   defp add_validation_error(bp, error) do
--- a/lib/absinthe/phase/parse.ex
+++ b/lib/absinthe/phase/parse.ex
@@ -25,11 +25,10 @@ defmodule Absinthe.Phase.Parse do
   def run(input, options) do
     telemetry = %{
       start_time: System.system_time(),
-      start_time_mono: System.monotonic_time(),
-      source: input
+      start_time_mono: System.monotonic_time()
     }
 
-    run(%Absinthe.Blueprint{input: input, telemetry: telemetry}, options)
+    run(%Absinthe.Blueprint{input: input, source: input, telemetry: telemetry}, options)
   end
 
   defp add_validation_error(bp, error) do
--- a/lib/absinthe/phase/parse.ex
+++ b/lib/absinthe/phase/parse.ex
@@ -24,12 +24,7 @@ defmodule Absinthe.Phase.Parse do
   end
 
   def run(input, options) do
-    telemetry = %{
-      start_time: System.system_time(),
-      start_time_mono: System.monotonic_time()
-    }
-
-    run(%Absinthe.Blueprint{input: input, source: input, telemetry: telemetry}, options)
+    run(%Absinthe.Blueprint{input: input}, options)
   end
 
   defp add_validation_error(bp, error) do
--- a/lib/absinthe/phase/parse.ex
+++ b/lib/absinthe/phase/parse.ex
@@ -5,7 +5,8 @@ defmodule Absinthe.Phase.Parse do
 
   alias Absinthe.{Language, Phase}
 
-  @spec run(Language.Source.t(), Keyword.t()) :: Phase.result_t()
+  @type input_t :: Language.Source.t() | Blueprint.t()
+  @spec run(input_t, Keyword.t()) :: Phase.result_t()
   def run(input, options \\ [])
 
   def run(%Absinthe.Blueprint{} = blueprint, options) do
--- a/lib/absinthe/phase/subscription/subscribe_self.ex
+++ b/lib/absinthe/phase/subscription/subscribe_self.ex
@@ -28,7 +28,8 @@ defmodule Absinthe.Phase.Subscription.SubscribeSelf do
       for field_key <- field_keys,
           do: Absinthe.Subscription.subscribe(pubsub, field_key, subscription_id, blueprint)
 
-      {:replace, blueprint, [{Phase.Subscription.Result, topic: subscription_id}]}
+      {:replace, blueprint,
+       [{Phase.Subscription.Result, topic: subscription_id}, {Phase.Telemetry, options}]}
     else
       {:error, error} ->
         blueprint = update_in(blueprint.execution.validation_errors, &[error | &1])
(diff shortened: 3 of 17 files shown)


binaryseed:
Functionality is done, open to feedback about code organization, events, etc!

lib/absinthe/phase/document/execution/resolution.ex
```diff
@@ -280,6 +282,27 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
...
+        schema: res.schema,
+        path: resolution_path,
+        field_name: res.definition.name,
+        field_type: Absinthe.Type.name(res.definition.schema_node.type, res.schema),
+        parent_type: res.parent_type.name
+      }
+
+      :telemetry.execute([:absinthe, :resolver, :call], duration, metadata)
+    end
+  end
```
benwilson512: I'm really not sure that I'm comfortable making telemetry non optional this way. This function runs for absolutely every field in the entire schema. There's nothing stopping someone from adding telemetry middleware to every field, that's a one liner with `def middleware` or with 1.5 decorators. Baking it into resolution makes it impossible to remove without forking the resolution phase.

lib/absinthe/resolution.ex
```diff
@@ -200,6 +200,16 @@ defmodule Absinthe.Resolution do
...
 
   def call(%{state: :unresolved} = res, resolution_function) do
+    res = %{
+      res
+      | acc:
+          Map.put_new(
+            res.acc,
+            {:telemetry, path(res)},
+            {System.system_time(:millisecond), System.monotonic_time()}
+          )
```
benwilson512: Ah ok so the start time is only actually set when resolution functions are run. In this case what you should do is have this middleware to do `res = %{res | middleware ++ [EmitResolvedTelemetry]}` or something (find a better name). Then it'll only run telemetry emission for fields that are guaranteed to have the start time recorded, and then it doesn't need to be baked into resolution.

lib/absinthe/pipeline.ex
```diff
@@ -48,6 +48,7 @@ defmodule Absinthe.Pipeline do
       {Phase.Parse, options},
       # Convert to Blueprint
       {Phase.Blueprint, options},
+      {Phase.Telemetry, event: [:absinthe, :resolution, :start]},
```
benwilson512: We may want to specifically emit telemetry about the resolution phase. That does get tricky since resolution can run multiple times over the course of executing a document, but for example putting the start this far up the pipeline won't work with subscriptions which store the cached documents at a far later level of processing.

lib/absinthe/resolution.ex
binaryseed: Right, this is what I do in the New Relic Absinthe middleware. I baked it in here because I wanted to make it built-in, ie: w/o needing to define a custom middleware stack in your schema.

Is there a way I can build in this middleware into all resolvers? It'd only do anything for `Resolution.call` based resolvers, so it'd skip `Map.get` etc..

https://github.com/binaryseed/new_relic_absinthe/blob/687e4fe7d152ba95e6af4e8b021309658f452d10/lib/new_relic/absinthe/middleware.ex#L6-L26

lib/absinthe/pipeline.ex
binaryseed: I'm open to suggestions here. I could even go w/o the explicit `start` event, the minimum we need is simply to track the `start_time` to be able to calculate the `duration`

lib/absinthe/phase/telemetry.ex
```diff
@@ -0,0 +1,30 @@
...
+
+    {:ok, %{blueprint | execution: execution}}
+  end
+
+  def run(blueprint, event: [:absinthe, :pipeline, :end]) do
+    with {start_time, start_time_mono} <- blueprint.execution.acc[:telemetry_phase] do
+      end_time_mono = System.monotonic_time()
+      duration = System.convert_time_unit(end_time_mono - start_time_mono, :native, :millisecond)
+
+      :telemetry.execute([:absinthe, :pipeline, :end], duration, %{start_time: start_time})
```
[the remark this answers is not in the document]
binaryseed: Those are both included in later commits..

lib/absinthe/phase/document/execution/resolution.ex
binaryseed: Do you want it configurable in general, or is this concern simply about preventing it from running for every single field resolution? (which I also agree is critical to avoid)

lib/absinthe/resolution.ex
binaryseed: I found a place to stick an extra middleware on each resolver, so I pushed a commit with that. Not sure if that's the right location but it isolates all the Telemetry stuff into a module.

lib/absinthe/middleware/telemetry.ex
```diff
@@ -0,0 +1,57 @@
...
+      | middleware:
+          res.middleware ++
+            [
+              {{__MODULE__, :complete},
+               [
+                 start_time: start_time,
+                 start_time_mono: start_time_mono,
+                 resolver_fn: resolver_fn
+               ]}
+            ]
```
benwilson512: 💯

lib/absinthe/phase/document/execution/resolution.ex
```diff
@@ -180,7 +180,7 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
 
   # bp_field needs to have a concrete schema node, AKA no unions or interfaces
   defp do_resolve_field(res, source, path) do
-    res
+    call_middleware({Absinthe.Middleware.Telemetry, []}, res)
```
benwilson512: Right, this is actually kind of tricky. The hard part is two fold:

1) do_resolve_field is called on all fields, so we're back to having middleware run on all fields instead of just on fields with "full" resolvers
2) If however we simply:
```
defmacro resolve(fun) do
  quote do
    middleware Telemetry
    middleware Absinthe.Resolution, unquote(fun)
  end
end
```
this gets us most of the way there, but we run into trouble the minute people start messing with the middleware stack. For example if they add an auth middleware to the front of all resolvers on the root query object (a very common pattern) then the telemetry won't actually pick up the runtime cost of auth.

I should also note that we can't try to add middleware dynamically in a document blueprint phase due to how abstract types work. If a type is abstract then it the concrete fields with middleware won't actually be fetched until the resolution phase when we can determine the type, so consequently it isn't possible to rework the middleware list prior to resolution.

HOWEVER we may still have some options. I think we can do this when the schema gets built. If we go https://github.com/absinthe-graphql/absinthe/blob/master/lib/absinthe/schema/notation.ex#L1665 and in that third clause look through the list of middleware to see if `Absinthe.Resolution` is one of them, then we could prepend Telemetry to the front. We could even actually append the complete one as well and thus avoid the added complexity within the Resolution middlware.

I probably should have suggested this earlier but I only just thought of it!

binaryseed: I put the middleware in `__ensure_middleware__` but I didn't split it into 2 middleware since it's actually very handy to mark the start_time in the first and send it in as options to the second

benwilson512: Yup that's totally fine. I thought of the perfect way to make this apply to ONLY fields that use `resolve` while also making it possible to opt out.

Change https://github.com/absinthe-graphql/absinthe/blob/master/lib/absinthe/schema/notation.ex#L528 to 
```
defmacro resolve(fun_ast) do
  quote do
    meta :absinthe_telemetry, true
    middleware Absinthe.Resolution, unquote(fun_ast)
  end
end
```

And then in the `__ensure_middleware__` you can

```
def __ensure_middleware__(middleware, field, object) do
  if Absinthe.Type.meta(field, :absinthe_telemetry) do
    # prepend telemetry
  else
    middleware
  end
end
```

What's nice about this is that
1) By default it'll only happen for `resolve` based fields
2) However anyone could opt in easily
2) Opting out is easy too, you just configure `:absinthe_telemetry false` via any of the schema modification mechanisms.

binaryseed: I can do that, however the opt out would only be partial, just for that specific resolver function, which is only a small part of the overall telemetry events.

I think telemetry is intended to be very low impact, so if there is no handler attached it's pretty much just an ets lookup and a function call or two

binaryseed: ok, updated!

lib/absinthe/pipeline.ex
binaryseed: What other telemetry about the resolution phase do you have in mind?

lib/absinthe/phase/telemetry.ex
```diff
@@ -0,0 +1,41 @@
+defmodule Absinthe.Phase.Telemetry do
+  @moduledoc false
+
+  use Absinthe.Phase
+
+  def run(blueprint, event: [:absinthe, :resolution, :start]) do
+    start_time = System.system_time(:millisecond)
+    start_time_mono = System.monotonic_time()
+    metadata = %{type: :absinthe_resolution_start}
```
lswith: I'd love to see the query/mutation name and type in the metadata here. A function to get this information from the blueprint is below:
```
def get_operation_metadata(blueprint) do
    case Blueprint.current_operation(blueprint) do
        nil -> {name: "", type: ""}
        op -> {name: op.name, type: op.type}
     end
end
```

binaryseed:
Updated to `:telemetry` `0.4.0` & rebased on master

lib/absinthe/phase/document/execution/resolution.ex
garthk: Would using a macro to make that choice configurable at compile time like [` compile_time_purge_matching`](https://hexdocs.pm/logger/Logger.html#module-application-configuration) mean you'd then have [two problems](http://regex.info/blog/2006-09-15/247)?

binaryseed:
I've simplified the code and event names a bit & arranged it so that we can report both the query and the variables / arguments in the metadata.

* Added a `telemetry` key to the blueprint struct rather than pretending to be a plugin and using `acc`
* Removed the explicit phase & event at the start of the pipeline in favor of putting timing code inside the very first phase (which means `duration` now includes parse time which is important)

@bruce i'm sad that my force-pushing habit is now plain for everyone to see ;)

binaryseed:
I may need to update the middleware based on this fix demonstrated in https://github.com/binaryseed/new_relic_absinthe/pull/6

binaryseed:
@benwilson512 @bruce ... I was reviewing other telemetry integrations and remembered that `plug` simply sticks the whole `conn` in the metadata of the event. 

We could put the `blueprint` / `resolution` structs in the events we emit. Right now I'm unpacking some select details, but really I did that because that's what my New Relic adapter does. However, I can totally imagine use cases that want other data, and to support them we might play whack-a-mole adding things to the telemetry event.

The downside is the fact that those structs are pretty huge. Telemetry handlers are called inline so they won't be copied across processes by default, but a handler might unintentionally cause a lot of memory copying (granted any middleware / resolver can already run into this)

Any thoughts? I'm inclined to just put the structs in there whole.

garthk:
Having caught up on [Arkadiusz Gil's launch video](https://www.youtube.com/watch?v=cOuyOmcDV1U), strikes me you might need a bit of both:
* Supply the whole struct in the metadata map supplied as the third argument to [`:telemetry.execute/3`](https://hexdocs.pm/telemetry/telemetry.html#execute-3), _and_
* Pull out important measurements and tags to the measurements and metadata maps (second and third) so people can apply [`Telemetry.Metrics`](https://hexdocs.pm/telemetry_metrics/0.2.1/overview.html) without wading through your data structures enough to educate extraction functions for [`t: Telemetry.Metrics.tag_values/0`](https://hexdocs.pm/telemetry_metrics/0.2.1/Telemetry.Metrics.html#t:tag_values/0).

benwilson512:
@garthk that does sound like the way to go. cc: @binaryseed

binaryseed:
That's pretty much where the PR is now..

I could put back a few of the common fields to use as tags just to make it easier, though we could consider providing some built-in "extractor" functions that could be applied to the structs - i think that might be a little more flexible in the long run

lib/absinthe/middleware/telemetry.ex
```diff
@@ -0,0 +1,43 @@
+defmodule Absinthe.Middleware.Telemetry do
+  @moduledoc """
+  Gather and report telemetry about an individual resolver function
+  """
+  @telemetry_event [:absinthe, :resolver]
+
+  @behaviour Absinthe.Middleware
+
+  @impl Absinthe.Middleware
+  def call(%{middleware: [{{Absinthe.Resolution, :call}, resolver_fun} | _]} = resolution, _) do
```
benwilson512: @binaryseed It occurs to me that pattern matching on the resolution middleware being the first may be a problem. Consider:
```
field :foo, :foo do
  middleware Auth # or NodeId.Parser or w/e
  resolve &foo/1
end
```

Perhaps instead of `resolver_fun` you really just want the entire rest of the middleware stack?

binaryseed: Yeah, the issue I linked to above (https://github.com/binaryseed/new_relic_absinthe/pull/6) points this out too..

benwilson512: Sure, in that PR though it's extracting just the resolver fun. I'm suggesting pushing the whole stack through Telemetry, and then if a telemetry handler only wants the resolver fun they can extract it from that list. The whole middleware stack is important though because all of the middleware can do IO and take time.

binaryseed: OK, I think I get what you are suggesting now.. updated!

benwilson512: perfect, thanks! This seems good to go from my view, am I missing anything?

lib/absinthe/phase/telemetry.ex
```diff
@@ -0,0 +1,26 @@
+defmodule Absinthe.Phase.Telemetry do
+  @moduledoc """
+  Gather and report telemetry about a pipeline.
+  """
+  @telemetry_event [:absinthe, :query]
```
binaryseed: @benwilson512 @bruce 

Wanted to double-check with you on naming.. I went back and forth a little on this. This represents the `pipeline`, but there's also now a schema pipeline...

lib/absinthe/middleware/telemetry.ex
```diff
@@ -0,0 +1,41 @@
+defmodule Absinthe.Middleware.Telemetry do
+  @moduledoc """
+  Gather and report telemetry about an individual resolver function
+  """
+  @telemetry_event [:absinthe, :resolver]
```
benwilson512: This should probably be `[:absinthe, :resolved]` since that sounds more like an event.

lib/absinthe/phase/telemetry.ex
benwilson512: if we're going with event sounding names then probably `[:absinthe, :executed]`

lib/absinthe/phase/parse.ex
```diff
@@ -23,7 +23,12 @@ defmodule Absinthe.Phase.Parse do
   end
 
   def run(input, options) do
-    run(%Absinthe.Blueprint{input: input}, options)
+    telemetry = %{
+      start_time: System.system_time(),
+      start_time_mono: System.monotonic_time()
+    }
+
+    run(%Absinthe.Blueprint{input: input, source: input, telemetry: telemetry}, options)
```
benwilson512: Initializing all of this stuff is tricky. For example, how do we want to handle subscription documents? We probably want to re-initialize the telemetry information every time we execute the stored subscriptions. I'm inclined to say that this should get a dedicated phase just so we can make sure to loop it into the correct place.

Plus, with document providers, parsing may happen at compile time.

lib/absinthe/phase/telemetry.ex
binaryseed: Other examples:

* Plug: `[:my, :plug, :start]` / `[:my, :plug, :stop]`
* Ecto: `[:repo, :query]`
* Redix: `[:redix, :pipeline]` (redis pipelined commands)

benwilson512: Hm. Well I definitely want to avoid `:query` because in a graphql context that word has specific meaning. `[:absinthe, :exec]` seems to fit the existing styles maybe?

lib/absinthe/phase/parse.ex
binaryseed: I had a separate phase earlier, but I wasn't able to include parse time, since that's wired up in the current first phase.

I went down the path of splitting up the `Parse` phase to insert a `Telemetry` phase but that got messy real quick, lots of things depend on how the `Parse` phase works - do you think I should try that anyway?

The idea was to have the pipeline start with `[Init, Telemetry, Parse, ...`

lib/absinthe/phase/telemetry.ex
binaryseed: My question was more about trying to align the language to the exact details of what Absinthe is doing (execute pipeline) vs what users understand to be happening (query)

I just looked at the language in the GraphQL spec..  https://graphql.github.io/graphql-spec/draft/#sec-Executing-Requests  what about:

`[:absinthe, :execute, :request]`

lib/absinthe/middleware/telemetry.ex
binaryseed: After reviewing GraphQL spec, how about `[:absinthe, :resolve, :field]`

benwilson512: 👍

lib/absinthe/phase/telemetry.ex
benwilson512: Sorry, wasn't following. `[:absinthe, :execute, :operation]` perhaps? Request sounds too web focused to me.

lib/absinthe/phase/parse.ex
benwilson512: Could we parameterize the Telemetry phase?

```
    [
      {Phase.Telemetry, :start},
      # Parse Document
      {Phase.Parse, options},
      ...
      {Phase.Telemetry, :stop}
    ]
```

binaryseed: Yeah, I had a version of that earlier, but the tricky part is that `Phase.Parse` expects a `Language.Source.t()` & initializes a `Blueprint`, so before it there is nowhere to save the timestamp. I tried to split it but a lot of places (especially tests) expect `Phase.Parse` to take in a string

I could take on this refactor, I think i'd create a new first phase, something like `Phase.Init` which simply takes the `Language.Source` and inserts it into a new `Blueprint`, letting the `Phase.Parse` take it from there...

I'd probably try that as a separate PR then rebase this on top of that (just to keep things from getting out of hand here)

lib/absinthe/phase/telemetry.ex
binaryseed: 👍

lib/absinthe/phase/parse.ex
conradwt: @benwilson512 Is this something that one needs to opt-in or is this on by default?  Would it make sense to move this to another hex package within `absinthe-graphql` organization?

binaryseed: The goal of the `telemetry` project is to build instrumentation into the libraries themselves.. projects like Plug, Ecto & Phoenix are all building this into the core packages.

If there are no `telemetry` handlers installed, these will effectively be no-op calls, doing at most a fast handler lookup...

binaryseed:
I split the `Parse` phase (w/o touching it to avoid breaking hundreds of tests)

That let me insert the telemetry start phase early in the pipeline.

I also started working on getting subscriptions wired up with events via the `BatchResolver`

benwilson512:
@binaryseed Ah great, I wondered what the plan was for subscriptions. In theory it should be as simple as adding the new phases to here: https://github.com/absinthe-graphql/absinthe/blob/master/lib/absinthe/subscription/local.ex#L50

However the nature of the batch resolver makes this a bit complicated. Specifically, document execution is interleaved. AND fun fact: all of this can happen _within_ the resolution phase of a mutation document, since when a mutation field resolves it (by default) goes and lookups up subscription documents and executes them.

@binaryseed this does imply that there should probably be a dedicated `[:absinthe, :subscription, :publish, :start]` event and `[:absinthe, :subscription, :publish, :stop]` events.

garthk:
> Specifically, document execution is interleaved. 

Yeah, noticed. I'm having a blast trying to get [opencensus](https://hexdocs.pm/opencensus) to work well with Absinthe despite the `<a><b></a><c></b></c>` execution plan.

benwilson512:
@garthk yeah I wonder if the Absinthe.Execution struct needs some kind of `:erlang.unique_identifier` inside of it or something so that acts as an execution identifier. This could be spit out with each telemetry event so that it was clear which document an event was for. Would that help?

binaryseed:
todo:

* [ ] subscription specific publish events
* [ ] extractor functions

binaryseed:
At one point I put a unique ID in the `Blueprint`, but I scrapped it since it didn't get passed to the `Resolution` struct to connect a resolver to the containing query..

Also, it's pretty safe bet that at all points that we execute telemetry events, `self()` is the same, we only cross process boundaries inside the resolver functions (async, etc)

garthk:
The recent work makes my `opencensus_absinthe` hookup both easier and harder. Neither might be _relevant_, but I'll jot it down here if only to help anyone else thinking of using it to hook up tracing:

The lost call to `:telemetry.execute/3` at the beginning of field resolution to match `[:absinthe, :resolve, :field]` means I'll probably need to keep my middleware. Not so bad, especially given the metadata (see below).

The lost call at the beginning of the operation to match `[:absinthe, :execute, :operation]` at the end means I'll definitely need to keep my pipeline and phase code to call `:ocp.with_child_span/1` and `:ocp.finish_span/0`. I don't mind the code. I do mind the burden on the users to integrate pipelines. I already had a pipeline, for example, so I had to write my own `pipeline` target:

```elixir
  def custom_pipeline(config, pipeline_opts \\ []) do
    config
    |> Absinthe.Plug.default_pipeline(pipeline_opts)
    |> ApolloTracing.Pipeline.add_phases()
    |> Opencensus.Absinthe.Pipeline.add_phases()
  end
```

Arguably not so bad, again, but bigger than the middleware.

On the other hand, automatically inserting `:absinthe_telemetry` into the `field` metadata when someone uses `resolve` — am I understanding that right? — makes it _so much easier_ to filter out the noise Ben warned us about:

```elixir
  def middleware(middleware, field, object) do
    if Absinthe.Type.meta(field, :absinthe_telemetry) do
      [Opencensus.Absinthe.Middleware] ++ middleware
    else
      middleware
    end
  end
```

That spares me days of frustration wading through `Absinthe.Type.Field` and `Absinthe.Type.Object` looking for `Absinthe.Middleware.MapGet` and other clues. I couldn't be happier.

Zooming out, it's fine for telemetry and helps me with at least one of the hard things. When's the release?

binaryseed:
Nothing is lost, I'm open to adding back `start` events for tracing purposes - I was just under the impression that a span could be generated at the end with all the data, not that a `start_span` type function needed to be called inline when start happens.. Though it does make lots of sense for Distributed Tracing, which would need to know the current span ID in order to attach that info as part of an outgoing header

binaryseed:
>  uh, were you going to put any object or field information in the telemetry metadata?

I haven't put any additional fields on the metadata. They currently contain the `blueprint` / `resolution` which will enable you to extract out anything you might need.

Passing these structs enables the code listening to the telemetry events to decide what is important.

In my New Relic reporter, here are roughly the details I expect to pull out: 

#### Blueprint

```elixir
operation = Absinthe.Blueprint.current_operation(blueprint)

%{
  query: blueprint.source[:body] || blueprint.source,
  schema: blueprint.schema,
  variables: options[:variables],
  operation_complexity: operation.complexity,
  operation_type: operation.type,
  operation_name: operation.name
}
```


#### Resolution

```elixir
%{
  schema: res.schema,
  arguments: res.arguments,
  path: Absinthe.Resolution.path(res),
  field_name: res.definition.name,
  field_type: Absinthe.Type.name(res.definition.schema_node.type, res.schema),
  parent_type: res.parent_type.name
}
```

garthk:
Oh, cool. I thought you were ~finished. Anyhow, I've got the first POC for this telemetry-free tracing almost done so I can put it up and get design feedback.

binaryseed:
I'm now thinking my changes to `BatchResolver` for subscription publishing aren't quite right.

TODO: 
* [ ] Correct location to emit subscription publish events
* [ ] Test events emitted when doing an actual transport batch request

lib/absinthe/phase/subscription/subscribe_self.ex
```diff
@@ -28,7 +28,11 @@ defmodule Absinthe.Phase.Subscription.SubscribeSelf do
       for field_key <- field_keys,
           do: Absinthe.Subscription.subscribe(pubsub, field_key, subscription_id, blueprint)
 
-      {:replace, blueprint, [{Phase.Subscription.Result, topic: subscription_id}]}
+      {:replace, blueprint,
+       [
+         {Phase.Subscription.Result, topic: subscription_id},
+         {Phase.Telemetry, [:execute, :operation, options]}
```
benwilson512: Nice catch.

benwilson512:
🎉

PR #666, opened 2019-01-09
Allow capital camel case in LanguageConventions adapter

ctbucha:
The GraphQL client I'm using has a convention where some of the requested fields are capitalized camel case ("MyBackendResource"). Since it is difficult for me to change the client code convention for this, I created a new adapter that allows a capitalized field in absinthe like this:

```
query name: "Query" do
  field :My_backend_resource, :my_backend_resource do
  ... 
```

Then it seemed to me like this is probably acceptable behavior for the `LanguageConventions` adapter. Let me know if its okay to add this in or if you have any questions.

### DIFF
--- a/lib/absinthe/adapter/language_conventions.ex
+++ b/lib/absinthe/adapter/language_conventions.ex
@@ -59,6 +59,10 @@ defmodule Absinthe.Adapter.LanguageConventions do
     camelized_name
   end
 
+  def to_internal_name(<<c::utf8, rest::binary>>, _) when c in ?A..?Z do
+    <<c>> <> Macro.underscore(rest)
+  end
+
   def to_internal_name(camelized_name, _role) do
     camelized_name
     |> Macro.underscore()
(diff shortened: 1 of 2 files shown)


benwilson512:
Hey! I appreciate anyone who wants to contribute to Absinthe.

in this case though, instead of editing the built in adapter, you should write a custom adapter for your own purposes and specify that adapter when you configure plug.

ctbucha:
That's fine. I just saw that the inverse of this function exists here (https://github.com/absinthe-graphql/absinthe/blob/master/lib/absinthe/adapter/language_conventions.ex#L76) so I thought that it would make sense to include the one in this PR.

Specifically, `Absinthe.Adapter.LanguageConventions.to_external_name("My_backend_resource")` returns `MyBackendResource`, so it seemed like it would make sense to also have `Absinthe.Adapter.LanguageConventions.to_internal_name("MyBackendResource")` return `My_backend_resource` so that the two functions are the inverse of each other.

PR #676, opened 2019-01-26
Fix mix task to pass `pretty` flag to custom codecs, add tests

hadfieldn:
This fixes an issue preventing the `pretty` flag from being passed to custom codecs. It also adds tests to ensure that CLI options are parsed correctly, appropriate defaults are used, and custom codecs are called successfully.

Note that in order to make the tests perform an actual conversion to JSON, we would have to add Jason or Poison as a test-mode dependency. I opted against this because if we can ensure that the codec is accessed correctly, the risk of Poison's or Jason's `encode/2` method failing should be low.

However, if we don't add Jason or Poison as production-mode dependencies, we may want .to update the README to indicate that the application must load the the codec module  in order for the mix task to work, even if using the default Jason encoder.

### DIFF
--- a/lib/mix/tasks/absinthe.schema.json.ex
+++ b/lib/mix/tasks/absinthe.schema.json.ex
@@ -6,7 +6,6 @@ defmodule Mix.Tasks.Absinthe.Schema.Json do
   @shortdoc "Generate a schema.json file for an Absinthe schema"
 
   @default_filename "./schema.json"
-  @default_codec_name "Jason"
 
   @moduledoc """
   Generate a schema.json file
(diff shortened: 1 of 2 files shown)


hadfieldn:
Just some context behind this -- I need this fix in order to support pretty formatting in a custom codec I wrote that generates schemas with types sorted alphabetically. This results in cleaner diffs when committing schema files to version control. See https://github.com/hadfieldn/absinthe_sorting_codec

lib/mix/tasks/absinthe.schema.json.ex
```diff
@@ -100,4 +119,20 @@ defmodule Mix.Tasks.Absinthe.Schema.Json do
         [value] |> Module.safe_concat()
     end
   end
+
+  defp write_schema(content, filename) do
+    create_directory(Path.dirname(filename))
+    create_file(filename, content, force: true)
+  end
+
+  defp check_function_available(module, func) do
```
benwilson512: This basically just recreates the Elixir standard library error right?

```
iex(1)> Foo.encode!
** (UndefinedFunctionError) function Foo.encode!/0 is undefined (module Foo is not available)
    Foo.encode!()
```

I think we should just call encode! and if it doesn't work let the default error mesage do its thing.

lib/mix/tasks/absinthe.schema.json.ex
```diff
@@ -47,50 +47,69 @@ defmodule Mix.Tasks.Absinthe.Schema.Json do
...
-    %{module: module, opts: codec_opts(module, opts)}
+  @doc "Encode the Absinthe schema to a string, using the specified codec"
+  @spec generate_schema(Options.t()) :: String.t()
+  def generate_schema(%Options{
+        pretty: pretty,
+        schema: schema,
+        json_codec: json_codec
+      }) do
+    with {:ok, query} <- File.read(@introspection_graphql),
+         {:ok, result} <- Absinthe.run(query, schema),
```
benwilson512: This should probably use the newer https://hexdocs.pm/absinthe/Absinthe.Schema.html#introspect/2 function

lib/mix/tasks/absinthe.schema.json.ex
```diff
@@ -47,50 +47,69 @@ defmodule Mix.Tasks.Absinthe.Schema.Json do
...
+      json_codec: json_codec_as_atom(opts),
+      pretty: Keyword.get(opts, :pretty, false)
+    }
   end
 
-  defp codec_opts(_, _) do
-    []
+  defp json_codec_as_atom(opts) do
+    codec_name = Keyword.get(opts, :json_codec, @default_codec_name)
+    String.to_atom("Elixir." <> codec_name)
```
benwilson512:
```
opts
|> Keyword.fetch(:json_codec)
|> case do
  {:ok, codec} -> Module.concat([codec]}
  -> Jason
end
```

lib/mix/tasks/absinthe.schema.json.ex
hadfieldn: Yeah, that makes sense. I can't remember what led me down that path. ;-)

lib/mix/tasks/absinthe.schema.json.ex
hadfieldn: Oh, nice!

lib/mix/tasks/absinthe.schema.json.ex
hadfieldn: That's better, thanks!

hadfieldn:
Thanks for the helpful feedback @benwilson512. I implemented your suggestions and also cleaned up the docs a bit.

hadfieldn:
@benwilson512 Since this issue prevents the use of custom codecs in 1.4, would it be okay to create another PR that applies these changes to that branch?

benwilson512:
Yup!

benwilson512:
Thank you!

PR #677, opened 2019-01-27
(v1.4) Fix mix task to pass `pretty` flag to custom codecs, add tests

hadfieldn:
This applies to v1.4 the fix for custom codecs in https://github.com/absinthe-graphql/absinthe/pull/676. (The current release of v1.4 has a bug that prevents custom JSON codecs from loading -- it doesn't convert the module name to an atom.)


benwilson512:
thanks!

hadfieldn:
@benwilson512 Just wondering what the ETA is for publishing a new release of v1.4? It would be nice  to have this fix for a project I'm involved in, but my team is reluctant to pull a commit that isn't part of an official release.

benwilson512:
Done!

hadfieldn:
Wow, that was fast! Thanks so much!

Issue #678, opened 2019-01-28
Various typos, example of field options missing


fireproofsocks:
This issue was created as a target for an incoming PR that addresses a few typos in the docs and includes an example and cross-reference for the `name` option that is available to the `field` macro.  Inspired by this post: https://elixirforum.com/t/problematic-database-column-names/19734/6

maartenvanvliet:
This issue can be closed as the related PR was merged.

PR #679, opened 2019-01-28
Updates to documentation and examples.

fireproofsocks:
Adds an example of schema definition that uses additional options and provides a cross reference to the field macro.

This PR addresses issue #678 
**No code was changed**

I attempted to run all tests, but I am encountering this issue with my setup that I cannot figure out:

```
mix test
==> fs (compile)
Compiling c_src/mac/cli.c
c_src/mac/cli.c:1:10: fatal error: 'getopt.h' file not found
#include <getopt.h>
         ^~~~~~~~~~
1 error generated.
ERROR: compile failed while processing /Users/everettgriffiths/DropboxDM/absinthe/deps/fs: rebar_abort
** (Mix) Could not compile dependency :fs, "/path/to/.mix/rebar compile skip_deps=true deps_dir="/path/to/absinthe/_build/test/lib"" command failed. You can recompile this dependency with "mix deps.compile fs", update it with "mix deps.update fs" or clean it with "mix deps.clean fs"
```

So unfortunately, I cannot run the tests, so even though this PR only updates documentation and comments, I have to disclose that fact.

### DIFF
--- a/guides/schemas.md
+++ b/guides/schemas.md
@@ -58,14 +58,14 @@ object :item do
 end
 ```
 
-Now, you can use Absinthe to execute a query document. Keep in mind that for
+Now you can use Absinthe to execute a query document. Keep in mind that for
 HTTP, you'll probably want to use
 [Absinthe.Plug](plug-phoenix.html) instead of executing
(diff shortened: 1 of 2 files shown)


benwilson512:
Thank you!

PR #680, opened 2019-02-02
Add support for passing tasks to async middleware

maartenvanvliet:
The docs mentioned that you could use the bare plugin api but the module did not support it:

See https://hexdocs.pm/absinthe/Absinthe.Middleware.Async.html

  Using the bare plugin API
  ```elixir
  field :time_consuming, :thing do
    resolve fn _, _, _ ->
      task = Task.async(fn ->
        {:ok, long_time_consuming_function()}
      end)
      {:middleware, #{__MODULE__}, task}
    end
  end
  ```

This PR adds support for using the bare api.


benwilson512:
Thanks!

PR #686, opened 2019-02-16
Remove extra double quote from @if description.

potatosalad:
I had a few paragraphs written describing the changes in this pull request, but decided the title might also suffice.


benwilson512:
Thanks!

PR #693, opened 2019-03-02
Update Helpers.onload doc

### LINKED ISSUE #689: Resolution Helpers example may be innacurate
https://github.com/absinthe-graphql/absinthe/blob/fa207be01d8f2e4697ff9f0aaec320a5bec56562/lib/absinthe/resolution/helpers.ex#L91

Should this be using `Dataloader.get(…)` instead of `load`?

shamshirz:
fixes #689


binaryseed:
fixed in #697

PR #696, opened 2019-03-09
Update default for use_parent arg in dataloader/3 helper to match docs

bmuller:
The docs for the `dataloader/3` function say:
> `:use_parent` default: `true` This option affects whether or not ....

This is just changes the default value for the `:use_parent` option to actually be `true`.

Thanks for contributing such a great project!

### DIFF
--- a/lib/absinthe/resolution/helpers.ex
+++ b/lib/absinthe/resolution/helpers.ex
@@ -270,7 +270,7 @@ defmodule Absinthe.Resolution.Helpers do
     end
 
     defp use_parent(loader, source, resource, parent, args, opts) do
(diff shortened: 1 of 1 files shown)


bmuller:
Aaaand the tests fail for dataloader - so maybe this should actually just be a PR to change the docs to note the default is `false`.

tlvenn:
Hi @bmuller , this is actually a duplicate of #557 which is still outstanding as @benwilson512 wanted to take a look on how to address the issue with `nil` value in the KV scenario.

bmuller:
Ahh - thanks @tlvenn - somehow I missed this when looking through existing issues.

PR #707, opened 2019-04-11
Fix Absinthe.run/3 typespec

archseer:
Dialyzer would fail if the data response was a nested map, i.e. `%{"node" => %{"zone" => _zone}}`.


benwilson512:
Thanks!

PR #709, opened 2019-04-16
Upgrade nimble_parsec to 0.50

maartenvanvliet:
I upgraded nimble_parsec to 0.50 (see https://github.com/plataformatec/nimble_parsec/blob/master/CHANGELOG.md). 

One change was the deprecation of traverse/3 over post_traverse/3. I also made some changes to the lexer to make the tests pass again.

### DIFF
--- a/lib/absinthe/lexer.ex
+++ b/lib/absinthe/lexer.ex
@@ -92,12 +92,12 @@ defmodule Absinthe.Lexer do
       ]),
       times(ascii_char([?.]), 3)
     ])
-    |> traverse({:atom_token, []})
+    |> post_traverse({:atom_token, []})
 
   boolean_value_or_name_or_reserved_word =
     ascii_char([?_, ?A..?Z, ?a..?z])
     |> repeat(ascii_char([?_, ?0..?9, ?A..?Z, ?a..?z]))
-    |> traverse({:boolean_value_or_name_or_reserved_word, []})
+    |> post_traverse({:boolean_value_or_name_or_reserved_word, []})
 
   # NegativeSign :: -
   negative_sign = ascii_char([?-])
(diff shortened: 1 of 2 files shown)


lib/absinthe/lexer.ex
```diff
@@ -199,19 +199,17 @@ defmodule Absinthe.Lexer do
...
-    |> traverse({:string_value_token, []})
+    |> post_traverse({:string_value_token, []})
 
   block_string_value =
-    ignore(
-      string(~S("""))
-      |> traverse({:mark_block_string_start, []})
-    )
+    ignore(string(~S(""")))
+    |> post_traverse({:mark_block_string_start, []})
```
maartenvanvliet: Without this change of moving the `post_traverse` out of the `ignore` a "An unknown error occurred during parsing: key :token_location not found in: %{}" was seen in the tests. This was because `mark_block_string_start/5` was not called, and the compiler was complaining that it was not called. 

Changing this ensured that the `mark_block_string_start/5` was called.

lib/absinthe/lexer.ex
```diff
@@ -340,8 +338,8 @@ defmodule Absinthe.Lexer do
     {[chars], Map.put(context, :token_location, line_and_column(loc, byte_offset, 1))}
   end
 
-  defp mark_block_string_start(_rest, chars, context, loc, byte_offset) do
-    {[chars], Map.put(context, :token_location, line_and_column(loc, byte_offset, 3))}
+  defp mark_block_string_start(_rest, _chars, context, loc, byte_offset) do
```
maartenvanvliet: After changing the above two tests failed: "basic document, multiple lines with block string" and the ""document with emojis". This was because of usage of multiple lines with block string. 
It was because the first value (so `[[]]`) of the returned tuple of this function was a nested empty list instead of being an empty list.

bruce:
Heel erg bedankt Maarten! ✨

maartenvanvliet:
Haha geen probleem

PR #711, opened 2019-04-18
Customize Introspection

### LINKED ISSUE #710: Feature Request: Limited Visibility
I'm suggesting a way to limit the visibility of parts of the schema. The GraphQL Ruby gem has added this feature so I'll quote:

> With GraphQL-Ruby, it’s possible to hide parts of your schema from some users. This isn’t exactly part of the GraphQL spec, but it’s roughly within the bounds of the spec.
>
> Here are some reasons you might want to hide parts of your schema:
>
> You don’t want non-admin users to know about administration functions of the schema.
You’re developing a new feature and want to make a gradual release to only a few users first.

https://graphql-ruby.org/authorization/visibility

benwilson512:
Reminder: DO NOT SQUASH

Fixes #710

### DIFF
--- a/lib/absinthe/blueprint.ex
+++ b/lib/absinthe/blueprint.ex
@@ -139,4 +139,70 @@ defmodule Absinthe.Blueprint do
 
     %{blueprint | operations: ops}
   end
+
+  @doc """
+  Append the given field or fields to the given type
+  """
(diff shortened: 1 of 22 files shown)


benwilson512:
@michaelkschmidt This really is an absolutely heroic effort, thank you for your work here. Very pleased with what you did.

michaelkschmidt:
@benwilson512 Glad to help!  Let me know if there is anything you want tweaked

PR #714, opened 2019-04-19
Validate scalars on the way out

### LINKED ISSUE #603: Absinthe does not validate types on the way out

### Environment

* Elixir version (elixir -v):

```
Erlang/OTP 21 [erts-10.0.4] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe]

Elixir 1.7.1 (compiled with Erlang/OTP 21)
```

```
* absinthe 1.4.13 (Hex package) (mix)
  locked at 1.4.13 (absinthe) 81eb2ff4
* absinthe_metrics 0.9.0 (Hex package) (mix)
  locked at 0.9.0 (absinthe_metrics) 59b608f6
* absinthe_plug 1.4.5 (Hex package) (mix)
  locked at 1.4.5 (absinthe_plug) f63d52a7
* absinthe_phoenix 1.4.3 (Hex package) (mix)
  locked at 1.4.3 (absinthe_phoenix) cea34e7e
```

N/A.

### Expected behavior

If I define a field type as `:integer` and send out a string, I would expect to see a runtime error.

### Actual behavior

Absinthe sends a string even though this is not allowed by the GraphQL schema.

benwilson512:
Fixes #603 

- [x] Provide standard exception
- [ ] Provide unchecked scalars for backwards compatibility
- [x] ChangeLog note

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ For changes pre-v1.5 see the [v1.4](https://github.com/absinthe-graphql/absinthe
 - Feature: SDL directives, other improvements
 - Feature: Output rendered SDL for a schema
 - Documentation: Testing guide, numerous fixes and updates
+- Breaking Change: Scalar outputs are now type checked and will raise exceptions
+if the result tries to send the wrong data type in the result.
 
 ## v1.5.0 (Alpha)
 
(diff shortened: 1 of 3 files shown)


benwilson512:
Instead of having dedicated unchecked scalars, the changes in 1.5 to schema construction should allow us to actually just swap out the resolvers to the built in scalars. We may want to support some kind of first class option like `use Absinthe.Schema, unchecked_scalars: [:integer, :float]`. Alternatively, if the API for editing types was sufficiently could we could just document how to do that, which seems more generally useful.

PR #716, opened 2019-04-21
Use dialyxir over unmaintained dialyze

maartenvanvliet:
Dialyze is no longer maintained see https://github.com/fishcakez/dialyze#mixtaskdialyze
Dialyxer is the most active project to use Dialyzer so I suggest it is
used to replace Dialyze

A run of `mix dialyzer --format dialyxir` shows several errors.


benwilson512:
Thanks!

PR #717, opened 2019-04-21
Absinthe.Subscription.Local.publish_mutation/3 doc

### LINKED ISSUE #706: Proposal: Node local version of Absinthe.Subscription.publish/3
### Background

The app I'm working on is proxying updates from an external source and exposing these updates as GraphQL subscription fields.

Each Elixir node in our cluster has a process monitoring the external source and then publishes updates to the related GraphQL field.  Since each Elixir node is independently monitoring the external source, we only need to publish subscription updates to the local node.

### Proposal

To "officially" support this use case, I was hoping to either:

- add a public `Absinthe.Subscription.publish_local/3` function, or
- add docs to `Absinthe.Subscription.Local.publish_mutation/3` function

Thoughts?  I'm happy to make a PR with the preferred approach (or another approach)

darrenclark:
Fixes #706

### DIFF
--- a/lib/absinthe/subscription/local.ex
+++ b/lib/absinthe/subscription/local.ex
@@ -1,5 +1,7 @@
 defmodule Absinthe.Subscription.Local do
-  @moduledoc false
+  @moduledoc """
+  This module handles broadcasting documents that are local to this node
+  """
 
   require Logger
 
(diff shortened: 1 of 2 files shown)


benwilson512:
Thank you!

PR #718, opened 2019-04-22
Fix several dialyzer errors

maartenvanvliet:
This PR fixes several dialyzer errors. There are quite a few more but these had the most obvious fixes.


benwilson512:
Thanks you!

PR #719, opened 2019-04-24
Add Absinthe.Adapter.StrictLanguageConventions

potatosalad:
Alternative adapter for enforcing strict incoming document directive, field, and argument naming to be explicitly camelCase.

##### What needs to be solved?

The default [`Absinthe.Adapter.LanguageConventions`](https://github.com/absinthe-graphql/absinthe/blob/master/lib/absinthe/adapter/language_conventions.ex) allows non-camelcase external names to be converted to their respective internal names.

A stricter version of the language conventions adapter would be preferable where the "external" in an "external-to-internal" conversion matches the "external" in an "internal-to-external" conversion.

It is also preferable that any changes to absinthe cause as little disruption as possible to existing users who are not concerned about this implicit flexibility.

##### Why?

@ferigis Initially brought this unintentional feature-or-bug to my attention today where incoming documents by default can use a variety of invalid directive, field, and/or argument names and the server will respond normally without any errors.

For example, for a given field:

```elixir
field :expires_at, :datetime
```

The default [`Absinthe.Adapter.LanguageConventions`](https://github.com/absinthe-graphql/absinthe/blob/master/lib/absinthe/adapter/language_conventions.ex) accepts any inbound name like: `expiresAt`, `expires_at`, `EXPIRES_AT`, or `ExpiresAt` and maps them to the internal `expires_at`.

Clients, like [GraphQL Playground built into `absinthe_plug`](https://github.com/absinthe-graphql/absinthe_plug), properly warn that these are invalid as they do not match the field names exposed in the introspection query:

Playground - 2019-04-24 14-41-53

The server, however, responds to the request as if it is a valid query:

Playground - 2019-04-24 14-42-47

I am of the opinion that this is an error or bug on the server's part.

In my opinion, the server should not respond to a directive, field, or argument that has not been returned as part of the introspection query (with the exception of [two underscore `__` reserved names](https://graphql.github.io/graphql-spec/draft/#sec-Reserved-Names)).

At the very least, I think there should be an optional way to enforce this behavior server-side even if it is not the default.

##### How?

I reached out on [#absinthe-graphql](https://elixir-lang.slack.com/messages/C0PR49P4P/) to provide a brief description of the behavior and to find out whether this was intentional or well-known at this point:

absinthe-graphql | Elixir Slack 2019-04-24 11-44-40

I initially began by changing [`Absinthe.Adapter.LanguageConventions`](https://github.com/absinthe-graphql/absinthe/blob/master/lib/absinthe/adapter/language_conventions.ex) directly, but found that there are several tests (like [`test/absinthe/custom_types_test.exs`](https://github.com/absinthe-graphql/absinthe/blob/master/test/absinthe/custom_types_test.exs)) that actually make use of the underscored version of field names.

Unless we're okay with enforcing this behavior by default (which may be a breaking change for some users of absinthe), I think the simplest route may be to just provide an alternative adapter that allows the server to operate in strict mode.

Summary of changes:

* Added [`Absinthe.Adapter.StrictLanguageConventions`](https://github.com/potatosalad/absinthe/blob/strict-language-conventions/lib/absinthe/adapter/strict_language_conventions.ex) which first converts the name to the internal (underscored) representation.  It then takes this internal representation and converts it back to the external representation.  If the external representation does not match the originally supplied name, then `nil` is returned.
* Modified [`Absinthe.Phase.Document.Validation.FieldsOnCorrectType`](https://github.com/potatosalad/absinthe/blob/strict-language-conventions/lib/absinthe/phase/document/validation/fields_on_correct_type.ex) to handle the `nil` returned by `to_internal_name/2` so that suggested field names can still be returned on mismatch.
* Added tests to [`test/absinthe/adapters/language_conventions_test.exs`](https://github.com/potatosalad/absinthe/blob/strict-language-conventions/test/absinthe/adapters/language_conventions_test.exs) to demonstrate the existing behavior of the default adapter.
* Added tests to [`test/absinthe/adapters/strict_language_conventions_test.exs`](https://github.com/potatosalad/absinthe/blob/strict-language-conventions/test/absinthe/adapters/strict_language_conventions_test.exs) to demonstrate the alternative behavior of the strict adapter.
* Added tests to [`test/absinthe/strict_schema_test.exs`](https://github.com/potatosalad/absinthe/blob/strict-language-conventions/test/absinthe/strict_schema_test.exs) to demonstrate the effect of the different adapters for directives, queries, and mutations.
* Added some basic documentation describing the adapter.

### DIFF
--- a/lib/absinthe/adapter/language_conventions.ex
+++ b/lib/absinthe/adapter/language_conventions.ex
@@ -47,6 +47,7 @@ defmodule Absinthe.Adapter.LanguageConventions do
   """
 
   @doc "Converts a camelCase to snake_case"
+  @impl Absinthe.Adapter
   def to_internal_name(nil, _role) do
     nil
   end
@@ -55,16 +56,17 @@ defmodule Absinthe.Adapter.LanguageConventions do
     "__" <> to_internal_name(camelized_name, role)
   end
 
-  def to_internal_name(camelized_name, :operation) do
+  def to_internal_name(camelized_name, :operation) when is_binary(camelized_name) do
     camelized_name
   end
 
-  def to_internal_name(camelized_name, _role) do
+  def to_internal_name(camelized_name, _role) when is_binary(camelized_name) do
     camelized_name
     |> Macro.underscore()
   end
 
   @doc "Converts a snake_case name to camelCase"
+  @impl Absinthe.Adapter
   def to_external_name(nil, _role) do
     nil
   end
@@ -73,11 +75,15 @@ defmodule Absinthe.Adapter.LanguageConventions do
     "__" <> to_external_name(underscored_name, role)
   end
 
+  def to_external_name(underscored_name, :operation) when is_binary(underscored_name) do
+    underscored_name
+  end
+
   def to_external_name(<<c::utf8, _::binary>> = name, _) when c in ?A..?Z do
     name |> Utils.camelize()
   end
 
-  def to_external_name(underscored_name, _role) do
+  def to_external_name(underscored_name, _role) when is_binary(underscored_name) do
     underscored_name
     |> Utils.camelize(lower: true)
   end
--- a/lib/absinthe/phase/document/validation/fields_on_correct_type.ex
+++ b/lib/absinthe/phase/document/validation/fields_on_correct_type.ex
@@ -180,7 +180,12 @@ defmodule Absinthe.Phase.Document.Validation.FieldsOnCorrectType do
   end
 
   defp suggested_type_names(external_field_name, type, blueprint) do
-    internal_field_name = blueprint.adapter.to_internal_name(external_field_name, :field)
+    internal_field_name =
+      case blueprint.adapter.to_internal_name(external_field_name, :field) do
+        nil -> external_field_name
+        internal_field_name -> internal_field_name
+      end
+
     possible_types = find_possible_types(internal_field_name, type, blueprint.schema)
 
     possible_interfaces =
@@ -193,7 +198,11 @@ defmodule Absinthe.Phase.Document.Validation.FieldsOnCorrectType do
   end
 
   defp suggested_field_names(external_field_name, %{fields: _} = type, blueprint) do
-    internal_field_name = blueprint.adapter.to_internal_name(external_field_name, :field)
+    internal_field_name =
+      case blueprint.adapter.to_internal_name(external_field_name, :field) do
+        nil -> external_field_name
+        internal_field_name -> internal_field_name
+      end
 
     Map.values(type.fields)
     |> Enum.map(& &1.name)
--- a/test/absinthe/adapters/language_conventions_test.exs
+++ b/test/absinthe/adapters/language_conventions_test.exs
@@ -11,6 +11,16 @@ defmodule Absinthe.Adapter.LanguageConventionsTest do
     test "converts external camelcase variable names to underscore" do
       assert "foo_bar" = LanguageConventions.to_internal_name("fooBar", :variable)
     end
+
+    test "preserves external operation names" do
+      assert "fooBar" = LanguageConventions.to_internal_name("fooBar", :operation)
+    end
+
+    test "converts external field names that do not match internal name" do
+      assert "foo_bar" = LanguageConventions.to_internal_name("foo_bar", :field)
+      assert "foo_bar" = LanguageConventions.to_internal_name("FooBar", :field)
+      assert "foo_bar" = LanguageConventions.to_internal_name("FOO_BAR", :field)
+    end
   end
 
   describe "to_external_name/2" do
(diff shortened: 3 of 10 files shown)


binaryseed:
This looks like a good change to make, and I like that it is opt-in since existing APIs likely rely on the current behavior.

We are nearing up on the 1.5 release, could you perhaps rebase on top of the latest master so we can give a final review?

potatosalad:
@binaryseed I force pushed these changes based on the latest master a few minutes ago.

Fixed one of the tests that was failing due to better error messages from master (specifically, the error message `"Unknown directive."` now says <tt>"Unknown directive `foo_bar_directive'."</tt>).

test/absinthe/adapters/language_conventions_test.exs
```diff
@@ -11,6 +11,16 @@ defmodule Absinthe.Adapter.LanguageConventionsTest do
     test "converts external camelcase variable names to underscore" do
       assert "foo_bar" = LanguageConventions.to_internal_name("fooBar", :variable)
     end
+
+    test "preserves external operation names" do
+      assert "fooBar" = LanguageConventions.to_internal_name("fooBar", :operation)
```
binaryseed: Does this imply a change in behavior? Seems like previously an operation name would be transformed?

I imagine this isn't a breaking change in the usual sense, but I suspect some APIs might track the operation name for monitoring, etc and could be relying on the existing behavior - say you have a dashboard tracking API usage based on operation name...

lib/absinthe/phase/document/validation/fields_on_correct_type.ex
```diff
@@ -193,7 +198,11 @@ defmodule Absinthe.Phase.Document.Validation.FieldsOnCorrectType do
   end
 
   defp suggested_field_names(external_field_name, %{fields: _} = type, blueprint) do
-    internal_field_name = blueprint.adapter.to_internal_name(external_field_name, :field)
+    internal_field_name =
+      case blueprint.adapter.to_internal_name(external_field_name, :field) do
+        nil -> external_field_name
```
binaryseed: Where do these `suggested_*_names` changes come into play?

test/absinthe/adapters/language_conventions_test.exs
binaryseed: moving this comment to the relevant function

lib/absinthe/adapter/language_conventions.ex
```diff
@@ -73,11 +75,15 @@ defmodule Absinthe.Adapter.LanguageConventions do
     "__" <> to_external_name(underscored_name, role)
   end
 
+  def to_external_name(underscored_name, :operation) when is_binary(underscored_name) do
```
binaryseed: Does this imply a change in behavior? Seems like previously an operation name would be transformed?

I imagine this isn't a breaking change in the usual sense, but I suspect some APIs might track the operation name for monitoring, etc and could be relying on the existing behavior - say you have a dashboard tracking API usage based on operation name...

potatosalad: It's been several months since I made these changes, but I'll try to remember...

I think the previous version was unnecessarily performing `Utils.camelize(name, lower: true)` on things like `query`, `mutation`, and `subscription`.

This change was made to match the `to_internal_name` behavior above where the `:operation` names are left untouched.  With this change in place `to_internal_name` and `to_external_name` are roughly roundtrip versions of each other, which is how the strict version of the adapter operates (by checking that they do in fact match).  Only those using custom operations would be affected and I'm not even sure whether this was even working correctly in the first place.

lib/absinthe/phase/document/validation/fields_on_correct_type.ex
potatosalad: The strict adapter returns `nil` if there a malformed name is seen.  Things down the line that depend on this value to be a binary break if `nil` is set to `internal_field_name` here.

The default adapter smooths over whatever mess `external_field_name` comes in looking like and it maps to a nice version of the internal name (so `FOO_BAR` becomes `foo_bar`).

In the strict adapter case, it makes more sense to return the original input (`FOO_BAR`) so it won't match correctly down the line and suggestions can be sent back to the client.

lib/absinthe/adapter/language_conventions.ex
binaryseed: So I looked around and couldn't find a use of `to_external_name` or `to_internal_name` using `:operation`... Also looked in `absinthe_plug`

I even put a `raise` in both functions and all the tests pass just fine.

I suspect we could just remove `:operation` from the `@type role_t` and remove the function heads. I looked back and `to_internal_name` used to be called in the inner workings of `Adapter` but that doesn't happen anymore

cc @bruce

lib/absinthe/phase/document/validation/fields_on_correct_type.ex
binaryseed: sounds good

lib/absinthe/adapter/language_conventions.ex
bruce: I think that's fine.

bruce:
Provided this is opt-in, I'm very 👍.

lib/absinthe/adapter/language_conventions.ex
binaryseed: I'll follow up with a quick PR to remove those

PR #727, opened 2019-05-06
Fail fast in formatting check

binaryseed:
@benwilson512 I think it's helpful to fail fast if the formatting is incorrect, instead of waiting for the full test suite to run only to fail.


benwilson512:
good point!

PR #729, opened 2019-05-09
Re-add Schema callbacks with docs

maartenvanvliet:
The callbacks for the Absinthe.Schema were removed in this commit https://github.com/absinthe-graphql/absinthe/commit/01be95b540c3a2a4419fd2e8b117ef1f9eb8a569#diff-a12a7aea6a6a7790c1826d981d259707L242

I think they should be re-added as they are still called on a schema. I've also expanded the docs a little and added the callback for `decorations/2`


bruce:
@maartenvanvliet Thanks for this! I've expanded on it a bit, and changed references to "decorations" to the new "hydrate" terminology we're using after #731.

PR #732, opened 2019-05-20
Phase.Init: pass blueprints through

archseer:
The telemetry branch added a new phase which runs before Phase.Parse: https://github.com/absinthe-graphql/absinthe/commit/8ff081647bff240c80b259a6032204d1c652c1a1#diff-bf69d2d0e622a30890fc6e89dc61c094R1

The problem is, while Phase.Parse allowed passing in an existing blueprint, Phase.Init currently doesn't account for that, and essentially returns `%Blueprint{input: blueprint}`. This is a problem for us, because the invalid state would later fail saying "No operations provided.". 

This PR restores the old behaviour.

### DIFF
--- a/lib/absinthe/phase/init.ex
+++ b/lib/absinthe/phase/init.ex
@@ -5,8 +5,14 @@ defmodule Absinthe.Phase.Init do
 
   alias Absinthe.{Blueprint, Language, Phase}
 
-  @spec run(Language.Source.t(), Keyword.t()) :: Phase.result_t()
-  def run(input, _options \\ []) do
+  @spec run(Blueprint.t() | Language.Source.t(), Keyword.t()) :: Phase.result_t()
+  def run(blueprint, options \\ [])
+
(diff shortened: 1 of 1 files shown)


benwilson512:
Hey @archseer thanks for catching this! Can you include a test case to make sure that future changes don't cause the same regression?

binaryseed:
Thanks. `Phase.Parse` still behaves the same way.. how are you interacting with the phases to send in a blueprint? That's probably what we should test..

archseer:
We manually constructed a blueprint that we passed into Absinthe.run with a single operation to run some sort of introspection in our error middleware (I didn't write the module so I don't understand it fully).

The problem is, while the old pipeline allowed either a blueprint type or a language source type, the new init phase runs first and only accepts a language source. Before, Phase.Parse was first and allowed passing in either.

binaryseed:
Makes sense, thanks!

binaryseed:
I'll add a test

PR #738, opened 2019-05-27
Fix #650 (@desc not working on field arguments)

### LINKED ISSUE #650: @desc does not work for field arguments
### Environment

* Elixir version (elixir -v): 1.7.3
* Absinthe version (mix deps | grep absinthe): 1.5.0-alpha.1

### Expected behavior

```
query do
  field :my_operation do
    @desc "My argument description"
    arg :my_arg, :integer
  end
end
```

I expect the `@desc` call to add a `:description` to the argument.

### Actual behavior

The `@desc` above an argument does not work, it appears to be completely ignored.

darrenclark:
- Fix #650 (`@desc` not working on field arguments)
- Also fixes the `:name` option not working for arguments

### Approach I took

Since `Absinthe.Blueprint.Schema.build_types/3` doesn't push a `InputValueDefinition` onto the stack (it sets is on the `FieldDefinition`), a `{:desc, _}` tuple doesn't get applied to the `Schema.InputValueDefinition`.

I added a clause to `build_types/3` that peeks ahead of a `InputValueDefinition` to consume the `{:desc, _}` tuple

### Other thoughts

Looking back, would it make more sense to instead to:

- Have `Absinthe.Schema.Notation` create a `:close` element after `Schema.InputValueDefinition` (and any `{:desc, _}` for that argument), and,
- put the `Schema.InputValueDefinition` on the stack in `build_types/3`, then wait for the `:close` to set it on the `Schema.FieldDefinition`, and
- let the existing `build_types/3` clause for `{:desc, _}` handle setting `description` on the `Schema.InputValueDefinition`

Thoughts?

### DIFF
--- a/lib/absinthe/blueprint/schema.ex
+++ b/lib/absinthe/blueprint/schema.ex
@@ -94,6 +94,7 @@ defmodule Absinthe.Blueprint.Schema do
     Schema.EnumTypeDefinition,
     Schema.DirectiveDefinition,
     Schema.InputObjectTypeDefinition,
+    Schema.InputValueDefinition,
     Schema.InterfaceTypeDefinition,
     Schema.UnionTypeDefinition,
     Schema.EnumValueDefinition
@@ -141,16 +142,6 @@ defmodule Absinthe.Blueprint.Schema do
     build_types(rest, [enum | stack], buff)
   end
 
-  defp build_types([%Schema.InputValueDefinition{} = arg, {:desc, desc} | rest], stack, buff) do
-    arg = %{arg | description: desc}
-    build_types([arg | rest], stack, buff)
-  end
-
-  defp build_types([%Schema.InputValueDefinition{} = arg | rest], [field | stack], buff) do
-    arg = Map.update!(arg, :default_value, fn val -> {:unquote, [], [val]} end)
-    build_types(rest, [push(field, :arguments, arg) | stack], buff)
-  end
-
   defp build_types([{:sdl, sdl_definitions} | rest], [schema | stack], buff) do
     # TODO: Handle directives, etc
     build_types(rest, [concat(schema, :type_definitions, sdl_definitions) | stack], buff)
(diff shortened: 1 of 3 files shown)


lib/absinthe/schema/notation.ex
```diff
@@ -1201,15 +1201,19 @@ defmodule Absinthe.Schema.Notation do
...
   end
 
   def record_arg!(env, identifier, attrs) do
     arg = build_arg(identifier, Keyword.put(attrs, :module, env.module), env)
-    put_attr(env.module, arg)
+    ref = put_attr(env.module, arg)
+
+    [
+      get_desc(ref)
+    ]
```
benwilson512: Ah yeah, if we weren't adding the `get_desc` call here, then it wouldn't have ended up on the stack at all.

lib/absinthe/schema/notation.ex
```diff
@@ -1201,15 +1201,19 @@ defmodule Absinthe.Schema.Notation do
       attrs
       |> handle_deprecate
       |> Keyword.put(:identifier, identifier)
-      |> Keyword.put(:name, to_string(identifier))
+      |> Keyword.put_new(:name, to_string(identifier))
```
benwilson512: Good catch!

lib/absinthe/blueprint/schema.ex
```diff
@@ -141,6 +141,11 @@ defmodule Absinthe.Blueprint.Schema do
     build_types(rest, [enum | stack], buff)
   end
 
+  defp build_types([%Schema.InputValueDefinition{} = arg, {:desc, desc} | rest], stack, buff) do
```
benwilson512: What's weird is that this shouldn't be necessary because the https://github.com/absinthe-graphql/absinthe/blob/ad2e4c2508379c323ed37b1556e224865f350fba/lib/absinthe/blueprint/schema.ex#L110 clause should already handle this. Doing a lookahead to find the desc should be exactly identical to just moving one more item through the stack, matching on desc, and modifying the previous thing.

However if I remove this clause locally, the tests no longer pass. Investigating.

benwilson512: Ah of course, this is what you were saying, I see it now. https://github.com/absinthe-graphql/absinthe/blob/ad2e4c2508379c323ed37b1556e224865f350fba/lib/absinthe/blueprint/schema.ex#L149 eagerly removes the InputValue definition from the stack, so further items can't do anything to it.

darrenclark:
Any thoughts on if I should go as far as adding it to `@scoped_defs` and having it handled via `record!`?

Or just update `record_arg!` to emit a `:close` as well?

https://github.com/absinthe-graphql/absinthe/blob/7f7cb9ef545a6319f55b88649d3952bd0aca7b39/lib/absinthe/schema/notation.ex#L1182-L1197

benwilson512:
@darrenclark go ahead and add it to scoped defs.

darrenclark:
@benwilson512 updated

benwilson512:
Thank you!!!

FloatingFront:
Likewise thanks!

PR #741, opened 2019-06-02
SDL Directive Argument Support

bruce:
This PR adds support for directive arguments used in SDL to define a schema (eg, `@deprecated(reason: "...")`). Unlike the rest of the schema definition pipeline, this requires mapping blueprint nodes to built-in types (setting `schema_node`, as is done when processing documents) so that values can be "cast" appropriately.


bruce:
- [ ] Deep assignment of schema nodes for blueprint input values (this is done in `Absinthe.Phase.Schema.Arguments.Directives`)

PR #736, opened 2019-05-24
Render SDL

binaryseed:
This PR adds the ability to render out `Blueprint` structs as GraphQL SDL.

If you have a blueprint struct, you may view it's SDL by `inspect`ing it with the `:pretty` flag:

```elixir
IO.inspect(blueprint, pretty: true)
```

From the command line, you can invoke a Mix Task generate a `.graphql` file:

```bash
mix absinthe.schema.sdl --schema MySchema
```

PR #743, opened 2019-06-03
Add directive locations for schema

binaryseed:
* Updating directive locations for schemas..

As noted in https://github.com/absinthe-graphql/absinthe/pull/736/files/d3d9dcde3bc6623ff839148fbe21df7f6bef2b6b#r289664556

* Update built-in `@deprecated` to allow it on enum values

### DIFF
--- a/lib/absinthe/type/built_ins/introspection.ex
+++ b/lib/absinthe/type/built_ins/introspection.ex
@@ -77,7 +77,6 @@ defmodule Absinthe.Type.BuiltIns.Introspection do
 
   enum :__directive_location,
     values: [
-      # OPERATIONS
       :query,
       :mutation,
       :subscription,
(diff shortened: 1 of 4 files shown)


lib/absinthe/type/built_ins/introspection.ex
```diff
@@ -84,8 +84,19 @@ defmodule Absinthe.Type.BuiltIns.Introspection do
       :field,
       :fragment_definition,
       :fragment_spread,
-      :inline_fragment
-      # TODO: Schema definitions to support Schema input
+      :inline_fragment,
+      :operation_definition,
```
binaryseed: Is this correct? I don't think it is...

directive locations are defined here: https://graphql.github.io/graphql-spec/draft/#sec-Schema-Introspection

bruce: Agreed. I removed this in a commit, pushed.

PR #750, opened 2019-06-14
Add testing guide to docs

maartenvanvliet:
I felt that the guides missed a page detailing how to test Absinthe. I've added a page with a simple test for testing the entire pipeline with Plug and Absinthe, as this would be the most common combination. 

Consider this an initial write-up, e.g. it could be expanded with a guide on dealing with authentication with tests.


bruce:
Thanks, @maartenvanvliet! 💐

PR #761, opened 2019-07-23
fix: Interface and union on project

fahchen:
This PR fix: if the type of a query is an interface or union, the project function would ignore selections under the interface or union.

### DIFF
--- a/test/absinthe/resolution_test.exs
+++ b/test/absinthe/resolution_test.exs
@@ -16,8 +16,24 @@ defmodule Absinthe.ResolutionTest do
       field :name, :string
     end
 
+    object :member do
+      interface :named
+      field :mid, :id
+      field :name, :string
+    end
+
     query do
       field :user, :user do
+        resolve fn _, info ->
+          fields = Absinthe.Resolution.project(info) |> Enum.map(& &1.name)
+
+          # ghetto escape hatch
+          send(self(), {:fields, fields})
+          {:ok, nil}
+        end
+      end
+
+      field :named, :named do
         resolve fn _, info ->
           fields = Absinthe.Resolution.project(info) |> Enum.map(& &1.name)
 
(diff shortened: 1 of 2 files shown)


fahchen:
```elixir
# schema
defmodule Schema do
  use Absinthe.Schema

  interface :named do
    field :name, :string

    resolve_type fn _, _ -> :user end
  end

  object :user do
    interface :named
    field :id, :id
    field :name, :string
  end

  object :member do
    interface :named
    field :mid, :id
    field :name, :string
  end

  query do
    field :named, :named do
      resolve fn _, info ->
        fields = Absinthe.Resolution.project(info) |> Enum.map(& &1.name)

        # ghetto escape hatch
        send(self(), {:fields, fields})
        {:ok, nil}
      end
    end
  end
end

# doc
doc = """
{
  named {
    ... on User {
      id
    }

    name

    ... on Member {
      mid
    }
  }
}
"""
```

In this test https://github.com/absinthe-graphql/absinthe/pull/761/files#diff-e9d2f1e11376ab3aa5772e2f22a98e5eR125, `named` is an interface, and it is the type of `named` query. And `user` and `member` are both the interface of the `named`.
In doc above, the parent type of the resolution is `named`, which is neither `user` nor `member`

test/absinthe/resolution_test.exs
```diff
@@ -66,6 +101,54 @@ defmodule Absinthe.ResolutionTest do
...
+        }
+      }
+    }
+    """
+
+    {:ok, _} = Absinthe.run(doc, Schema)
+
+    assert_receive({:fields, fields})
+
+    assert ["id", "name", "mid"] == fields
```
benwilson512: An abstract field resolves to exactly _one_ concrete type for any given resolver invocation. There is no concrete type where `["id", "name", "mid"]` are all valid fields.

fahchen: The type of `named` query is `named` that is an interface, and `user` and `member` implement it, that’s why returning with `["id", "name", "mid"]` that isn’t a concrete type.

benwilson512: The point of project however is to return the concrete fields to execute, not all fields that may be possibly executed by some type. `project/1` should honestly raise when provided an abstract type from the parent and no concrete type name to project on to.

benwilson512: The function you want to use is this: https://hexdocs.pm/absinthe/Absinthe.Resolution.html#project/2

`Resolution.project(info, :user)` or `Resolution.project(info, :member)`. Then you'll get the fields for each of those types. That's what I mean about `Resolution.project(info)` basically being invalid for abstract types.

fahchen: Thanks, I got it

Issue #737, opened 2019-05-26
AST values in meta attributes


stavro:
Elixir version: 1.8.1

On Absinthe v1.4.13, meta attributes can contain anonymous functions:

```elixir
field :foo, :foo do
  meta :key, &(&1)
end
```

And when retrieving the meta value during runtime, the anonymous function works as expected.

On the latest `master` (https://github.com/absinthe-graphql/absinthe/commit/7f7cb9ef545a6319f55b88649d3952bd0aca7b39 at this time), the value obtained during runtime is the AST of the function (`({:&, [line: 93], [{:&, [line: 93], [1]}]}`) instead of the function.

Our app currently walks the fields of the query and gathers some information via meta tags prior to execution, and the use of anonymous functions here has been very valuable.

binaryseed:
What is the expected behavior of having an anonymous function in this location?

stavro:
The expected behavior would be when calling `Absinthe.Type.meta(schema_node, :key)` during runtime you would receive an anonymous function instead of the AST.

benwilson512:
Hey @stavro, you should be able to solve this by extracting the function to a named function and doing `meta :key, &SomeModule.some_function/1`. While Absinthe supports inline anonymous resolver functions, doing so has come at an enormous complexity cost, and we don't plan to continue supporting inline anonymous functions for other properties.

PR #771, opened 2019-08-19
Enable hydrating meta

binaryseed:
For an SDL defined schema, this PR enables setting the `meta` information like you can with a macro based schema


benwilson512:
Looks correct, and per #737 I think we can simply not worry about non escapable values. It's impractical to handle them for every possible value. Needs tests though.

binaryseed:
Thanks, pushed a test..

PR #782, opened 2019-09-14
Use Telemetry start/stop convention

josevalim:
This is the same convention used in Phoenix.

/cc @binaryseed @benwilson512


binaryseed:
Thanks!

binaryseed:
I added a note in the changelog

PR #785, opened 2019-09-15
Fix telemetry event name in guide

stavro:
Small documentation duplication in the telemetry changes just merged

cc @binaryseed


binaryseed:
thanks!

PR #786, opened 2019-09-16
Input union

binaryseed:


binaryseed:
Part of an old experiment I'm looking to revive

tlvenn:
That would be awesome, union of input type is dearly missed and I believe there is an ongoing effort to standardize it.

binaryseed:
Yep, that's what I'm working on!

https://github.com/graphql/graphql-spec/blob/master/rfcs/InputUnion.md

PR #789, opened 2019-09-23
Fix meta(absinthe_telemetry: true/false)

darrenclark:
### Changes

- Defining a resolver now sets `:absinthe_telemetry_implicit` in a field's meta so that a user-provided `:absinthe_telemetry` value is not overridden / doesn't conflict
- Additionally, setting `absinthe_telemetry: true` on a field without a resolver now adds the telemetry middleware.

---

### Issue

Ran into an exception when setting `meta(abinsthe_telemetry: false)` on a field that defined a resolver.  i.e.

```elixir
field :example, :string do
  meta(abinsthe_telemetry: false)
  resolve(fn _, _, _ -> {:ok, "hello world"} end)
end
```

would raise an error when compiling:

```
== Compilation error in file test/absinthe/schema_test.exs ==
** (FunctionClauseError) no function clause matching in Keyword.merge/3

    The following arguments were given to Keyword.merge/3:

        # 1
        false

        # 2
        true

        # 3
        #Function<12.43655748/3 in Absinthe.Blueprint.Schema.update_private/2>

    Attempted function clauses (showing 1 out of 1):

        def merge(keywords1, keywords2, fun) when is_list(keywords1) and is_list(keywords2) and is_function(fun, 3)

    (elixir) lib/keyword.ex:763: Keyword.merge/3
    (elixir) lib/keyword.ex:776: Keyword.do_merge/6
    (elixir) lib/keyword.ex:776: Keyword.do_merge/6
    (elixir) lib/map.ex:752: Map.update!/3
    (absinthe) lib/absinthe/blueprint/schema.ex:143: Absinthe.Blueprint.Schema.build_types/3
    (absinthe) expanding macro: Absinthe.Schema.Notation.__before_compile__/1
    test/absinthe/schema_test.exs:436: Absinthe.SchemaTest.TelemetrySchema (module)
    (elixir) lib/code.ex:767: Code.require_file/2
    (elixir) lib/kernel/parallel_compiler.ex:211: anonymous fn/4 in Kernel.ParallelCompiler.spawn_workers/6
```

### DIFF
--- a/guides/telemetry.md
+++ b/guides/telemetry.md
@@ -13,7 +13,16 @@ handler function to any of the following event names:
 
 By default, only fields with a resolver get measured. You can override this
 by setting `absinthe_telemetry` in its metadata to `true` or `false` with
-`Absinthe.Schema.Notation.meta/1`.
+`Absinthe.Schema.Notation.meta/1`. For example:
+
+```elixir
+field :greeting, :string do
+  resolve fn _, _, _ -> {:ok, "Hello world"} end
+
+  # must come after `resolve`
+  meta absinthe_telemetry: false
+end
+```
 
 For async, batch, and dataloader fields, Absinthe sends the final event when
 it gets the results. That might be later than when the results are ready. If
--- a/lib/absinthe/blueprint/schema.ex
+++ b/lib/absinthe/blueprint/schema.ex
@@ -240,8 +240,11 @@ defmodule Absinthe.Blueprint.Schema do
 
   defp update_private(existing_private, private) do
     Keyword.merge(existing_private, private, fn
-      _, v1, v2 ->
+      _, v1, v2 when is_list(v1) and is_list(v2) ->
         update_private(v1, v2)
+
+      _, _v1, v2 ->
+        v2
     end)
   end
 end
--- a/lib/absinthe/schema/notation.ex
+++ b/lib/absinthe/schema/notation.ex
@@ -536,7 +536,7 @@ defmodule Absinthe.Schema.Notation do
     |> recordable!(:resolve, @placement[:resolve])
 
     quote do
-      meta :absinthe_telemetry, true
+      meta :absinthe_telemetry_implicit, true
       middleware Absinthe.Resolution, unquote(func_ast)
     end
   end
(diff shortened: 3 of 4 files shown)


darrenclark:
This unfortunately breaks code like https://github.com/opencensus-beam/opencensus_absinthe/blob/df4b6cb15a684bec67c56d4c1fce6d450a4c33e3/lib/opencensus/absinthe.ex#L90 that checks for only `absinthe_telemetry`

Some potential options:

1. Don't worry about it
    - A quick Github code search suggests that `opencensus_absinthe` is the only open source project that will be impacted by this change (https://github.com/search?q=absinthe_telemetry&type=Code), I can submit a PR so it checks `absinthe_telemetry_implicit` too
    - `opencensus_absinthe` mentions eventually using the telemetry events to power tracing: https://github.com/opencensus-beam/opencensus_absinthe#next-steps, in which case this change eventually won't matter
1. Have two separate `meta` collections (user defined & Absinthe-defined) that get merged during schema compilation, user-defined would take precedence over Absinthe-defined
    - might be valuable if we expect more Absinthe-defined flags to end up in `meta` in the future
1. Have a separate key in `Absinthe.Blueprint.Schema.FieldDefinition` for enabling/disabling telemetry, rather than using `meta`
    - Would need to add something to `Absinthe.Schema.Notation` for schema's to enable/disable this per field

benwilson512:
Hey Darren. I'm not a huge fan of having two different keys that configure telemetry. If people want to disable telemetry for fields with resolvers can't they just do:

```
resolve &Foo.bar/3
meta absinthe_telemetry: false
```

If they want to do so on a more general scale they can always add a schema phase that manipulates the meta key directly.

darrenclark:
> I'm not a huge fan of having two different keys that configure telemetry

Makes sense, agreed

There is a bug in `Absinthe.Blueprint.Schema.update_private/2` preventing that, but I will update this PR to fix it.

Will also document that `meta absinthe_telemetry: false` should come after the resolver

darrenclark:
@benwilson512 Updated the PR so that only `absinthe_telemetry` is used for enabling/disabling telemetry

binaryseed:
tbh, I don't really love that we add this `meta` field at all, i'm going to look into seeing if we can drop it. I don't think "disabling" telemetry is really useful - if there are no handlers installed then it's really a no-op ( just a quick ets lookup)

binaryseed:
The flag has been removed as of #793

PR #793, opened 2019-10-11
Ensure Telemetry is installed on SDL hydrate based resolvers

binaryseed:
The way we were ensuring that Telemetry was installed on resolvers only worked for Macro based schemas. This PR ensures that resolvers get installed on SDL based schemas as well

* Remove the `meta :absinthe_telemetry` flag
* Add `resolution` to metadata for resolver start event so the field can be identified

### DIFF
--- a/lib/absinthe/schema/notation.ex
+++ b/lib/absinthe/schema/notation.ex
@@ -536,7 +536,6 @@ defmodule Absinthe.Schema.Notation do
     |> recordable!(:resolve, @placement[:resolve])
 
     quote do
-      meta :absinthe_telemetry, true
       middleware Absinthe.Resolution, unquote(func_ast)
     end
   end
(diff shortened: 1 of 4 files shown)


lib/absinthe/schema/notation.ex
```diff
@@ -1683,12 +1682,16 @@ defmodule Absinthe.Schema.Notation do
     [{Absinthe.Middleware.MapGet, identifier}]
   end
 
-  def __ensure_middleware__(middleware, field, _object) do
-    if Absinthe.Type.meta(field, :absinthe_telemetry) do
-      [{Absinthe.Middleware.Telemetry, []} | middleware]
-    else
-      middleware
-    end
+  # Don't install Telemetry middleweare for Introspection fields
```
maartenvanvliet:
```suggestion
  # Don't install Telemetry middleware for Introspection fields
```

lib/absinthe/schema/notation.ex
```diff
@@ -1683,12 +1682,16 @@ defmodule Absinthe.Schema.Notation do
...
+  # Don't install Telemetry middleware for Introspection fields
+  @introspection [Absinthe.Phase.Schema.Introspection, Absinthe.Type.BuiltIns.Introspection]
+  def __ensure_middleware__(middleware, %{definition: definition}, _object)
+      when definition in @introspection do
+    middleware
+  end
+
+  # Install Telemetry middleware
+  def __ensure_middleware__(middleware, _field, _object) do
+    [{Absinthe.Middleware.Telemetry, []} | middleware]
```
benwilson512: This installs it _literally_ every field right? This will have non trivial overhead on MapGet fields.

binaryseed: No I don't believe so, my understanding is that the `__ensure_middleware__` function heads above install `Absinthe.Middleware.MapGet` alone for the default middleware case...

binaryseed: @benwilson512 ping

PR #794, opened 2019-10-18
Doc fixes

maartenvanvliet:
Some doc fixes
 * Added documented modules to the doc structure in mix.exs
 * marked undocumented modules with @moduledoc false
 * Some small fixes


benwilson512:
Thanks!

Issue #791, opened 2019-10-01
Scalar parse function called when input not a scalar


benwilson512:
We should only call the scalar parse functions if the input is a valid scalar literal.

PR #795, opened 2019-10-18
fix: #791 custom scalar parse

hauptbenutzer:
Hi!

This PR addresses #791. The idea is to check for the correct input type in the parse phase. As far as I understand the spec, that means any built-in scalars are valid as input for a custom scalar. 

I've made some adjustments to the `ArgumentsSchema` fixture in order to be able to test the parse function being called. Let me know what you think.

Cheers!


benwilson512:
Looks perfect, thank you!

PR #796, opened 2019-10-22
Fix default dataloader middleware when get_policy is 'tuples'

newtrat:
**Description**
This PR fixes the default implementation of `Absinthe.Resolution.Helpers.dataloader` for loaders using `get_policy: :tuples`.

**Context**
`Absinthe.Resolution.Helpers.dataloader/3` takes a `callback` option which must return an ok/error tuple. The default callback assumes that `Dataloader.get/4` returns a raw value, and wraps it in an ok tuple. When the loader's `get_policy` is `:tuples`, however, the returned value is already wrapped in an ok/error tuple, so the default callback adds an extra `:ok`, which causes a `BadMapError` in `Absinthe.Middleware.MapGet` down the line.

The practical effect is that loaders using `get_policy: :tuples` must pass a custom callback do `dataloader/3`. They also can't use `dataloader/1` and `dataloader/2`, since those functions use the default callback.


benwilson512:
Thanks!

PR #800, opened 2019-10-31
Fix crash when an inline fragment causes a query to go over complexity limits

antew:
I noticed this is v1.4.16 so it may be good to port it back there as well.

The `describe_node` function was missing a variant for inline fragments, so if the inline fragment caused it to go over the complexity limit you would get a crash with a message like:

```
** (FunctionClauseError) no function clause matching in Absinthe.Phase.Document.Complexity.Result.describe_node/1

     The following arguments were given to Absinthe.Phase.Document.Complexity.Result.describe_node/1:

         # 1
         %Absinthe.Blueprint.Document.Fragment.Inline{complexity: 100, directives: []
```

If you comment out the added code in `lib/absinthe/phase/document/complexity/result.ex` and run `mix test` it'll reproduce the crash.

Also, thank you to everyone who works on Absinthe, I'm relatively new to Elixir but it has been great to work with! I just stumbled on this error when adding complexity limiting to an application at work.

### DIFF
--- a/lib/absinthe/phase/document/complexity/result.ex
+++ b/lib/absinthe/phase/document/complexity/result.ex
@@ -79,4 +79,8 @@ defmodule Absinthe.Phase.Document.Complexity.Result do
   defp describe_node(%Blueprint.Document.Fragment.Spread{name: name}) do
     "Spread #{name}"
   end
+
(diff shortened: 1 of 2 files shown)


benwilson512:
Thank you!

Issue #807, opened 2019-11-15
Cannot run Absinthe.Schema.introspect in compile phase because of telemetry


jfrolich:
```
Compiling 1 file (.ex)
** (ArgumentError) argument error
    (stdlib) :ets.lookup(:telemetry_handler_table, [:absinthe, :execute, :operation, :start])
    /Users/Jaap/Documents/development/family-five/server/deps/telemetry/src/telemetry_handler_table.erl:47: :telemetry_handler_table.list_for_event/1
    /Users/Jaap/Documents/development/family-five/server/deps/telemetry/src/telemetry.erl:114: :telemetry.execute/3
    lib/absinthe/phase/telemetry.ex:18: Absinthe.Phase.Telemetry.run/2
    lib/absinthe/pipeline.ex:366: Absinthe.Pipeline.run_phase/3
    lib/absinthe.ex:110: Absinthe.run/3
    lib/mix/tasks/compile.schema_json.ex:12: Mix.Tasks.Compile.SchemaJson.run/1
    (mix) lib/mix/task.ex:331: Mix.Task.run_task/3
    (mix) lib/mix/tasks/compile.all.ex:73: Mix.Tasks.Compile.All.run_compiler/2
    (mix) lib/mix/tasks/compile.all.ex:53: Mix.Tasks.Compile.All.do_compile/4
    (mix) lib/mix/tasks/compile.all.ex:24: anonymous fn/1 in Mix.Tasks.Compile.All.run/1
    (mix) lib/mix/tasks/compile.all.ex:40: Mix.Tasks.Compile.All.with_logger_app/1
    (mix) lib/mix/task.ex:331: Mix.Task.run_task/3
    (mix) lib/mix/tasks/compile.ex:96: Mix.Tasks.Compile.run/1
    (mix) lib/mix/task.ex:331: Mix.Task.run_task/3
    (mix) lib/mix/tasks/app.start.ex:58: Mix.Tasks.App.Start.run/1
    (mix) lib/mix/task.ex:331: Mix.Task.run_task/3
    (mix) lib/mix/tasks/run.ex:128: Mix.Tasks.Run.run/5
    (mix) lib/mix/tasks/run.ex:85: Mix.Tasks.Run.run/1
    (mix) lib/mix/task.ex:331: Mix.Task.run_task/3
```

Would be great to have an option to drop the  telemetry phase. Otherwise generating a schema on compile is not possible anymore.

benwilson512:
[...]

PR #808, opened 2019-11-15
Add telemetry flag to Absinthe.run and resolution

maartenvanvliet:
See https://github.com/absinthe-graphql/absinthe/issues/807

The flag to Absinthe.run ensure the Telemetry phases are skipped in the
document pipeline.

Furthermore, the flag is passed on the
`Absinthe.Phase.Document.Execution.Resolution` phase where it sets
the (added) telemetry flag on the resolution struct. This is done so the
telemetry middleware can be skipped when the flag is set to false.

Let me know if this is the right approach :)

### DIFF
--- a/lib/absinthe.ex
+++ b/lib/absinthe.ex
@@ -68,6 +68,7 @@ defmodule Absinthe do
   executing an operation.
   * `:max_complexity` -> An integer (or `:infinity`) for the maximum allowed
   complexity for the operation being executed.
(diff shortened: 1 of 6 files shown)


binaryseed:
I would rather not have flags to get around this, it looks like:

* The mix task should start absinthe
* Telemetry shouldn't blow up an app if it hasn't been started

maartenvanvliet:
👍 Yes, I'd also prefer if this was not needed. I'll close this pr for now

PR #811, opened 2019-11-23
Fix SDL doc

steverice:
Fixes what looks like a copy/paste error in the `absinthe.schema.sdl` documentation.


benwilson512:
Thanks!

PR #818, opened 2019-12-07
Fix issue with sub_triggers

binaryseed:
This PR fixes an issue introduced in this commit:

https://github.com/absinthe-graphql/absinthe/commit/522790fa870ab34fe041b43ad06f7d562bc0342b

Tests on `master` have been failing ever since.

I assume the change was intended to ensure that `sub_triggers` could  only ever be a Map, but looks like it was actually causing it to never get through that step in the `for` comprehension.

@benwilson512


benwilson512:
Ah yes, good catch thanks!

PR #819, opened 2019-12-07
Fix infinite loop in interface covariance calculation

### LINKED ISSUE #597: Absinthe hangs indefinitely when an interface returns itself in a field
I mentioned this to @benwilson512 on Slack a month or so and I totally forgot about this. Here's a [commit](https://github.com/kenips/absinthe/commit/f3ed6490d3f595e414d9a46272ce90c9374d35d6) demonstrating the infinite hangs. I am not sure if an interface _can_ define a field that returns itself, but I stumble upon it when I simply wanted all of the implementations to support returning their parents.

### Steps to reproduce
Take the patch from the commit above, then run `mix test test/absinthe/resolution_test.exs`.

### Environment
* Elixir version (elixir -v): 1.7
* Absinthe version (mix deps | grep absinthe): master

### Expected behavior
If this is supported, then expecting this to work. If not supported, expecting an error message.

### Actual behavior
Hangs indefinitely at compile time. Maybe infinite macro expansion. Curious what it is myself...

N/A

binaryseed:
A particular usage of interface fields led to an infinite loop in `check_covariant`

closes #597 
@benwilson512

### DIFF
--- a/lib/absinthe/phase/schema/validation/object_must_implement_interfaces.ex
+++ b/lib/absinthe/phase/schema/validation/object_must_implement_interfaces.ex
@@ -146,6 +146,16 @@ defmodule Absinthe.Phase.Schema.Validation.ObjectMustImplementInterfaces do
   end
 
   defp check_covariant(
+         %Blueprint.Schema.InterfaceTypeDefinition{identifier: interface_ident},
+         field_type,
+         field_ident,
+         types
+       ) do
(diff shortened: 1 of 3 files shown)


binaryseed:
The additions I made to the `test/absinthe/resolution_test.exs` are ones that caused the infinite loop.. is that enough?

binaryseed:
I see, testing that it can be queried is a good call, i'll do that..

Issue #772, opened 2019-08-20
Support extend in SDL schema


binaryseed:
When defining a schema with SDL, the `extend` keyword should add the fields to the existing structure

Example from the spec:
https://graphql.github.io/graphql-spec/draft/#sec-Object-Extensions

```graphql
extend type Query {
  newField: String
}
```

mayel:
What is missing to get extend functionality?

benwilson512:
@mayel basically everything. The parser works I think, but we need a first class entity to model the extension values, and then compilation phases to integrate them in.

mayel:
Thanks. And this PR doesn't do any of that?  https://github.com/absinthe-graphql/absinthe/pull/821

binaryseed:
No, my PR was more of a shortcut, but that turned out not to be sufficient since it didn't treat the extension as a first class concept with structs / etc. It'll probably still be helpful but it needs a different approach

mmmries:
@binaryseed I was just looking into this as part of the work I'm doing on [absinthe_compose](https://hex.pm/packages/absinthe_compose). I'm hoping to support a feature where you can `import_sdl` for an existing server and have proxy all requests for those fields to a single backend, and then sprinkle in other definitions. It would help to support the use case of breaking up a big legacy API by pulling off the different fields one-by-one.

Is there anything I can do to help look into supporting `extend`?

binaryseed:
It's a big task, it will involve adding the `Language` and `Blueprint` extension structs, then using a phase to merge them in when the final schema is being generated.

PR #821, opened 2019-12-09
Add extend to macro defined schemas

binaryseed:
This PR introduces an `extend` macro that behaves as the "extend" concept in the GraphQL spec:

* https://graphql.github.io/graphql-spec/draft/#sec-Object-Extensions

We can extend `object`, `input_object`, `enum`, `union`, and `interface`.

The GraphQL spec allows extending `schema` and `scalar` but that isn't supported here - there isn't a `schema` macro to extend and scalars can only be extended to have more directives which also isn't supported in macro schemas

For now this works with macro based schemas, but it should be possible to adapt this to SDL based schemas as well (see #772)

### DIFF
--- a/lib/absinthe/schema/notation.ex
+++ b/lib/absinthe/schema/notation.ex
@@ -28,8 +28,7 @@ defmodule Absinthe.Schema.Notation do
     Module.register_attribute(__CALLER__.module, :absinthe_blueprint, accumulate: true)
     Module.register_attribute(__CALLER__.module, :absinthe_desc, accumulate: true)
     put_attr(__CALLER__.module, %Absinthe.Blueprint{schema: __CALLER__.module})
-    Module.put_attribute(__CALLER__.module, :absinthe_scope_stack, [:schema])
-    Module.put_attribute(__CALLER__.module, :absinthe_scope_stack_stash, [])
+    put_scope_stack(__CALLER__.module, [:schema])
 
     quote do
       import Absinthe.Resolution.Helpers,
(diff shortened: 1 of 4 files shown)


benwilson512:
Your strategy so far looks sound, is there any particular part you'd like feedback on?

binaryseed:
At first I wasn't sure this would work but I can see it coming together now, I'll let you know if anything arises!

binaryseed:
So this works for straight forward things like `object`, `input_object`, `enum` just fine

It's a little tricky with `union` - I got it to merge `types` which is cool but it seems pretty complicated to handle stuff like `resolve_type`... I guess in that situation people could use `is_type_of` 

The spec lets you extend just about everything via SDL but we probably don't have to go quite that far to get a lot of value...

lib/absinthe/schema/notation.ex
```diff
@@ -28,8 +28,7 @@ defmodule Absinthe.Schema.Notation do
     Module.register_attribute(__CALLER__.module, :absinthe_blueprint, accumulate: true)
     Module.register_attribute(__CALLER__.module, :absinthe_desc, accumulate: true)
     put_attr(__CALLER__.module, %Absinthe.Blueprint{schema: __CALLER__.module})
-    Module.put_attribute(__CALLER__.module, :absinthe_scope_stack, [:schema])
-    Module.put_attribute(__CALLER__.module, :absinthe_scope_stack_stash, [])
+    put_scope_stack(__CALLER__.module, [:schema])
```
benwilson512: @binaryseed In master I added https://github.com/absinthe-graphql/absinthe/blob/master/lib/absinthe/schema/notation.ex#L1508 to deal with the fact that we needed two stacks to handle stack stashing for Absinthe Relay, we should unify these helpers.

Leeaandrob:
Any news about the issue?

benwilson512:
This has since been implemented in a different PR.

PR #822, opened 2019-12-10
Implement placement enforcement

### LINKED ISSUE #820: Placement checking for schema notation macros
We need to check the `@placement` annotations when building the schema based on macros

`master`:
* https://github.com/absinthe-graphql/absinthe/blob/master/lib/absinthe/schema/notation.ex#L1751

Logic in `1.4`:
* https://github.com/absinthe-graphql/absinthe/blob/v1.4.16/lib/absinthe/schema/notation.ex#L1608

binaryseed:
This PR implements placement checking in the new 1.5 schemas

@benwilson512 @bruce 
closes #820

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -32,8 +32,6 @@ locals_without_parens = [
   object: 3,
   on: 1,
   parse: 1,
-  record_object!: 4,
-  recordable!: 4,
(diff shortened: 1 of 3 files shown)


benwilson512:
🎉 this is excellent

Issue #726, opened 2019-05-05
Bad operation name can result in exception


xxhenglyxx:
### Environment

* Elixir version (elixir -v):
```
Erlang/OTP 21 [erts-10.3.4] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe] [dtrace]

Elixir 1.8.1 (compiled with Erlang/OTP 21)
```
* Absinthe version (mix deps | grep absinthe):
```
* absinthe 1.4.16 (Hex package) (mix)
  locked at 1.4.16 (absinthe) 0933e4d9
* absinthe_ecto 0.1.3 (Hex package) (mix)
  locked at 0.1.3 (absinthe_ecto) 420b6812
* absinthe_plug 1.4.7 (Hex package) (mix)
  locked at 1.4.7 (absinthe_plug) 939b6b9e
* absinthe_phoenix 1.4.4 (Hex package) (mix)
  locked at 1.4.4 (absinthe_phoenix) af3b7b44
```
* Client Framework and version (Relay, Apollo, etc): `graphql_flutter: 1.0.1-beta.7`

### Expected behavior

No error and log out any result from an operation(query or mutation) on the client side(flutter)

### Actual behavior

```
[error] #PID<0.3584.0> running AppServerWeb.Endpoint (connection #PID<0.3583.0>, stream id 1) terminated
Server: 10.0.2.2:4000 (http)
Request: POST /api/graphql
** (exit) an exception was raised:
    ** (FunctionClauseError) no function clause matching in Absinthe.Phase.Document.Result.data/2
        (absinthe) lib/absinthe/phase/document/result.ex:50: Absinthe.Phase.Document.Result.data(nil, [])
        (absinthe) lib/absinthe/phase/document/result.ex:19: Absinthe.Phase.Document.Result.process/1
        (absinthe) lib/absinthe/phase/document/result.ex:11: Absinthe.Phase.Document.Result.run/2
        (absinthe) lib/absinthe/pipeline.ex:274: Absinthe.Pipeline.run_phase/3
        (absinthe_plug) lib/absinthe/plug.ex:421: Absinthe.Plug.run_query/4
        (absinthe_plug) lib/absinthe/plug.ex:247: Absinthe.Plug.call/2
        (phoenix) lib/phoenix/router/route.ex:39: Phoenix.Router.Route.call/2
        (phoenix) lib/phoenix/router.ex:275: Phoenix.Router.__call__/1
[...]

PR #824, opened 2019-12-12
Return error if operation name did not match single operation in document

maartenvanvliet:
In https://github.com/absinthe-graphql/absinthe/blob/2d423f6ad45cb3e1c194d3cb0843fd358d298b2d/lib/absinthe/phase/document/current_operation.ex#L26
there is the edge case that when only one operation was set in the input
and it did not match the `:operation_name` in the options, there would be
no current operation.

In `Absinthe.Phase.Document.Validation.SelectedCurrentOperation` it would
usually be caught that there is no current operation set. However, it was
assumed this would be happen when there were more than 1 operations in the
document. https://github.com/absinthe-graphql/absinthe/blob/2d423f6ad45cb3e1c194d3cb0843fd358d298b2d/lib/absinthe/phase/document/validation/selected_current_operation.ex#L18

This was changed, when there is 1 operation in the document but no
current operation is found an error is returned.

I've also expanded on the error messages that may be returned as to
better discern what happens in these cases

 * No operation name given but multiple in document
 * Operation name given but did not match any of the multiple in the document
 * Operation name given but did not match the single operation in the document

Another approach could have been to ignore the `:operation_name` when
only one operation is given in the document, and set this operation as
the current operation. I think this hides the error.

Fixes https://github.com/absinthe-graphql/absinthe/issues/726

### DIFF
--- a/lib/absinthe/phase/document/validation/selected_current_operation.ex
+++ b/lib/absinthe/phase/document/validation/selected_current_operation.ex
@@ -38,10 +38,9 @@ defmodule Absinthe.Phase.Document.Validation.SelectedCurrentOperation do
     }
   end
 
-  @error_message "Must provide a valid operation name if query contains multiple operations."
   def error_message(nil, _) do
     """
-    #{@error_message}
+    Must provide a valid operation name if query contains multiple operations.
 
     No operation name was given.
     """
@@ -57,7 +56,7 @@ defmodule Absinthe.Phase.Document.Validation.SelectedCurrentOperation do
 
   def error_message(operation_name, _) do
     """
-    #{@error_message}
+    Must provide a valid operation name if query contains multiple operations.
 
     The provided operation name was: #{inspect(operation_name)}
     """
--- a/lib/absinthe/phase/document/validation/selected_current_operation.ex
+++ b/lib/absinthe/phase/document/validation/selected_current_operation.ex
@@ -12,13 +12,15 @@ defmodule Absinthe.Phase.Document.Validation.SelectedCurrentOperation do
   Run the validation.
   """
   @spec run(Blueprint.t(), Keyword.t()) :: Phase.result_t()
-  def run(input, _options \\ []) do
+  def run(input, options \\ []) do
     result =
       case {Blueprint.current_operation(input), length(input.operations)} do
-        {nil, count} when count > 1 ->
+        {nil, count} when count >= 1 ->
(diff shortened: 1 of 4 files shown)


test/absinthe/integration/execution/operation_by_name_test.exs
```diff
@@ -85,4 +91,28 @@ defmodule Elixir.Absinthe.Integration.Execution.OperationByNameTest do
...
+  }
+  """
+
+  test "return error when single operation in document does not match given operation name" do
+    assert {:ok,
+            %{
+              errors: [
+                %{
+                  message: """
+                  The provided operation name dit not match the operation in the query.
```
binaryseed:
```suggestion
                  The provided operation name did not match the operation in the query.
```

lib/absinthe/phase/document/validation/selected_current_operation.ex
```diff
@@ -28,15 +30,36 @@ defmodule Absinthe.Phase.Document.Validation.SelectedCurrentOperation do
...
+    """
+    #{@error_message}
+
+    No operation name was given.
+    """
+  end
+
+  def error_message(operation_name, 1) do
+    """
+    The provided operation name dit not match the operation in the query.
```
binaryseed:
```suggestion
    The provided operation name did not match the operation in the query.
```

lib/absinthe/phase/document/validation/selected_current_operation.ex
```diff
@@ -28,15 +30,36 @@ defmodule Absinthe.Phase.Document.Validation.SelectedCurrentOperation do
...
     %Phase.Error{
       phase: __MODULE__,
-      message: error_message()
+      message: error_message(operation_name, operation_count)
     }
   end
 
-  def error_message do
-    ~s(Must provide a valid operation name if query contains multiple operations.)
+  @error_message "Must provide a valid operation name if query contains multiple operations."
```
binaryseed: Not sure this needs to be a module attribute, it's static and only used 2 times...

maartenvanvliet: I think I copied this from somewhere else in the code but was also unsure about it. Will fix

maartenvanvliet: Fixed :)

binaryseed:
Thanks!

PR #827, opened 2019-12-19
Bump to v1.5.0-rc.0

bruce:
Releasing `rc0` after discussion with @binaryseed.


jeroenvisser101:
@bruce it seems the tag/release for this is missing? That causes links from the docs to the source to also 404.

PR #829, opened 2019-12-20
Fix skipped list_of validation

### LINKED ISSUE #752: list_of validation error does not stop the resolution

### Environment

* Elixir version (elixir -v):
```
Erlang/OTP 22 [erts-10.4.2] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe]
Interactive Elixir (1.8.2)
```
* Absinthe version (mix deps | grep absinthe): `1.4.*` and `master`
* Client Framework and version (Relay, Apollo, etc): not relevant

### Expected behavior

stop resolution on non list type

### Actual behavior

returns `null`

[https://github.com/kaaboaye/absinthe_bug/tree/absinthe-master](https://github.com/kaaboaye/absinthe_bug/tree/absinthe-master)

For this schema

```ex
defmodule AbsintheBug.Schema do
  use Absinthe.Schema

  object :key_value do
    field(:key, non_null(:string))
    field(:value, non_null(:string))
  end

  input_object :key_value_input do
    field(:key, non_null(:string))
    field(:value, non_null(:string))
  end

  query do
    field :test, list_of(non_null(:key_value)) do
      arg(:list, non_null(list_of(non_null(:key_value_input))))

      resolve(fn _, a, _ ->
        {:ok, a[:list]}
      end)
    end
  end
end
```

and following tests

```ex
defmodule AbsintheBug.SchemaTest do
  use ExUnit.Case

  test "positive case" do
    res =
      """
      {
        test(list: [{key: "a", value: "1"}, {key: "b", value: "2"}]) {
          key
          value
        }
      }
      """
      |> Absinthe.run!(AbsintheBug.Schema)

    assert %{
             data: %{"test" => [%{"key" => "a", "value" => "1"}, %{"key" => "b", "value" => "2"}]}
           } == res
  end

  test "integer instead of list" do
    res =
      """
      {
        test(list: 1234) {
          key
          value
        }
      }
      """
      |> Absinthe.run!(AbsintheBug.Schema)

    assert %{errors: [%{locations: [%{column: _, line: _}]}]} = res
  end

  test "string instead of list" do
    res =
      """
      {
        test(list: "abcd") {
          key
          value
        }
      }
      """
      |> Absinthe.run!(AbsintheBug.Schema)

    assert %{errors: [%{locations: [%{column: _, line: _}]}]} = res
  end
end
```

tests number `2` and `3` fails with following output

```
  1) test integer instead of list (AbsintheBug.SchemaTest)
     test/schema_test.exs:21
     match (=) failed
     code:  assert %{errors: [%{locations: [%{column: _, line: _}]}]} = res
     right: %{data: %{"test" => nil}}
     stacktrace:
       test/schema_test.exs:33: (test)

  2) test string instead of list (AbsintheBug.SchemaTest)
     test/schema_test.exs:36
     match (=) failed
     code:  assert %{errors: [%{locations: [%{column: _, line: _}]}]} = res
     right: %{data: %{"test" => nil}}
     stacktrace:
       test/schema_test.exs:48: (test)
```

binaryseed:
This PR adds in a missing validation inside of `Absinthe.Phase.Document.Arguments.Parse`

The validation catches cases when a Scalar value is provided for a `schema_node` that is an Input Object.

This fixes an issue that caused Absinthe to allow invalid arguments as described in #752 

fixes #752

### DIFF
--- a/lib/absinthe/phase/document/arguments/parse.ex
+++ b/lib/absinthe/phase/document/arguments/parse.ex
@@ -72,6 +72,11 @@ defmodule Absinthe.Phase.Document.Arguments.Parse do
     build_value(normalized, inner_type, context)
   end
 
+  defp build_value(%{__struct__: struct} = _normalized, _not_scalar, _context)
+       when struct in [Input.Boolean, Input.Float, Input.Integer, Input.String] do
+    {:error, :bad_parse}
+  end
+
   defp build_value(_, _, _) do
     :not_leaf_node
   end
--- a/lib/absinthe/phase/document/arguments/parse.ex
+++ b/lib/absinthe/phase/document/arguments/parse.ex
@@ -72,7 +72,7 @@ defmodule Absinthe.Phase.Document.Arguments.Parse do
     build_value(normalized, inner_type, context)
   end
 
-  defp build_value(%{__struct__: struct} = _normalized, _not_scalar, _context)
+  defp build_value(%{__struct__: struct}, %Type.InputObject{}, _)
        when struct in [Input.Boolean, Input.Float, Input.Integer, Input.String] do
     {:error, :bad_parse}
   end
--- a/test/absinthe/phase/document/validation/arguments_of_correct_type_test.exs
+++ b/test/absinthe/phase/document/validation/arguments_of_correct_type_test.exs
@@ -802,6 +802,37 @@ defmodule Absinthe.Phase.Document.Validation.ArgumentsOfCorrectTypeTest do
       )
     end
 
+    test "Partial object, list with correct value" do
+      assert_passes_validation(
+        """
+        {
+          complicatedArgs {
+            complexArgField(complexArgList: [{ requiredField: true }])
+          }
+        }
+        """,
+        []
+      )
+    end
+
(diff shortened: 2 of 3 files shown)


binaryseed:
@bruce or @benwilson512 Could you point me towards the location this validation should be happening at? I waded around a bit but couldn't quite locate it..

maartenvanvliet:
I did some investigation on this a while back:

In the `Absinthe.Phase.Document.Validation.ArgumentsOfCorrectType` phase, the incorrect input_object arguments are matched by the 
```elixir
  defp handle_node(%Blueprint.Input.Argument{schema_node: nil} = node, _schema) do
    {:halt, node}
  end
``` 
function and it did not add any validation errors. I did not find out whether `schema_node` should have been set for this case or not. 

If I  matched on `%Blueprint.Input.Argument{flags: %{invalid: _}}, schema_node: nil}` it would return the incorrect input objects, so they were flagged as invalid by an earlier phase. However, running the entire test suite it failed several other tests.

binaryseed:
Ok, I think this one is solved!

PR #832, opened 2019-12-28
Remove Absinthe.Traversal

maartenvanvliet:
The new schema flow no longer requires `Absinthe.Traversal` for checking
schema correctness. It all uses Blueprint now.

The only module depending on Traversal was `Absinthe.Schema.Verification`
but it was not invoked anywhere, so it (and its dependents) could be removed
as well.

All code implementing the `Absinthe.Traversal.Node` protocol was also
removed.

If any external code relies on Absinthe.Traversal it would break but
since it would be the 1.5.0 release this should not be a problem.


binaryseed:
Thanks!

PR #836, opened 2020-01-08
Set complexity of __typename to 0

binaryseed:
This PR sets the `complexity` score of `__typename` to 0. Some clients auto-add this field to every type in a query to do advanced functionality, and having `__typename` default to 1 can really increase the overall complexity score of a query.

Also I found a few functions / structs that had no impact on the codebase which I removed - they appear to be related to introspection in pre-1.5 schemas


lib/absinthe/type.ex
```diff
@@ -376,21 +376,6 @@ defmodule Absinthe.Type do
     true
   end
 
-  def field(_type, "__" <> meta_name) do
```
benwilson512: Ah yeah this is definitely old school introspection.

PR #840, opened 2020-01-09
allow hydrating builtins

benwilson512:
@bruce @binaryseed Hydration right now excludes builtins, but I'm inclined to say that we should allow people to mess with them. This starts from a concrete use case on our end (we need to patch the scalar serialization functions to be more lenient). More generally, I think it's just generally powerful to let users really have complete control over their schema.


binaryseed:
Looks good - maybe a little test that can serve as an example of leveraging this? That's how I learned how all the hydrate stuff works

binaryseed:
nice!

Issue #831, opened 2019-12-23
UndefinedFunctionError `nil.identifier/0` from `import_sdl` with a directive like `@foo(a: { b: {} })`


myronmarston:
### Environment

* Elixir version (elixir -v):

```
Erlang/OTP 22 [erts-10.6.1] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe]

Elixir 1.9.4 (compiled with Erlang/OTP 20)
```

* Absinthe version (mix deps | grep absinthe):

```
* absinthe 1.5.0-rc.0 (Hex package) (mix)
  locked at 1.5.0-rc.0 (absinthe) 5f828161
```

* Client Framework and version (Relay, Apollo, etc):

N/A: not using any client

### Expected behavior

I expect `import_sdl` to either properly parse and import my schema file or fail with a clear error

### Actual behavior

When calling `import_sdl` with a schema that has a field directive like `@some_directive(a: { b: {} })`, it crashes with this:

```
     ** (UndefinedFunctionError) function nil.identifier/0 is undefined
     stacktrace:
       nil.identifier()
       (absinthe) lib/absinthe/phase/schema/arguments/data.ex:32: anonymous fn/2 in Absinthe.Phase.Schema.Arguments.Data.handle_node/1
       (elixir) lib/enum.ex:1948: Enum."-reduce/3-lists^foldl/2-0-"/3
       (absinthe) lib/absinthe/phase/schema/arguments/data.ex:31: Absinthe.Phase.Schema.Arguments.Data.handle_node/1
       (absinthe) lib/absinthe/blueprint/transform.ex:47: anonymous fn/3 in Absinthe.Blueprint.Transform.postwalk/2
       (absinthe) lib/absinthe/blueprint/transform.ex:145: anonymous fn/4 in Absinthe.Blueprint.Transform.walk_children/5
       (elixir) lib/enum.ex:1948: Enum."-reduce/3-lists^foldl/2-0-"/3
       (absinthe) lib/absinthe/blueprint/transform.ex:114: Absinthe.Blueprint.Transform.walk/4
       (elixir) lib/enum.ex:1440: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
       (absinthe) lib/absinthe/blueprint/transform.ex:145: anonymous fn/4 in Absinthe.Blueprint.Transform.walk_children/5
[...]

PR #842, opened 2020-01-10
Validate directive arguments are known

myronmarston:
This is a failing test for #831, as requested by @binaryseed.

### DIFF
--- a/test/mix/tasks/absinthe.schema.sdl_test.exs
+++ b/test/mix/tasks/absinthe.schema.sdl_test.exs
@@ -52,4 +52,31 @@ defmodule Mix.Tasks.Absinthe.Schema.SdlTest do
       assert schema =~ "helloWorld(name: String!): String"
     end
   end
+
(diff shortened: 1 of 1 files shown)


binaryseed:
I'm noticing that if I actually define the arguments on the `directive`, everything works just fine:

```elixir
defmodule Directives do
  use Absinthe.Schema.Prototype

  directive :some_directive do
    arg :a, :directive_arg

    on [:field_definition]
  end

  input_object :directive_arg do
    field :b, :directive_nested_arg
  end

  input_object :directive_nested_arg do
    field :c, :string
  end
end
```

Perhaps what we need here is a validation that fails if unknown arguments are passed to the directive in the SDL. We have a `KnownDirectives` validation

cc @bruce

binaryseed:
From the spec, all arguments must be defined (same applies to directives & fields)

http://spec.graphql.org/June2018/#sec-Argument-Names

binaryseed:
Merging into a feature branch

Issue #845, opened 2020-01-16
SDL export excess types


pyromaniac:
With the following schema:

```elixir
defmodule Schema do
  use Absinthe.Schema

  object :query_fields do
    field :foo, :integer
  end

  query name: "Query" do
    import_fields :query_fields
  end
end
```

I perform `mix absinthe.schema.sdl --schema Schema`

And it looks like all the objects that are used for fields import are also reflected in the SDL dump.
It happens on any level, not only on the schema level.

JSON schema works as expected. I believe this is because JSON schema dump uses introspection query and SDL simply dumps everything in the namespace.

### Environment

* Elixir version (elixir -v): Elixir 1.9.4
* Absinthe version (mix deps | grep absinthe): 1.5.0-rc.2

### Expected behavior

It should produce:

```graphql
schema {
  query: Query
}

type Query {
  foo: Int
}
```

### Actual behavior

But produces:

```graphql
schema {
  query: Query
}

type QueryFields {
  foo: Int
}

type Query {
  foo: Int
}
```

PR #846, opened 2020-01-16
SDL renders unwanted types

pyromaniac:
A failing test for #845

### DIFF
--- a/test/absinthe/schema/sdl_render_test.exs
+++ b/test/absinthe/schema/sdl_render_test.exs
@@ -186,6 +186,7 @@ defmodule SdlRenderTest do
     object :order do
       field :id, :id
       field :name, :string
+      import_fields :imported_fields
     end
 
     object :category do
@@ -194,6 +195,10 @@ defmodule SdlRenderTest do
 
     union :search_result do
       types [:order, :category]
+    end
+
+    object :imported_fields do
+      field :imported, non_null(:boolean)
     end
   end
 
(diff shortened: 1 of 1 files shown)


benwilson512:
Merging into a bug fix branch, thanks!

binaryseed:
This one looks like it's because we're generating the SDL based off of `Blueprint` structs, and not off of the result of `Absinthe.Schema.used_types` which walks through the `referenced_types`

I'm not sure of any clear way to bridge the gap and trim down the blueprint structs to just those used - `used_types` returns `Absinthe.Type` structs, not `Blueprint`

@benwilson512 Any ideas here?

benwilson512:
@binaryseed I took a crack at implementing this earlier this week, it's a bit of a pain, basically we just have to reimplement the logic from `used_types`. On the upside, I think by doing it at the type definition stage we can actually mark the types as used or unused, and then `Absinthe.Schema.used_types` won't need to do any actual logic, it can just filter by `used`.

binaryseed:
That would be pretty great actually, it'd probably improve the performance of the introspection queries for large schemas...

You have a branch up?

benwilson512:
@binaryseed I'm working on it now, will push later today. Now that I'm thinking about it, do we even need to generate actual types for unused types? `import_fields` is run at the type def layer now, the actual object you're importing is completely irrelevant. It could be an issue if people were expecting to be able to look it up but there's just straight up better ways to do that in 1.5.

I think the ideal thing to do is mark the used type defs, and then at the Build phase only construct `Absinthe.Type` entities for those things that are actually used. `to_sdl` can filter on the `used` flag, and introspection stops needing to care about it at all.

binaryseed:
Sounds like a good direction, I can say we've never needed the intermediate types and have only interacted with them when needing to ignore them

PR #848, opened 2020-01-17
Mark used types better

### LINKED ISSUE #845: SDL export excess types

benwilson512:
Fixes #845
@binaryseed this solves the original issue where SDL would export undesired types. I've hit some snags trying to eliminate unused types entirely, will keep iterating on it.


lib/absinthe/schema.ex
```diff
@@ -556,6 +552,11 @@ defmodule Absinthe.Schema do
   @doc """
   Converts a schema to an SDL string
 
+  By default it excludes any types not referenced directly or transitively from
+  the root query, subscription, or mutation objects. This prevents objects
+  merely used as field imports from appearing in the SDL. If you wish all
+  types to appear, use the `:include_disconnected` option.
```
benwilson512: I hate this option name, not sure what a better one is. `:include_unused` would also work, there's also an option for `include: [:unused]` but I don't know what else could go in that list. `:scalars` or something?

test/absinthe/schema/sdl_render_test.exs
```diff
@@ -104,7 +104,8 @@ defmodule SdlRenderTest do
   end
 
   test "Render SDL from blueprint defined with SDL" do
-    assert Absinthe.Schema.to_sdl(SdlTestSchema) == SdlTestSchema.sdl()
+    assert Absinthe.Schema.to_sdl(SdlTestSchema, include_disconnected: true) ==
```
benwilson512: The SDL in that module has unattached types, so we need to use the option to make the test pass. We could also connect all the types, but I think it's helpful to have a test case showing that things can be preserved.

lib/absinthe/schema.ex
benwilson512: Other general options could include whether types should be alpha sorted or not. Right now we preserve the original order which is probably the right default, but I know many would prefer it if field / arg / types were all sorted alphabetically.

lib/absinthe/schema.ex
```diff
@@ -533,14 +533,10 @@ defmodule Absinthe.Schema do
...
-    |> Enum.map(&lookup_type(schema, &1))
-    |> Enum.concat(directives(schema))
-    |> Enum.filter(&(!is_nil(&1)))
-    |> Enum.flat_map(&Type.referenced_types(&1, schema))
-    |> MapSet.new()
+    schema.__absinthe_types__
+    |> Map.keys()
     |> Enum.map(&Schema.lookup_type(schema, &1))
-    |> Enum.filter(&(!Type.introspection?(&1)))
+    |> Enum.filter(&(!Type.introspection?(&1) && &1.__private__[:__absinthe_used__]))
```
benwilson512: We still get some nice performance benefits here because this just becomes a simple boolean filter instead of the complicated tree walk.

I do wonder if we should deprecate this function, I'm not sure that it has any use outside of trying to show stuff for introspection. Particularly if we can eliminate unused types from the compiled schema altogether I just don't see any point for it.

lib/absinthe/schema.ex
pyromaniac: I would definitely go with a single term on this, so `include_unused` is the best option. As for the `include: [:unused]`, if you don't yet know what else to put there, just use a more specific option for now. Later it will be possible to deprecate it painlessly.

test/absinthe/schema/sdl_render_test.exs
binaryseed: That's just an artifact of the previous situation - I'd be fine tweaking the test such that all of them are "used"...

lib/absinthe/schema.ex
binaryseed: I'd suggest we don't even bother - let's just match the same behavior as the Introspection mechanism

lib/absinthe/schema.ex
binaryseed: I know we call it in our codebase since we tweak Introspection to further filter results, so even if it's just for modifying Introspection, it's still useful...

lib/absinthe/schema.ex
binaryseed: Rendering un-used types via SDL was an unintended side effect of how I put it together in the first place

lib/absinthe/schema.ex
benwilson512: Sorry to be clear I agree that making it easy to get introspectable types is useful, I just wonder if the function would be better named `introspectable_types` or something of that sort. `used_types` is sorta misleading since there wouldn't be any _unused_ types anymore, and introspection types are definitely used, they just aren't introspectable.

binaryseed: Gotcha, that makes sense. The latest spec has this to say:

> When returning the set of types from the __Schema introspection type, all referenced built‐in scalars must be included. If a built‐in scalar type is not referenced anywhere in a schema (there is no field, argument, or input field of that type) then it must not be included.
> 
> When representing a GraphQL schema using the type system definition language, all built‐in scalars must be omitted for brevity.

https://spec.graphql.org/draft/#sec-Scalars.Built-in-Scalars

Given the spec language, maybe it's `referenced_types` that we want, and the flag in the structs should be `referenced`.

And since it's mentioned in the spec, do we want to have `sdl_types` that will just do the right thing? ("referenced_types - builtin")

Do you want to remove `used_types` or soft-deprecate?

lib/absinthe/schema.ex
```diff
@@ -575,6 +576,21 @@ defmodule Absinthe.Schema do
...
 
+    bp =
+      if Keyword.get(opts, :include_disconnected, false) do
+        bp
+      else
+        Map.update!(bp, :schema_definitions, fn schema_defs ->
+          for schema_def <- schema_defs do
+            Map.update!(schema_def, :type_definitions, fn type_defs ->
+              Enum.filter(type_defs, fn type_def ->
+                type_def.__private__[:__absinthe_used__]
```
binaryseed: The concept of "referenced" is in the spec, so maybe we could elevate it into the struct itself vs sticking it in `__private__`?

lib/absinthe/schema.ex
benwilson512: Gonna use referenced_types as the term. Instead of flagging the types, I'm currently doing:
```
        def __absinthe_types__() do
          __absinthe_types__(:used)
        end

        def __absinthe_types__(:used) do
          unquote(Macro.escape(used_types))
        end

        def __absinthe_types__(:all) do
          unquote(Macro.escape(type_list))
        end
```

This way `__absinthe_types__` by default just refers to referenced types.

test/absinthe/schema/sdl_render_test.exs
benwilson512: Given your comments elsewhere referencing the spec, I think this makes sense. People can always run the pipeline themselves if they want absolutely every type.

lib/absinthe/schema.ex
benwilson512: @binaryseed my concern with making it a first class field is that it isn't really something you should mess with yourself. Whether something is referenced or not requires a holistic look at the schema using relatively complex rules.

If people start needing it regularly then we can look at exposing it more directly. Right now it's a pretty low level detail.

binaryseed: makes sense

lib/absinthe/phase/schema/hydrate.ex
```diff
@@ -205,4 +205,16 @@ defmodule Absinthe.Phase.Schema.Hydrate do
 
     root
   end
+
+  def apply_hydration(root, result) do
```
pyromaniac: Should not this be deleted?

benwilson512: No, providing a dedicated error message is better than simply getting a standard "no clause matching" error.

PR #850, opened 2020-01-22
Fix Dialyzer errors and add Dialyzer to CI

devonestes:
There were Dialyzer errors in here that were causing Dialyzer errors in
applications that use Absinthe, so I've fixed them all and also added
Dialyzer to CI so that new errors don't pop up again in the future.


benwilson512:
Thank you!

Issue #852, opened 2020-01-23
Running validation-only pipeline barfs on good data.


ekobi:
This is an issue I reported on the Elixir forum, for which @benwilson512 provided a workaround. Wondering if either
- `Absinthe.Pipeline.run()` could handle applying the workaround automatically;

or, alternatively
- `Absinthe.Pipeline` could provide a `for_document_validation()` function that builds a suitably patched input for `run()`

If submitting a bug, please provide the following:

### Environment

* Elixir version (elixir -v): Elixir 1.9.2 (compiled with Erlang/OTP 22)
* Absinthe version (mix deps | grep absinthe):
  - absinthe 1.4.16 (Hex package) (mix)
     locked at 1.4.16 (absinthe) 0933e4d9
  - absinthe_plug 1.4.7 (Hex package) (mix)
     locked at 1.4.7 (absinthe_plug) 939b6b9e
  - absinthe_phoenix 1.4.4 (Hex package) (mix)
    locked at 1.4.4 (absinthe_phoenix) af3b7b44`
* Client Framework and version (Relay, Apollo, etc): ExUnit test framework

### Expected behavior
`Absinthe.Pipeline.run()` expected to validate provided executable GQL document and parameters against schema.

```elixir
  def validate_graphql_docs() do
    options = [operation_name: "SomeMutation", variables: %{"message" => "string message"}]
    pipeline =
      My.Api.Schema
      |> Absinthe.Pipeline.for_document(options)
      |> Absinthe.Pipeline.without(Absinthe.Phase.Subscription.SubscribeSelf)
      |> Absinthe.Pipeline.without(Absinthe.Phase.Document.Execution.Resolution)
    case Absinthe.Pipeline.run(@badMutation, pipeline) do
      {:ok, %{execution: %{validation_errors: []}}, _} -> {:ok, []}
      {:ok, %{execution: %{validation_errors: validation_errors}}, _} -> {:error, validation_errors}
      other -> {:error, other}
    end
  end
```

### Actual behavior
Does the right thing given invalid data/parameters, but barfs on well-formed documents thus:
```elixir
     The following arguments were given to Absinthe.Phase.Document.Result.data/2:
         # 1
         nil
         # 2
         []
     Attempted function clauses (showing 5 out of 5):
         defp data(%{errors: [_ | _] = field_errors}, errors)
         defp data(%{value: nil}, errors)
         defp data(%{value: value, emitter: emitter}, errors)
[...]

PR #853, opened 2020-01-24
Fix issue #852, RE: validation-only pipeline robustness.

ekobi:
Implements recommended fix for Issue #852 in v1.4. Should also work in 1.5.


benwilson512:
Hi @ekobi please base this on 1.5, there will not be any further 1.4 releases outside of critical security issues if any are found.

binaryseed:
This was merged into master

PR #854, opened 2020-01-24
Fix issue #852, RE: validation-only pipeline robustness.

ekobi:
Applying to master as requested.

### DIFF
--- a/lib/absinthe/phase/document/result.ex
+++ b/lib/absinthe/phase/document/result.ex
@@ -15,6 +15,9 @@ defmodule Absinthe.Phase.Document.Result do
   defp process(blueprint) do
     result =
       case blueprint.execution do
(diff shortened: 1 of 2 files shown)


lib/absinthe/phase/document/result.ex
```diff
@@ -15,6 +15,9 @@ defmodule Absinthe.Phase.Document.Result do
   defp process(blueprint) do
     result =
       case blueprint.execution do
+        %{validation_errors: [], result: nil} ->
```
benwilson512: Seems sensible, can you create a test case that exercises this?

ekobi: Added a couple of tests; unsure of correct placement -- please advise better location.

ekobi: nm.

PR #855, opened 2020-01-24
Serialize Int result into Float

binaryseed:
The strict validation on results goes a bit further than what I can tell from this spec language:

> GraphQL servers may coerce non‐floating‐point internal values to Float when reasonable without losing information, otherwise they must raise a field error. Examples of this may include returning 1.0 for the integer number 1, or 123.0 for the string "123".

https://spec.graphql.org/draft/#sec-Float

@benwilson512

### DIFF
--- a/lib/absinthe/type/built_ins/scalars.ex
+++ b/lib/absinthe/type/built_ins/scalars.ex
@@ -39,6 +39,7 @@ defmodule Absinthe.Type.BuiltIns.Scalars do
   end
 
   def serialize_float(n) when is_float(n), do: n
+  def serialize_float(n) when is_integer(n), do: n * 1.0
 
   def serialize_float(n) do
     raise Absinthe.SerializationError, """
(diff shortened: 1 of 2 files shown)


binaryseed:
@benwilson512 Do you think we should adopt the "123" string -> float behavior too? Since all this is a `may` I'd probably say we're best off doing just numeric types

benwilson512:
Yeah, definitely disinclined to allow numeric values. These days it's entirely possible to replace the built in serialization function, so if someone really wants to relax it, all they have to do is hydrate the type and emit `{:serialize`.

binaryseed:
So you'd rather not coerce Int to Float here?

benwilson512:
Sorry the original post had a confusing typo, I think it's correct now. Int -> Float is fine, String -> Float is less fine, folks can add it themselves if they want it.

binaryseed:
Ok, I'm going to merge as-is. I don't think any other conversions should be done by default (string -> anything, nor float -> int due to the risk of actually changing the value)

PR #857, opened 2020-01-28
Optimize schema introspection type list

binaryseed:
Now that we know "referenced" types upfront (https://github.com/absinthe-graphql/absinthe/pull/848), we can avoid traversing the complete list 2 times in the Introspection query for `__schema.types`. This should help for schemas with lots of types!

Also a little tweak to avoid mutating the blueprint before we render it's SDL

### DIFF
--- a/lib/absinthe/schema.ex
+++ b/lib/absinthe/schema.ex
@@ -534,9 +534,8 @@ defmodule Absinthe.Schema do
   """
   @spec used_types(t) :: [Type.t()]
   def used_types(schema) do
-    schema.__absinthe_types__
-    |> Map.keys()
-    |> Enum.map(&Schema.lookup_type(schema, &1))
+    schema
+    |> Schema.types()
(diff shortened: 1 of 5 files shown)


lib/absinthe/schema.ex
```diff
@@ -534,9 +534,8 @@ defmodule Absinthe.Schema do
   """
   @spec used_types(t) :: [Type.t()]
   def used_types(schema) do
```
benwilson512: We should maybe rename this to `referenced_types` and deprecate this function?

binaryseed: Would you want to hard-deprecate or just log a warning?

benwilson512: I think we `@deprecate` it, so that it gets a compile time warning.

binaryseed: done!

PR #859, opened 2020-01-30
Hydrate enum values dynamically

### LINKED ISSUE #843: 1.5.0-rc.2: Calling functions in enum value 'as:' option results in compile error.
### Environment

* Elixir version (elixir -v): Elixir 1.9.4 (OTP 22)
* Absinthe version (mix deps | grep absinthe):  locked at 1.5.0-rc.2 (absinthe) 132ceb90

### Expected behavior

Expected successful compilation, with enum values defined by function calls. (Worked in 1.4)

### Actual behavior

```
== Compilation error in file lib/app_web/schemas/animal.ex ==                                                                                                                                                             
** (ArgumentError) you attempted to apply :aliases on [as: {{:., [line: 9], [{:__aliases__, [counter: {AppWeb.Schemas.Animal, 5}, line: 9], [:Animal]}, :female]}, [line: 9], []}]. If you are using apply/3, make sure the 
module is an atom. If you are using the dot syntax, such as map.field or module.function, make sure the left side of the dot is an atom or a map                                                                                              
    :erlang.apply([as: {{:., [line: 9], [{:__aliases__, [counter: {AppWeb.Schemas.Animal, 5}, line: 9], [:Animal]}, :female]}, [line: 9], []}], :aliases, [])                                                               
    (elixir) lib/macro.ex:1217: Macro.do_expand_once/2                                                                                                                                                                                        
    (elixir) lib/macro.ex:1308: Macro.do_expand_once/2                                                                                                                                                                                        
    (elixir) lib/macro.ex:1379: Macro.expand_until/2                                                                                                                                                                                          
    (elixir) lib/macro.ex:300: anonymous fn/3 in Macro.prewalk/2                                                                                                                                                                              
    (elixir) lib/macro.ex:270: Macro.do_traverse/4                                                                                                                                                                                            
    (elixir) lib/enum.ex:1440: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3                                                                                                                                                                     
    (elixir) lib/macro.ex:276: Macro.do_traverse/4
```

```elixir
  # ...
  alias App.Context.Animal
  # ...

  enum :animal_sex do
    value :female, as: Animal.female()
    value :male, as: Animal.male()
  end
```

binaryseed:
* raise when calling a function in enum value `:as`
* enable hydrating an enum value to configure it directly

@benwilson512 open to feedback on the approach for this one...

closes #843

### DIFF
--- a/test/absinthe/schema/hydrate_dynamic_values.exs
+++ b/test/absinthe/schema/hydrate_dynamic_values.exs
@@ -1,0 +1,56 @@
+defmodule HydrateDynamicValuesTest do
+  use ExUnit.Case, async: true
+
+  defmodule Schema do
+    use Absinthe.Schema
+
+    enum :color do
+      value :red
+      value :blue
+      value :green
+    end
+
+    query do
+      field :all, list_of(:color) do
+        resolve fn _, _, _ -> {:ok, [1, 2, 3]} end
+      end
+    end
+
+    def hydrate(%Absinthe.Blueprint.Schema.EnumValueDefinition{identifier: identifier}, [])
+        when identifier in [:red, :blue, :green] do
+      {:as, color_map(identifier)}
+    end
+
+    def hydrate(_, _) do
+      []
+    end
+
+    defp color_map(:red), do: 1
+    defp color_map(:blue), do: 2
+    defp color_map(:green), do: 3
+  end
+
+  test "can hydrate enum values dynamically" do
+    assert {:ok, %{data: %{"all" => ["RED", "BLUE", "GREEN"]}}} == Absinthe.run("{ all }", Schema)
+  end
+
+  test "can't call functions to configure enum values dynamically" do
+    schema = """
+    defmodule KeywordExtend do
+      use Absinthe.Schema
+
+      enum :color do
+        value :red
+        value :blue, as: color_map(:blue)
+        value :green
+      end
+    end
+    """
+
+    error = ~r/Invalid Enum value/
+
+    assert_raise(Absinthe.Schema.Notation.Error, error, fn ->
+      Code.eval_string(schema)
+    end)
+  end
+end
--- a/lib/absinthe/phase/schema/hydrate.ex
+++ b/lib/absinthe/phase/schema/hydrate.ex
@@ -66,11 +66,18 @@ defmodule Absinthe.Phase.Schema.Hydrate do
 
   @impl Absinthe.Schema.Hydrator
 
-  def apply_hydration(node, {:meta, keyword_list}) when is_list(keyword_list) do
+  def apply_hydration(
+        node,
+        {:meta, keyword_list}
+      )
+      when is_list(keyword_list) do
     %{node | __private__: Keyword.put(node.__private__, :meta, keyword_list)}
   end
 
-  def apply_hydration(node, {:description, text}) do
+  def apply_hydration(
+        node,
+        {:description, text}
+      ) do
     %{node | description: text}
   end
 
@@ -126,6 +133,13 @@ defmodule Absinthe.Phase.Schema.Hydrate do
       )
       when is_function(is_type_of) do
     %{node | is_type_of: is_type_of}
+  end
+
+  def apply_hydration(
+        %Blueprint.Schema.EnumValueDefinition{} = node,
+        {:as, value}
+      ) do
+    %{node | value: value}
   end
 
   @hydration_level1 [
--- a/lib/absinthe/schema/notation.ex
+++ b/lib/absinthe/schema/notation.ex
@@ -1366,10 +1366,21 @@ defmodule Absinthe.Schema.Notation do
   end
 
   def handle_enum_value_attrs(identifier, raw_attrs) do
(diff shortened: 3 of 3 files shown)


binaryseed:
@benwilson512 Last few issues blocking next 1.5 rc...

binaryseed:
Yeah good point, this just lets you edit an existing `value` - Perhaps I should just put a `:value` hydration on the enum instead, and folks can define any programatic values that way

binaryseed:
@benwilson512 do you have any thoughts on this?

It seems like most other hydration takes pretty basic data types (strings, functions, etc), whereas setting the full `values` involves setting up the correct blueprint structs, which is quite a bit more complex...

benwilson512:
That's a good point. I think it's fair to say that if you want to mess with the actual type structs then you should use a pipeline modifier to add a new schema phase. We definitely could add some docs around that to make sure it feels intuitive.

I think we can merge what is here as in this case, and promote general schema manipulation strategies as the way to add whole values.

test/absinthe/schema/hydrate_dynamic_values.exs
```diff
@@ -0,0 +1,56 @@
...
+
+    query do
+      field :all, list_of(:color) do
+        resolve fn _, _, _ -> {:ok, [1, 2, 3]} end
+      end
+    end
+
+    def hydrate(%Absinthe.Blueprint.Schema.EnumValueDefinition{identifier: identifier}, [])
+        when identifier in [:red, :blue, :green] do
+      {:as, color_map(identifier)}
```
bgentry: Is it correct to use the `:identifier` key here, or should we instead be using `:value`? For example, I have this enum definition:

```elixir
enum :purchase_payment_status do
  value(:paid, as: :purchase_payment_status_paid)
end
```

In that example, `:paid` corresponds to the external value of the enum, `"PAID"`. With my chosen value `as: :purchase_payment_status_paid`, I assumed I was supposed to provide something totally unique that I could reference during hydration. However, when `hydrate` gets called, it gets called with:

```elixir
[2] %Elixir.Absinthe.Blueprint.Schema.EnumValueDefinition{
    __reference__: %{
      location: %{
        file: "/MyApp/web/graphql/schema_types/purchase.ex"
        line: 12
      }
      module: :Elixir.MyApp.GraphQL.Schema.Types.Purchase
    }
    deprecation: nil
    description: "The purchase has not yet been paid"
    directives: []
    errors: []
    flags: %{}
    identifier: :not_paid
    module: :Elixir.MyApp.GraphQL.Schema.Types.Purchase
    name: "NOT_PAID"
    source_location: nil
    value: :purchase_payment_status_not_paid
  }
```

In that example it is `:value` that I actually want to match against in order to convert to its true internal value, `"Not Paid"`, which I pull in from a function in another module during `hydrate`:

```elixir
def hydrate(%Absinthe.Blueprint.Schema.EnumValueDefinition{value: value} = enum, [])
    when value in [:purchase_payment_status_not_paid] do
  {:as, MyApp.GraphQL.Schema.Types.Purchase.payment_status(value)}
end
```

Asking here for a few reasons:

1. In case we might want to alter the test to reflect different recommended usage
2. In case we want to alter any docs to reflect this
3. In case somebody else runs into a similar issue and this example is helpful to them.

binaryseed: It's just an example, people could do it whatever way works best for them...

Though the whole point is to assign the `:as` value, so already having one set will just overwrite it (ie: why set that at all in the schema?)

Seems like you have 3 "layers" of enum values going on here, which I don't think would be so common?

bgentry: Yeah, I can see how it might be less common. In this case, the external value `:paid`/"PAID"` is not something that I would be confident would be unique across _all_ of our enums, so I don't feel confident matching on that in `hydrate` because it might conflict with another value (say the same value on Order status, or Invoice status).

I haven't seen any GraphQL convention that advocates making all of your enums have globally unique values, so this seems like what's necessary to avoid a potential collision 🤷‍♂ Anyway, will assume this is a rarer use case and leave it at that, thanks @binaryseed :v:

benwilson512: >I haven't seen any GraphQL convention that advocates making all of your enums have globally unique values, so this seems like what's necessary to avoid a potential collision 

The normal solution here is that the `hydrate` function should be passed a list as the second arg of the parent values indicating where in the schema "tree" we are. This would allow you to pattern match for a particular enum value definition found under a specific enum definition. However @binaryseed I'm noticing that the second arg in your test case is an empty list. Am if misremembering how hydrate works or are we losing context for some reason?

binaryseed: yeah I noticed that too - it works currently, but i'm making a PR that matches on the ancestors since thats what we should be doing

binaryseed: https://github.com/absinthe-graphql/absinthe/pull/871

PR #862, opened 2020-01-31
Update Endpoint config for Subscriptions in 1.5

MikaAK:
With current config which was for (1.4) the following occurs in `1.5.0-rc.2`:

```elixir
 ** (EXIT) an exception was raised:
        ** (FunctionClauseError) no function clause matching in :elixir_aliases.do_concat/2
            (elixir) src/elixir_aliases.erl:130: :elixir_aliases.do_concat([[MyApp.Endpoint], :Registry], "Elixir")
            (elixir) src/elixir_aliases.erl:118: :elixir_aliases.concat/1
            (absinthe) lib/absinthe/subscription/supervisor.ex:11: Absinthe.Subscription.Supervisor.init/1
```

### DIFF
--- a/guides/subscriptions.md
+++ b/guides/subscriptions.md
@@ -31,7 +31,7 @@ line:
 [
   # other children ...
   supervisor(MyAppWeb.Endpoint, []), # this line should already exist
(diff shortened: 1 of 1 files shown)


binaryseed:
Thanks!

PR #863, opened 2020-02-03
Add missing type to the dataloader options

IvanIvanoff:
The `:callback` key in the options list is used here: https://github.com/absinthe-graphql/absinthe/blob/396e6c99b9fc2e26c683be840d1e5deea35adcea/lib/absinthe/resolution/helpers.ex#L292 and is described in the documentation, too.

This key is missing from the type specification, so using it causes dialyzer warnings.


benwilson512:
Thanks!

PR #867, opened 2020-02-14
Handle empty types in SDL renderer

bernardd:
The (excellent) new SDL renderer runs into trouble when there's an empty type in the GraphQL schema. As far as I can tell from reading the SDL docs, it's not actually legal to have an empty type (which it is in GraphQL, I think) so it's understandable that it wouldn't be happy. The failure mode, however, is extremely opaque (dumping a function clause error into the output file with the whole blueprint attached).

This change allows the SDL renderer to produce empty objects, even though they're probably technically not valid SDL (since, valid or not, it's an extremely useful tool for producing pretty-printed schemas for us). I'm happy to gate this functionality behind a flag (say `--allow-empty-types`) or something so that the default behaviour still only produces valid output if you think that's a good idea. I left it out for now because I didn't want to over-complicate things unnecessarily.

Perhaps @binaryseed would like to weigh in as well, since he wrote the functionality in the first place (thanks!).

### DIFF
--- a/lib/absinthe/schema/notation/sdl_render.ex
+++ b/lib/absinthe/schema/notation/sdl_render.ex
@@ -243,10 +243,12 @@ defmodule Absinthe.Schema.Notation.SDL.Render do
         false -> [seperator]
       end
 
+    start = empty()
+
     items
     |> Enum.reverse()
-    |> Enum.reduce(:start, fn
-      item, :start -> render(item, type_definitions)
+    |> Enum.reduce(start, fn
+      item, ^start -> render(item, type_definitions)
       item, acc -> concat([render(item, type_definitions)] ++ splitter ++ [acc])
     end)
   end
--- a/test/mix/tasks/absinthe.schema.sdl_test.exs
+++ b/test/mix/tasks/absinthe.schema.sdl_test.exs
@@ -20,6 +20,23 @@ defmodule Mix.Tasks.Absinthe.Schema.SdlTest do
 
   @test_schema "Mix.Tasks.Absinthe.Schema.SdlTest.TestSchema"
 
+  defmodule TestSchemaWithEmpty do
+    use Absinthe.Schema
+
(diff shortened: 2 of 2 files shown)


binaryseed:
I'm not sure about this.. It's definately invalid GraphQL (not just SDL):

* http://spec.graphql.org/draft/#sec-Objects.Type-Validation

@bruce @benwilson512 Either of you know why Absinthe doesn't blow up when it finds an empty `object` type?

Can you explain your use case? I'm unclear how rendering an invalid schema helps...

bernardd:
> I'm not sure about this.. It's definately invalid GraphQL (not just SDL):

Right you are. My fault for not actually reading the spec and just assuming I knew what I was talking about :)

We don't actually have any empty objects in our schema. The actual issue I'm striking (before I simplified it past the point of usefulness) is triggered by our use of https://github.com/absinthe-graphql/absinthe_relay. As best as I can tell, the blueprint passed to the renderer has not yet expanded out the `edges` fields to include `node` and `cursor`. Exactly how/when that's meant to happen is black magic to me at this point. I started doing a bit of digging, but it quickly got well past my expertise. Regardless, the upshot is that the renderer gets a blueprint that as far as it's concerned contains an empty object type for `edges`, which is what led me here.

So my best guess now is that the issue lies somewhere in the interaction between AbsintheRelay and the SDL renderer. Whether it's that the former does something illegal that still works, or that the latter doesn't handle some rarely used absinthe blueprint magic is beyond my ability to guess right now :)

benwilson512:
Ah, quite possibly the SDL renderer isn't running pipeline modifiers, so the Relay schema phases aren't being injected properly.

bernardd:
Ah, so the "blueprint magic" one :) Thanks, I'll take a look and see what I can figure out.

bernardd:
Closing this one in favour of #868 - thanks for the help.

PR #868, opened 2020-02-17
Run pipeline modifier before rendering to SDL

bernardd:
This replaces #867 - see the discussion over there for the backstory :)

The SDL renderer was not adding pipeline modifiers before rendering - this should fix it.


binaryseed:
Thanks!

PR #871, opened 2020-02-20
Clarify hydrate enum test

binaryseed:
This provides a better example of hydrating an enum value by matching on the ancestors more directly to target just one enum

thanks @bgentry for pointing this out in https://github.com/absinthe-graphql/absinthe/pull/859/files#r382307071


bgentry:
This helps a bunch, thanks all :v:

PR #874, opened 2020-02-28
Input type validation bug

binaryseed:
Fix a bug in `Validation.VariablesAreInputTypes` that happens when an unknown type is provided as a variable type. 

If a variable is defined "wrapped" - like a non-null, then we pass `%Absinthe.Type.NonNull{of_type: nil}` to `Type.name` and it eventually gets down to the `nil` and blows up.

### DIFF
--- a/lib/absinthe/phase/document/validation/variables_are_input_types.ex
+++ b/lib/absinthe/phase/document/validation/variables_are_input_types.ex
@@ -20,19 +20,22 @@ defmodule Absinthe.Phase.Document.Validation.VariablesAreInputTypes do
 
   # Find variable definitions
   @spec handle_node(Blueprint.node_t(), Schema.t()) :: Blueprint.node_t()
-  defp handle_node(%Blueprint.Document.VariableDefinition{schema_node: nil} = node, _) do
-    node
-  end
-
   defp handle_node(%Blueprint.Document.VariableDefinition{} = node, schema) do
-    type = Schema.lookup_type(schema, node.schema_node)
+    schema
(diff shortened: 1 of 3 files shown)


binaryseed:
ok, dialyzer is failing me.

it doesn't like `Type.unwrap(nil)` but the code is perfectly fine with it...

lib/absinthe/phase/document/validation/variables_are_input_types.ex
```diff
@@ -20,19 +20,22 @@ defmodule Absinthe.Phase.Document.Validation.VariablesAreInputTypes do
 
   # Find variable definitions
   @spec handle_node(Blueprint.node_t(), Schema.t()) :: Blueprint.node_t()
-  defp handle_node(%Blueprint.Document.VariableDefinition{schema_node: nil} = node, _) do
```
binaryseed: The problem with this is that `schema_node` can be "wrapped" like `%Absinthe.Type.NonNull{of_type: nil}` so it doesn't actually catch all the cases when there is an un-known type

binaryseed:
@benwilson512 could you take a peek at this?

lib/absinthe/phase/document/validation/variables_are_input_types.ex
bruce: Since, as you say, a `nil` `schema_node` can be "wrapped," it seems to make sense to update the `@spec` for `Type.unwrap/1` to add `nil` as a possible return value.

I've added a commit to that end (passes locally).

binaryseed:
Great, I'll merge this then!

PR #878, opened 2020-03-09
Do not allow nested non_null

### LINKED ISSUE #877: Nested `non_null` not caught by schema validator
### Environment

* Elixir version (elixir -v): 1.9.4
* Absinthe version (mix deps | grep absinthe): 1.5.0-rc.3
* Client Framework and version (Relay, Apollo, etc): Apollo

### Expected behavior

The schema validator should catch nested `non_null` field definitions at compile time.

### Actual behavior

It doesn't actually do the above.

Sample field definition that highlights this unexpected behaviour.

```elixir
field(:permissions, non_null(list_of(non_null(non_null(:string)))))
```

I can create a PR for this provided someone can give me pointers on where to start.

binaryseed:
Do not allow nested `non_null`, as that doesn't make sense, and is not allowed in the spec:

http://spec.graphql.org/draft/#sec-Type-System.Non-Null.Type-Validation

Fixes #877

### DIFF
--- a/lib/absinthe/schema/notation.ex
+++ b/lib/absinthe/schema/notation.ex
@@ -1081,6 +1081,12 @@ defmodule Absinthe.Schema.Notation do
 
   See `field/3` for examples
   """
+
+  defmacro non_null({:non_null, _, _}) do
+    raise Absinthe.Schema.Notation.Error,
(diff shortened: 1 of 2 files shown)


binaryseed:
This is already prevented in SDL schemas since it's a syntax error

Issue #700, opened 2019-03-30
Non-UTC datetimes are considered invalid


1player:
The DateTime parser in Absinthe.Type.Custom:
https://github.com/absinthe-graphql/absinthe/blob/cf46de10c10e155b413e8609d49e32da2999a839/lib/absinthe/type/custom.ex#L77-L78 returns `:error` if a non-UTC datetime (but still with timezone information) is passed to it.

Ex: `2019-03-30T17:19:00Z` is valid, `2019-03-30T17:19:00+01:00` is not.

What is the rationale behind that design choice? `DateTime.from_iso8601` converts the date to UTC in the second element of the return tuple, so non-UTC date strings are in theory parseable.

```
iex(1)> DateTime.from_iso8601("2019-03-30T17:19:00Z")
{:ok, #DateTime<2019-03-30T17:19:00Z>, 0}
iex(2)> DateTime.from_iso8601("2019-03-30T17:19:00+01:00")
{:ok, #DateTime<2019-03-30 16:19:00Z>, 3600}
```

In my project I've copy-pasted Absinthe.Type.Custom to add support for non-UTC datetimes, I'd like to know if I'm missing something and I should enforce UTC datetimes on my clients instead.

conradwt:
`DateTime.from_iso8601` is a function which comes from the Elixir project.  Thus, you may want to file an issue there.

1player:
There is no issue on DateTime.from_iso8601. The function works as intended.

My issue is related to the DateTime type parser in Absinthe.Type.Custom, which does not accept non-UTC date strings.

benwilson512:
@1player What value should the scalar return for non utc date times? Keep in mind, an offset is not the same thing as a timezone.

1player:
> What value should the scalar return for non utc date times? Keep in mind, an offset is not the same thing as a timezone.

I reckon a valid DateTime object. The common best practice is converting to UTC at the edges of the application, and dealing only with UTC objects internally, which is what my application does.

The only piece of information you'd lose is the original time offset, which I do not care about, and I doubt many applications require to know that information. My frontend client when it's doing a request encodes a JS Date object into a ISO8601 string, which is not always in UTC timezone. I would expect my GraphQL backend to be able to handle those.

[...]

PR #879, opened 2020-03-11
Allowing valid ISO8601 DateTimes to be converted to UTC

seivan:
Fixing: https://github.com/absinthe-graphql/absinthe/issues/700

Updated the wording on the header doc, but not sure if it's good enough. 
Sorry for the delay.


benwilson512:
thanks!

PR #889, opened 2020-03-26
Correct spec for Pipeline.replace

binaryseed:
I _think_ this is the correct spec - the third argument is a phase, not a pipeline (the `t` is `Pipeline.t` here)

### DIFF
--- a/lib/absinthe/pipeline.ex
+++ b/lib/absinthe/pipeline.ex
@@ -227,7 +227,7 @@ defmodule Absinthe.Pipeline do
       [A, {X, [name: "Nope"]}, C]
 
   """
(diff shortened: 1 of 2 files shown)


binaryseed:
one more dialyzer error to figure out:

```
lib/absinthe/subscription/local.ex:42:call
The function call will not succeed.
Absinthe.Pipeline.replace(
  any(),
  Absinthe.Phase.Telemetry,
  {Absinthe.Phase.Telemetry, [:publish | :start | :subscription, ...]}
)
breaks the contract
(t(), Absinthe.Phase.t(), phase_config_t()) :: t()
```

PR #890, opened 2020-03-27
further-dialyzer-cleanup

benwilson512:

### DIFF
--- a/lib/absinthe/phase/subscription/subscribe_self.ex
+++ b/lib/absinthe/phase/subscription/subscribe_self.ex
@@ -31,7 +31,7 @@ defmodule Absinthe.Phase.Subscription.SubscribeSelf do
       {:replace, blueprint,
        [
          {Phase.Subscription.Result, topic: subscription_id},
-         {Phase.Telemetry, [:execute, :operation, :stop, options]}
+         {Phase.Telemetry, Keyword.put(options, :event, [:execute, :operation, :stop])}
        ]}
     else
       {:error, error} ->
--- a/lib/absinthe/phase/telemetry.ex
+++ b/lib/absinthe/phase/telemetry.ex
@@ -10,12 +10,21 @@ defmodule Absinthe.Phase.Telemetry do
 
   use Absinthe.Phase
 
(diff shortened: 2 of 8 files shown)


lib/absinthe/phase/subscription/subscribe_self.ex
```diff
@@ -31,7 +31,7 @@ defmodule Absinthe.Phase.Subscription.SubscribeSelf do
       {:replace, blueprint,
        [
          {Phase.Subscription.Result, topic: subscription_id},
-         {Phase.Telemetry, [:execute, :operation, :stop, options]}
+         {Phase.Telemetry, Keyword.put(options, :event, [:execute, :operation, :stop])}
```
benwilson512: @binaryseed note this change. This is partially to appease Dialyzer since the spec for phase config says that these are keywords. It's also handy because it allows us to more consistently push out telemetry events with this info.

lib/absinthe/phase/telemetry.ex
```diff
@@ -10,12 +10,21 @@ defmodule Absinthe.Phase.Telemetry do
...
+  defp do_run(blueprint, [:execute, :operation, :start], options) do
     id = :erlang.unique_integer()
     start_time = System.system_time()
     start_time_mono = System.monotonic_time()
 
-    :telemetry.execute(@operation_start, %{start_time: start_time}, %{id: id})
+    :telemetry.execute(@operation_start, %{start_time: start_time}, %{
+      id: id,
+      blueprint: blueprint,
+      options: options
```
benwilson512: @binaryseed can you think of a reason to not push out the blueprint and options on all of these?

lib/absinthe/phase/subscription/subscribe_self.ex
binaryseed: makes sense!

lib/absinthe/phase/telemetry.ex
binaryseed: probably good to be consistent.  the only worry is the size of the struct, hopefully no handler passes everything across a process boundary

PR #891, opened 2020-03-27
return dataloader/3 use_parent to default false

bgentry:
In #882, some changes were made which not only fixed the behavior of `use_parent` when explicitly specified, but also changed the default value from `false` to `true`. While this was always documented as `true`, that wasn't actually the behavior until that PR. This is a breaking change which can result in some surprising and dangerous issues for users that were relying on the pre-existing behavior.

Rather than break those existing apps, and because defaulting to `false` is safer and less surprising, we return the default to `false` and correct the documentation to reflect how it has always worked. This was the conclusion of @benwilson512 in https://github.com/absinthe-graphql/absinthe/pull/882#issuecomment-605324628.

Side note: I think this was the last remaining issue for me getting our app on v1.5 🤞


benwilson512:
Thanks for raising this again!

PR #893, opened 2020-03-28
don't pass introspection fields or objects to the def middleware callback

### LINKED ISSUE #565: Schema middleware callback affecting introspection
See https://github.com/absinthe-graphql/absinthe_plug/issues/159#issuecomment-399488012

benwilson512:
Fixes #565
Fixes https://github.com/absinthe-graphql/absinthe_plug/issues/159

### DIFF
--- a/lib/absinthe/blueprint/schema/object_type_definition.ex
+++ b/lib/absinthe/blueprint/schema/object_type_definition.ex
@@ -49,7 +49,8 @@ defmodule Absinthe.Blueprint.Schema.ObjectTypeDefinition do
       fields: build_fields(type_def, schema),
       interfaces: type_def.interfaces,
       definition: type_def.module,
-      is_type_of: type_def.is_type_of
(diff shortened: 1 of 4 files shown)


binaryseed:
What's the impact of this - presumably some APIs out there were installing middleware on these fields, right?

benwilson512:
Probably not, they weren't available for this until one of the more recent 1.5 releases. I'll confirm that though.

binaryseed:
Oh great, that’s fine with me then.

binaryseed:
When does this expand function get called?

benwilson512:
It gets called at compile time in the InlineFunctions phase. If the result of `expand` can be safely macro escaped then that field's compile time struct gets populated with the middleware list. If it can't be macro escaped, then it happens when the type is loaded at runtime.

halostatue:
This should also close absinthe-graphql/absinthe_plug#159

Issue #892, opened 2020-03-28
Multiple telemetry events when using dataloader


akoutmos:
### Environment

* Elixir version (elixir -v): 1.10.2
* Absinthe version (mix deps | grep absinthe): 1.5.0-rc.4
* Client Framework and version (Relay, Apollo, etc): Apollo

### Expected behavior

When using a dataloader for a field, I would expect a single `[:absinthe, :resolve, :field, :start]` event and a single `[:absinthe, :resolve, :field, :stop]` event to be emitted via telemetry during the resolution of a field.

### Actual behavior

I am currently getting 1 pair of start/stop telemetry events per reference of that field in a list of items even though I am only reaching out once to fetch the batch data.

I'll try and throw together a dummy project to repro the issue more clearly, but for now any assistance would be greatly appreciated. Thanks for all the hard work you've put into Absinthe!! Appreciate it :).

akoutmos:
After digging through the code a bit, perhaps separate dataloader start/stop events should be emitted here: https://github.com/absinthe-graphql/dataloader/blob/master/lib/dataloader/kv.ex#L107. Thoughts?

benwilson512:
Hi @akoutmos, the behaviour you are seeing is correct and expected. Batch style resolvers like dataloader group the resolution of many fields together.

Suppose you have this:
```
{
  posts {
    author { name
  }
}
```
and there are 3 posts, the `author` field uses dataloader. Each posts' author field starts resolving, but then suspends to wait on the results from dataloader. This means you get 3 start event in a row as each field adds its work to the dataloader batch. Then dataloader runs, and then each result is fed back into each field and you get 3 stop events.

Dataloader should definitely emit telemetry events though. Please feel free to open up an issue about that there.

akoutmos:
Thanks for the detailed answer @benwilson512! That all makes sense. I'll get to work on adding telemetry to dataloader and will open up a PR. As a side note (mostly since I am new to Absinthe), how would I go about filtering out `[:absinthe, :resolve, :field, :start]` and [:absinthe, :resolve, :field, :stop]` events to avoid misreporting metrics? Is there anything in the `Absinthe.Resolution` struct in the `metadata` that could hint to me that a dataloader request can be ignored?

benwilson512:
Can you elaborate on why you want to filter those out? The times are 100% valid, they're just happening concurrently.

akoutmos:
My bad. I should have given some more context.

[...]

PR #894, opened 2020-03-29
Plugin telemetry events

akoutmos:
This PR addresses some of my thoughts that I raised in #892 and emits start+stop events for dataloader events where there are `pending_batches` to process. Those two additional Telemetry events are:
`[:absinthe, :dataloader, :resolve, :start]` when a dataloader with pending batches starts
`[:absinthe, :dataloader, :resolve, :stop]` when a dataloader with pending batches finishes

This is just an initial stab at the problem, so feel free to comment. Thanks :)!

### DIFF
--- a/lib/absinthe/middleware/dataloader.ex
+++ b/lib/absinthe/middleware/dataloader.ex
@@ -3,43 +3,14 @@ if Code.ensure_loaded?(Dataloader) do
     @behaviour Absinthe.Middleware
     @behaviour Absinthe.Plugin
 
-    @dataloader_start_event [:absinthe, :dataloader, :resolve, :start]
-    @dataloader_stop_event [:absinthe, :dataloader, :resolve, :stop]
-
     @impl Absinthe.Plugin
     def before_resolution(%{context: context} = exec) do
       context =
         with %{loader: loader} <- context do
-          start_time = System.monotonic_time()
-
-          emit_start_event(start_time, exec, loader)
-          updated_loader = Dataloader.run(loader)
-          emit_stop_event(start_time, exec, loader)
-
-          %{context | loader: updated_loader}
+          %{context | loader: Dataloader.run(loader)}
         end
 
       %{exec | context: context}
-    end
-
-    defp emit_start_event(start_time, blueprint, loader) do
-      if Dataloader.pending_batches?(loader) do
-        :telemetry.execute(
-          @dataloader_start_event,
-          %{start_time: start_time},
-          %{blueprint: blueprint}
-        )
-      end
-    end
-
-    defp emit_stop_event(start_time, blueprint, loader) do
-      if Dataloader.pending_batches?(loader) do
-        :telemetry.execute(
-          @dataloader_stop_event,
-          %{duration: System.monotonic_time() - start_time},
-          %{blueprint: blueprint}
-        )
-      end
     end
 
     @impl Absinthe.Middleware
--- a/lib/absinthe/middleware/dataloader.ex
+++ b/lib/absinthe/middleware/dataloader.ex
@@ -3,15 +3,46 @@ if Code.ensure_loaded?(Dataloader) do
     @behaviour Absinthe.Middleware
     @behaviour Absinthe.Plugin
 
+    @dataloader_start_event [:absinthe, :dataloader, :resolve, :start]
+    @dataloader_stop_event [:absinthe, :dataloader, :resolve, :stop]
+
+    @impl Absinthe.Plugin
     def before_resolution(%{context: context} = exec) do
       context =
         with %{loader: loader} <- context do
-          %{context | loader: Dataloader.run(loader)}
+          start_time = System.monotonic_time()
+
+          emit_start_event(start_time, exec, loader)
+          updated_loader = Dataloader.run(loader)
+          emit_stop_event(start_time, exec, loader)
+
+          %{context | loader: updated_loader}
         end
 
       %{exec | context: context}
     end
 
+    defp emit_start_event(start_time, blueprint, loader) do
+      if Dataloader.pending_batches?(loader) do
+        :telemetry.execute(
+          @dataloader_start_event,
+          %{start_time: start_time},
+          %{blueprint: blueprint}
+        )
+      end
+    end
+
+    defp emit_stop_event(start_time, blueprint, loader) do
+      if Dataloader.pending_batches?(loader) do
+        :telemetry.execute(
+          @dataloader_stop_event,
+          %{duration: System.monotonic_time() - start_time},
+          %{blueprint: blueprint}
+        )
+      end
+    end
+
+    @impl Absinthe.Middleware
     def call(%{state: :unresolved} = resolution, {loader, callback}) do
       if !Dataloader.pending_batches?(loader) do
         resolution.context.loader
@@ -36,10 +67,12 @@ if Code.ensure_loaded?(Dataloader) do
       Absinthe.Resolution.put_result(resolution, value)
     end
 
+    @impl Absinthe.Plugin
     def after_resolution(exec) do
       exec
     end
 
+    @impl Absinthe.Plugin
     def pipeline(pipeline, exec) do
       with %{loader: loader} <- exec.context,
            true <- Dataloader.pending_batches?(loader) do
--- a/lib/absinthe/phase/document/execution/resolution.ex
+++ b/lib/absinthe/phase/document/execution/resolution.ex
@@ -46,7 +46,7 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
     plugins = bp_root.schema.plugins()
     run_callbacks? = Keyword.get(options, :plugin_callbacks, true)
 
-    exec = run_callbacks(plugins, :before_resolution, exec, run_callbacks?)
+    exec = plugins |> run_callbacks(:before_resolution, exec, run_callbacks?)
 
     common =
       Map.take(exec, [:adapter, :context, :acc, :root_value, :schema, :fragments, :fields_cache])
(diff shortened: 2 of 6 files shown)


benwilson512:
Hey @akoutmos thanks for taking a stab at this!

Here's the challenge we need to solve: Dataloader is an _optional_ dependency within Absinthe. I don't think it's appropriate for Absinthe to emit dataloader specific events. I _do_ think that the resolution phase should emit events for when a plugin callback has started, and when a plugin callback has ended though, and the plugin module would be part of the event emitted.

For Dataloader specific events, I think we need to have those fire within Dataloader itself.

lib/absinthe/middleware/dataloader.ex
```diff
@@ -3,15 +3,46 @@ if Code.ensure_loaded?(Dataloader) do
...
+    @impl Absinthe.Plugin
     def before_resolution(%{context: context} = exec) do
       context =
         with %{loader: loader} <- context do
-          %{context | loader: Dataloader.run(loader)}
+          start_time = System.monotonic_time()
+
+          emit_start_event(start_time, exec, loader)
+          updated_loader = Dataloader.run(loader)
+          emit_stop_event(start_time, exec, loader)
```
benwilson512: Specifically, I think we should move these start and end events into https://github.com/absinthe-graphql/absinthe/blob/master/lib/absinthe/phase/document/execution/resolution.ex#L78 and it would be something like:

```
  defp run_callbacks(plugins, callback, acc, true) do
    Enum.reduce(plugins, acc, fn plugin, acc ->
      start_time = System.monotonic_time()
              :telemetry.execute(
          [:absinthe, :plugin, callback, :start]
          %{start_time: start_time},
          %{plugin: plugin, acc: acc}
        )
      acc = apply(plugin, callback, [acc])
      # stop event
     acc
    end)
  end
```

akoutmos:
@benwilson512 I think my latest commits address the comments that you brought up. I really like this approach compared to my original solution as it is more general purpose now and covers all the plugins. On that same note, is a dataloader specific event even needed? You can match on 
```elixir
{
  :telemetry_event, 
  [:absinthe, :plugin, :callback, :start], 
  %{start_time: _}, 
  %{loader:Absinthe.Middleware.Dataloader}
}
```
to get your dataloader (or any plugin for that matter) specific events. Thoughts?

lib/absinthe/phase/document/execution/resolution.ex
```diff
@@ -75,11 +78,35 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
   end
 
   defp run_callbacks(plugins, callback, acc, true) do
-    Enum.reduce(plugins, acc, &apply(&1, callback, [&2]))
+    Enum.reduce(plugins, acc, fn plugin, acc ->
+      start_time = System.monotonic_time()
+
+      emit_start_event(start_time, acc, plugin)
```
benwilson512: Very close! So, `callback` here is either the atom `:before_resolution` or `:after_resolution`, and I was thinking we could use that to dispatch more finely tuned events. Specifically this would get us:

```
[:absinthe, :plugin, :before_resolution, :start]
[:absinthe, :plugin, :before_resolution, :stop]
[:absinthe, :plugin, :after_resolution, :start]
[:absinthe, :plugin, :after_resolution, :stop]
```

lib/absinthe/phase/document/execution/resolution.ex
```diff
@@ -75,11 +78,35 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
...
+      %{start_time: start_time},
+      %{acc: acc, loader: loader}
+    )
+  end
+
+  defp emit_stop_event(start_time, acc, loader) do
+    :telemetry.execute(
+      @callback_stop_event,
+      %{duration: System.monotonic_time() - start_time},
+      %{acc: acc, loader: loader}
```
benwilson512: The `loader` variable here is I think the `plugin` right? We probably want to call it `plugin: plugin`

akoutmos:
My bad. I read your callback line as an atom and not the function variable. Third time is a charm I think I addressed all your feedback :D. Thoughts on the separate dataloader telemetry events? Thanks again!!

lib/absinthe/phase/document/execution/resolution.ex
```diff
@@ -75,11 +75,35 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
...
+      acc
+    end)
   end
 
   defp run_callbacks(_, _, acc, _), do: acc
 
+  defp emit_start_event(start_time, acc, callback, plugin) do
+    :telemetry.execute(
+      [:absinthe, :plugin, callback, :start],
+      %{start_time: start_time},
```
binaryseed: For this to be a useful measurement, it needs to be a `system_time()`... `monotonic_time` is still needed for duration tracking, but it doesn't actually mark a point in time

test/absinthe/middleware/dataloader_test.exs
```diff
@@ -158,11 +159,57 @@ defmodule Absinthe.Middleware.DataloaderTest do
...
     assert {:ok, %{data: data}} = Absinthe.run(doc, DefaultSchema)
     assert expected_data == data
 
     assert_receive(:loading)
     refute_receive(:loading)
+
+    # This test emits a total of 24 events, but there we are only validating the
+    # dataloader events in this test
+    results =
+      Enum.reduce(1..24, %{dataloader_starts: 0, dataloader_stops: 0}, fn _, acc ->
```
binaryseed: I'm unclear why the `reduce` is needed, the other telemetry tests managed to work just fine with `assert_receive`...

https://github.com/absinthe-graphql/absinthe/blob/7821d278b8c511fbae752f5f68bd94195a9c2607/test/absinthe/integration/execution/telemetry_test.exs#L62

test/absinthe/middleware/dataloader_test.exs
```diff
@@ -158,11 +159,57 @@ defmodule Absinthe.Middleware.DataloaderTest do
       ]
     }
 
+    # Get test PID and clear mailbox
+    self = self()
+    flush_process_mailbox()
```
binaryseed: Why do you need to do this?

binaryseed:
Also, we have a pattern of sticking an `id` on each event to correlate the start & stop...

https://github.com/absinthe-graphql/absinthe/blob/7821d278b8c511fbae752f5f68bd94195a9c2607/lib/absinthe/middleware/telemetry.ex#L12

lib/absinthe/phase/document/execution/resolution.ex
akoutmos: Updated the PR to have the `start_time` be from system_time()

test/absinthe/middleware/dataloader_test.exs
akoutmos: Fixed root cause of test flakiness by making tests non-async. Deleted this block

test/absinthe/middleware/dataloader_test.exs
akoutmos: The reason for the reduce is so that I can count only the specific events that I want to fetch. Since the telemetry events emit on all plugins, I was also getting Async and Batch events. But only wanted to assert on the Dataloader specific events.

benwilson512:
Is this ready for re-review?

akoutmos:
I think I addressed everything on my end. The only open question from what I gather is the number of events being generated...but I defer to you guys on that one.

binaryseed:
So, I looked into this locally on my API, and the problem is real.

We've got a few plugins installed for standard stuff like auth & monitoring, and this addition calls `execute` 2 times for every plugin for every phase of resolution.

That means when using stuff like `Async` / `Batch`, every new phase added by any plugin re-triggers a whole new set of telemetry events for every plugin.

For example, I tried it out with a really simple query, and there were 6 resolution phases. Each phase added 14 calls to execute (we have 7 plugins). That means a total of 84 additional calls to `telemetry.execute` - 84 more ETS lookups. And that's as a no-op, if there's a handler attached, then it's going to get called 84 times.

I'm a little worried about this... It's pretty unlikely that the raw data here is going to be leveraged, and that's a good bit of work to do for something that's rarely going to be used.

I"m wondering if directly instrumenting Dataloader would be a better way to go about this vs instrumenting every single plugin interaction?

benwilson512:
@binaryseed This is a good point, I didn't think about the fact that this will run N*M times, where N is the number of plugins, and M is the number of resolution phases that happen (roughly proportional to the depth of the GraphQL query). `:ets` is pretty fast, but I agree that there isn't a lot of point to firing events about plugins that don't execute.

Perhaps an alternative strategy would be to emit events only within the functions within the `batch` and `dataloader` plugins that actually execute user code so that you can track how long a given `batch` or `run_batch` call takes if any are queued up without firing events at all if none are present. I realize this is closer to what you had originally @akoutmos and I apologize for the back and forth here. In this proposal to be clear, we would move the event hooks to here: https://github.com/absinthe-graphql/absinthe/blob/master/lib/absinthe/middleware/batch.ex#L156 and within Dataloader itself. @binaryseed does that seem more reasonable?

akoutmos:
No worries at all @benwilson512! That sounds good to me in regards to moving the telemetry events to the plugins directly. If that sounds good to you as well @binaryseed then I can get going on this approach and update this PR + open up a new one for the dataloader repo.

binaryseed:
Instrumenting `call_batch_fun` sounds pretty good!

benwilson512:
@akoutmos do note the naming conventions mentioned here: https://github.com/absinthe-graphql/absinthe/pull/901#pullrequestreview-386405817

akoutmos:
Working on cleaning up this PR to reflect the conversation that we had. But managed to wrap up the Dataloader changes: https://github.com/absinthe-graphql/dataloader/pull/89

Will update this PR over the weekend. Thanks!

akoutmos:
I think I addressed everything that we discussed and also updated the documentation.

Question for you guys. Should I remove the telemetry tests in the `test/absinthe/middleware/dataloader_test.exs` file? The emitting of telemetry events is covered in the Dataloader PR that I made...but it would also be nice to ensure that those events are captured in an integration style test within Absinthe. On the flip side, the tests in Absinthe will fail until a newer version of Dataloader is published and the deps are updated in Absinthe. Thoughts?

binaryseed:
Yeah, I'd much rather we keep the tests for dataloader telemetry events inside the library itself...

akoutmos:
I went ahead and removed the test @binaryseed

binaryseed:
Thanks! I'm going to merge this and followup with a small commit to move the `telemetry.execute` call in the start case to execute after the task is spawned so that handlers don't impact the timing!

binaryseed:
Here it is: https://github.com/absinthe-graphql/absinthe/commit/38016db1216bcd133273dd6ff46885cb06611820

akoutmos:
Thanks @binaryseed! Let me know what you think of the dataloader changes and if you require any changes there https://github.com/absinthe-graphql/dataloader/pull/89.

PR #895, opened 2020-03-29
Persistent Term Backend

benwilson512:
NOTE: To use you need to add this worker to your supervision tree:

```
{Absinthe.Schema, MyAppWeb.Schema}
```

Where `MyAppWeb.Schema` is the module name of your schema.

# TODO
- [x] Default back to compiled backend
- [x] For Absinthe tests, enable a `SCHEMA_PROVIDER=persistent_term` env variable that we can use on travis tests utilizing OTP 22.

# Further Exploration

This enables for the first time the real possibility of schemaless modules


lib/absinthe/schema.ex
```diff
@@ -107,13 +107,7 @@ defmodule Absinthe.Schema do
...
       @before_compile unquote(__MODULE__)
 
-      defdelegate __absinthe_type__(name), to: __MODULE__.Compiled
-      defdelegate __absinthe_directive__(name), to: __MODULE__.Compiled
-      defdelegate __absinthe_types__(), to: __MODULE__.Compiled
-      defdelegate __absinthe_types__(group), to: __MODULE__.Compiled
-      defdelegate __absinthe_directives__(), to: __MODULE__.Compiled
-      defdelegate __absinthe_interface_implementors__(), to: __MODULE__.Compiled
-      defdelegate __absinthe_prototype_schema__(), to: __MODULE__.Compiled
+      @schema_provider Absinthe.Schema.PersistentTerm
```
benwilson512: This is only set this way temporarily, it should default back to the `Absinthe.Schema.Compiled` backend probably.

test/test_helper.exs
```diff
@@ -1,2 +1,22 @@
...
+  Absinthe.Fixtures.OnlyQuerySchema,
+  Absinthe.Fixtures.PetsSchema,
+  Absinthe.Fixtures.StrictSchema,
+  Absinthe.Fixtures.TimesSchema,
+  Absinthe.Fixtures.Things.SDLSchema,
+  Absinthe.Fixtures.Things.MacroSchema
+]
+
+for schema <- fixtures_schemas do
+  Absinthe.Schema.Manager.start_link(schema)
```
benwilson512: This has to be added otherwise the modules aren't actually available reliably. They _are_ available the first time they're compiled because compiling a module with the persistent term provider DOES put the schema into persistent term, this is how all of the inline schemas in the test files still work.

However for these `.ex` fixtures, after the test suite has been compiled once, re-running tests would fail because those schemas aren't being recompiled, and thus aren't going back in persistent term. Thus, we need to use the genserver.

This won't really be an issue for end users because they should always place the genserver in their supervision tree.

benwilson512:
Hm, the OTP 21 stuff is failing for some reason, dialyzer claims that `:persistent_term.put/2` doesn't exist.

benwilson512:
I think it has to do with PLT caching perhaps?

lib/absinthe/schema/persistent_term.ex
```diff
@@ -0,0 +1,81 @@
...
+
+    defp get(schema) do
+      :persistent_term.get({__MODULE__, schema})
+    end
+  end
+else
+  defmodule Absinthe.Schema.PersistentTerm do
+    @moduledoc false
+
+    @error "Can't be used without OTP >= 21"
```
binaryseed:
```suggestion
    @error "Can't be used without OTP >= 21.2"
```

benwilson512:
Dialyzer man...

binaryseed:
Google might solve this for us:

https://github.com/jeremyjh/dialyxir/issues/369#issuecomment-573088088

PR #896, opened 2020-03-30
Rename file to `hydrate_dynamic_values_test.exs`

maartenvanvliet:
Its previous name excluded the tests in the file from being run.


binaryseed:
Thanks!

PR #899, opened 2020-04-01
Update start.md

PLsergent:
Update link for Phoenix Framework documentation.


benwilson512:
Hi @PLsergent I'm happy to accept this if it is rebased against master. v1.5 is in RC, so we're only doing critical fixes for 1.4 at this time.

Issue #314, opened 2017-04-28
"did you mean" with input objects


benwilson512:
Right now invalid fields will provide "did you mean" suggests in the error, we should provide these for input object fields as well.

zacck:
Can I take a stab at this one? I think I can try this one out

zacck:
## Notes

Incorrect field names suggestions
```
 Validation.KnownArgumentNames #=> Handle misspelled args
@spec field_error(Blueprint.node_t(), Blueprint.node_t(), String.t()) :: Phase.Error.t()
  defp field_error(argument_node, field_node, type_name) do
    field_node #=> %Blueprint.Document.Field{}
    field_node.schema_node #=> %Absinthe.Type.Field{}
    field_node.schema_node.arguments
    possible_argument_name = 
    %Phase.Error{
      phase: __MODULE__,
      message: field_error_message(argument_node.name, field_node.name, type_name),
      locations: [argument_node.source_location]
    }
  end
```

Incorrect input_object suggestion 

```
 Validation.ArgumentsOfCorrectType #=> Handle misspelled input object fields

defp find_suggested_name(node, child_name) do
end 

defp collect_child_errors(%Blueprint.Input.Object{} = node, schema) do
    node.fields
    |> Enum.flat_map(fn
      %{flags: %{invalid: _}, schema_node: nil} = child ->
        suggested_name = find_suggested_name(node, child.name)
        [unknown_field_error_message(suggested_name, child.name)]
```

Ecksters:
I just want to make sure this covers an issue we saw recently, where an `object` was used for a mutation `input`, rather than an `input_object`, and no error was thrown.

benwilson512:
@Ecksters that specific issue is unrelated to this issue. This issue is about suggesting field names when they are misspelled within a query. Your issue is about checks done when a schema is built.

pmargreff:
@benwilson512 this issue can be closed!

PR #902, opened 2020-04-07
Feature/did you mean input objects

pmargreff:
It address #314. 

The steps to make things happen are: 
- Extract suggesion's to `Utils` module, so both input and objects can use the same implementation. 
- Complement unknown field error message with suggestions (when they exists). 

---

I've based the tests on current ones, feel free to request new tests where necessary to avoid any regressions. 

I also have no sure about the `Utils` module inside `/validation` dir is a good pattern, since each module inside `/validation` maps directly to a step, I've created this new `Utils` thing to differentiate it. 

Suggestions are always welcome! : )

### DIFF
--- a/lib/absinthe/phase/document/validation/arguments_of_correct_type.ex
+++ b/lib/absinthe/phase/document/validation/arguments_of_correct_type.ex
@@ -80,7 +80,7 @@ defmodule Absinthe.Phase.Document.Validation.ArgumentsOfCorrectType do
     |> Enum.flat_map(fn
       %{flags: %{invalid: _}, schema_node: nil} = child ->
         field_suggestions =
-          case node.schema_node do
+          case Type.unwrap(node.schema_node) do
             %Type.Scalar{} -> []
             %Type.Enum{} -> []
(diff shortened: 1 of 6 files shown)


lib/absinthe/phase/document/validation/arguments_of_correct_type.ex
```diff
@@ -79,7 +79,14 @@ defmodule Absinthe.Phase.Document.Validation.ArgumentsOfCorrectType do
     node.fields
     |> Enum.flat_map(fn
       %{flags: %{invalid: _}, schema_node: nil} = child ->
-        [unknown_field_error_message(child.name)]
+        field_suggestions =
+          case node.schema_node do
```
benwilson512, line 83: You should `Type.unwrap(node.schema_node)` because this could be wrapped in `non_null`.

pmargreff: Used this line as example, should I change this one also? It's not related with the new feature!

benwilson512:
Thank yoU!

PR #777, opened 2019-09-06
fix broken markdown links in guides/

lyninx:
just fixing some broken markdown links / images i stumbled across while reading through the guide!

PR #905, opened 2020-04-15
Fix broken guide links

parkerduckworth:
Hey!

I noticed that in the [tutorial guide](https://github.com/absinthe-graphql/absinthe/tree/master/guides/tutorial), the links to each next step in the tutorial are broken, as they refer to `<link-to-next-step>.html`, rather than using a `.md` extension.  This PR addresses this problem by changing each `<link-to-next-step>` extension to `.md`.

I see that PR #777 begins to address this issue, and there may be more.  I will look through when I get a chance to make sure all other guide links are updated.

### DIFF
--- a/guides/tutorial/complex-arguments.md
+++ b/guides/tutorial/complex-arguments.md
@@ -151,4 +151,4 @@ Here's our mutation in action in GraphiQL.
 
 ## Next Step
 
-Now let's [wrap things up](conclusion.html).
+Now let's [wrap things up](conclusion.md).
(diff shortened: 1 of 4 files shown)


benwilson512:
Thank you!

rafael-vitor:
Hey. This change makes it so that links are now broken when reading the external website version of the guides as they have the .html extension 
ex: https://hexdocs.pm/absinthe/start.html should link to https://hexdocs.pm/absinthe/our-first-query.html instead of https://hexdocs.pm/absinthe/our-first-query.md which doesn't exist

is there a way these paths could be mapped so the links can work on both github and the hexdocs page? or maybe a way of generating them dinamically?

parkerduckworth:
Excellent question and point. Yeah I’m going explore solutions.

PR #914, opened 2020-04-30
Docs/loading batch

pmargreff:
This pull request only introduces some improvements on docs organizations: 
- Move batching resolution details away from Ecto section.
- Move all explanation about Dataloader to Dataloader section, on batching resolution I've tried to keep it really straight. 
- Minor complementation on Dataloader advanced uses example. 
- Remove setup and explanations about `absinthe_ecto` (keeps only a reference to the project and the deprecation note). 
- Remove dead reference to `absinthe_ecto`. 
- Fix CI badge on readme :)


guides/dataloader.md
```diff
@@ -23,74 +23,87 @@ that encodes a way of retrieving data. More info in the [Sources](#sources) sect
...
-    end
-  end
-end
+```elixir
+def query(Post, %{has_admin_rights: true}), do: query
+
+def query(Post, _), do: from(Post in query, select_merge: %{views: nil})
+
+def query(queryable, _), do: queryable
 ```
```
billylanchantin, line 102: Two issues here:

1. The code here is not valid. We would want something like:

   ```elixir
   def query(Post, %{has_admin_rights: true}), do: Post

   def query(Post, _), do: from(p in Post, select_merge: %{views: nil})

   def query(queryable, _), do: queryable
   ```

2. IMO, `:select_merge` is a little niche. I've been using Ecto for a while and I had to look it up. I think for a tutorial, we might want something a little simpler. Maybe

   ```elixir
   def query(Post, %{has_admin_rights: true}), do: Post

   def query(Post, _), do: from p in Post, where: is_nil(p.deleted_at)

   def query(queryable, _), do: queryable
   ```

   Since this code block is supposed to show how you can add use `query/2` to enforce access rights, building on the previous example with a single new clause I think makes the concept clearer.

guides/batching.md
```diff
@@ -1,7 +1,201 @@
...
+
+### Dataloader
+
+Let's jump straight in to getting Dataloader working, and then we'll expand on
+what's actually happening behind the scenes.
+
+Using Dataloader is as simple as doing:
+
+```elixir
+import Absinthe.Resolution.Helpers, only: [dataloader: 1]
```
billylanchantin, line 138: (Being slightly nit-picky) We might want something like the following here:

```elixir
alias MyApp.Blog # Dataloader source, see below
```

The concept of a source is still being introduced. I think we should call it out explicitly because this code block might be confusing to someone who doesn't already know what `Blog` is supposed to be an example of.

guides/dataloader.md
pmargreff: Good catch, I've updated those points.

pmargreff:
Thank you for the feedback, I made the suggested changes, feel free to review again an request more it seems necessary.

guides/batching.md
billylanchantin: Fixed by: 3732ad8

guides/dataloader.md
billylanchantin: Fixed by: 3732ad8

benwilson512:
Thank you both for working on this!

PR #919, opened 2020-05-08
Fix a wrong file path in the schema docs

jonatanklosko:
Fixes an example in the [schemas#importing-types](https://hexdocs.pm/absinthe/schemas.html#importing-types) article.


benwilson512:
Thanks!

PR #923, opened 2020-05-17
fix-922

benwilson512:

### DIFF
--- a/lib/absinthe/phase/document/arguments/variable_types_match.ex
+++ b/lib/absinthe/phase/document/arguments/variable_types_match.ex
@@ -65,12 +65,12 @@ defmodule Absinthe.Phase.Document.Arguments.VariableTypesMatch do
          variable_defs
        ) do
     case Map.fetch(variable_defs, var.name) do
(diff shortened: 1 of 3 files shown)


lib/absinthe/phase/document/arguments/variable_types_match.ex
```diff
@@ -65,12 +65,12 @@ defmodule Absinthe.Phase.Document.Arguments.VariableTypesMatch do
          variable_defs
        ) do
     case Map.fetch(variable_defs, var.name) do
-      {:ok, %{schema_node: var_schema_type}} when not is_nil(var_schema_type) ->
```
benwilson512, line 68: @billylanchantin unfortunately it turns out this check isn't safe, because if the type is wrapped in `%Type.NonNull{}` then it isn't `nil`, but the inner type could be `nil`. We have to check for it being `nil` down on line 73 after the type has been unwrapped.

billylanchantin: Ah I see. Thanks for the fix!

Issue #656, opened 2018-12-25
`default_value` and explicit null/nil


lowip:
### Environment

* Elixir version:
```sh
Erlang/OTP 21 [erts-10.2.1] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe] [dtrace]
Elixir 1.7.4 (compiled with Erlang/OTP 21)
```
* Absinthe version (mix deps | grep absinthe):
```sh
* absinthe 1.4.13 (Hex package) (mix)
  locked at 1.4.13 (absinthe) 81eb2ff4
```
* Client Framework and version (Relay, Apollo, etc): NA

### Expected behavior

With the following `input_object`:
```elixir
input_object :comment_filter do
  field :user_id, :id
  field :post_id, :id
  field :parent_id, :id, default_value: nil
end
```
used in the `connection field`:
```elixir
connection field :comments, node_type: :comment do
  arg :filter, non_null(:comment_filter)
  middleware ParseIDs, filter: [user_id: :user, post_id: :post, parent_id: :comment]
  resolve &Resolvers.Core.comments/2
end
```
called with the query:
```graphql
{
  comments(first: 1, filter: {postId: "POST_ID"}) {
    edges { cursor }
  }
}
```
I expect my resolver `&Resolvers.Core.comments/2` to be called with the following args:
```elixir
def comments(args, _res) do
  IO.inspect args
  # %{filter: %{parent_id: nil, post_id: "POST_ID"}, first: 1}
  ...
end
```

### Actual behavior

Instead my args are missing the `nil` default value for `:parent_id`:
```elixir
def comments(args, _res) do
  IO.inspect args
  # %{filter: %{post_id: "POST_ID"}, first: 1}
  ...
end
```

There is currently no way to represent an explicit null/nil value using the `default_value` option.

According to `graphql-js`, when setting `null` as the default value, it should be taken into account:
- https://git.io/fhTT0
- https://git.io/fhTTB

Let me know if this is the expected behaviour for Absinthe, if so I will close the issue.

pmargreff:
I'm working on it.

pmargreff:
[...]

PR #933, opened 2020-05-22
Fix/allow null as default

pmargreff:
This issue is a Draft to solve #656. 

The problem as described on the issue happens because absinthe is using `nil` with an emptiness meaning. In this scenario, Absinthe always interprets `default_value: nil` as something to be removed, even when the user defines it. 

My current approach to the solution is to create a new `%UndefinedDefault{}` type and start it with this value. My current solution has some regressions which I couldn't solve at this moment: 

1. Some phase is trying to use the value when the `default_value` to `value` and `data` when the `default_value: %Type.UndefinedDefault{}`. And it's the root of current regressions.
2. I couldn't identify what is the best place to create a new type. My current guesses are `Absinthe.Blueprint.Schema` once `Absinthe.Type` is used only to deal with types who concern the users. 
3. Should I prevent undefined default from being set on `data`? If positive, where it happens?
4. Should the new type support some protocol/behavior? If positive, which ones? 

Opening as a draft, if we have a better way to do this or some refactor is necessary before implementing the feature itself we can close this one and start again.

### DIFF
--- a/lib/absinthe/blueprint/input/argument.ex
+++ b/lib/absinthe/blueprint/input/argument.ex
@@ -36,14 +36,7 @@ defmodule Absinthe.Blueprint.Input.Argument do
       %__MODULE__{input_value: %{normalized: %Blueprint.Input.Null{}}, value: nil} ->
         true
 
-      %__MODULE__{
-        input_value: %{
-          normalized: %Absinthe.Blueprint.Input.Generated{
-            by: Absinthe.Phase.Document.MissingVariables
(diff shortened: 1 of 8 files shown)


benwilson512:
Hey @pmargreff I think this was off to a good start, I just hadn't had a chance to review.

pmargreff:
Hello, @benwilson512. I took another path which seems better now, I'm currently opening the new PR. :)

PR #936, opened 2020-05-27
fix-persistent-term

benwilson512:

### DIFF
--- a/lib/absinthe/schema/provider.ex
+++ b/lib/absinthe/schema/provider.ex
@@ -25,6 +25,4 @@ defmodule Absinthe.Schema.Provider do
   @callback __absinthe_directives__(schema_identifier) :: Absinthe.Type.Directive.t()
 
   @callback __absinthe_interface_implementors__(schema_identifier) :: term
-
-  @callback __absinthe_prototype_schema__(schema_identifier) :: schema_identifier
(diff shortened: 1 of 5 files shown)


lib/absinthe/schema/provider.ex
```diff
@@ -25,6 +25,4 @@ defmodule Absinthe.Schema.Provider do
   @callback __absinthe_directives__(schema_identifier) :: Absinthe.Type.Directive.t()
 
   @callback __absinthe_interface_implementors__(schema_identifier) :: term
-
-  @callback __absinthe_prototype_schema__(schema_identifier) :: schema_identifier
```
benwilson512, line 29: Making this part of the behaviour was a mistake. The prototype schema is only used for the _construction_ of the schema, it isn't used at runtime, and it can't depend on the schema already existing.

derekbrown:
@benwilson512 No longer broken, and I believe it to be working! Seeing an increase in query perf. Need to do a bit more testing to better quantify this, though.

Also getting this upon build:
```
warning: undefined module attribute @prototype_schema, please remove access to @prototype_schema or explicitly set it before access
  lib/graphql/schema.ex:1: MyApp.GraphQL.Schema (module)
```

PR #940, opened 2020-06-04
Refactor/extract steps from fill defaults

pmargreff:
I'm sending this PR to facilitate to #656 implementation. 

I lost a lot of time (most then I want to admit) trying to change `default_value` value behavior and braking a lot of features not related with default values. Mainly because we have some steps of the pipeline who are responsible for multiple things, and, at the same time, the same step is implemented in multiple places.  

How the defaults are being handled: 
- MissingLiterals: it builds missing nodes **and** fills defaults **and** and mark missing ones. 
- MissingVariables: it fills defaults **and** marks missing `non_null` fields.  

After the refactor steps are better defined, happening once at time in this sequence: 
1. `BuildMissingNodes` - It creates the nodes if they will be used on the next steps (both for fill defaults or mark as missing). 
2. `FillDefaults` - If a valid default value exists and needs to be set the node, it happens only here.
3. `FlagMissing` - If it has `non_null` node without a value, it's flagged in this step.  

This PR should help to change any of these features in the future without the current side effects.

I also moved all the pipeline steps to `Argument` dir, once it looks coherent, and I think it's living on the root dir only because nobody changed those files for a long time.  

Last not least, [This line](https://github.com/absinthe-graphql/absinthe/compare/master...pmargreff:refactor/extract-steps-from-fill-defaults?expand=1#diff-5c2f8ac53c787b513d041ff0f2f6f9f0L42) seems like it is or redundant, or without a proper test. Would you help me to find out which is the test we're missing here?

My next steps should be:
- [ ] - Replace initial `default_value` with a proper type (`%UndefinedDefault{}`). 
- [ ] - Make `nil` a acceptable `default_value`.

### DIFF
--- a/lib/absinthe/blueprint/input/argument.ex
+++ b/lib/absinthe/blueprint/input/argument.ex
@@ -36,7 +36,10 @@ defmodule Absinthe.Blueprint.Input.Argument do
       %__MODULE__{input_value: %{normalized: %Blueprint.Input.Null{}}, value: nil} ->
         true
 
-      %__MODULE__{value: nil} ->
+      %__MODULE__{input_value: %{normalized: %Blueprint.Input.Generated{}}, value: nil} ->
+        true
+
+      %__MODULE__{value: :nil} ->
         false
 
       arg ->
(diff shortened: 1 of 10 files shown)


pmargreff:
Also, if you think it's better to merge it and create a new PR for each missing feature, please, let me know.

PR #943, opened 2020-06-08
Update subscriptions.md

vadimshvetsov:
I was little confused while tried to add this to `Endpoint` because felt to be in context of `Endpoint` module 😃 :
```elixir
use Absinthe.Phoenix.Socket,
    schema: MyApp.Web.Schema
```
I think setup guide can be more clear with adding `UserSocket` filename to the step instruction.


benwilson512:
Great, thank you!

PR #944, opened 2020-06-08
Fix enum default value to work properly with list values.

### LINKED ISSUE #941: [Enum] Default value not interpreted properly in case of keyword list
Given the following enum definition

```ex
enum :project_order_by do
  value :inserted_at_asc, as: [asc: :inserted_at]
  value :inserted_at_desc, as: [desc: :inserted_at]
end
```

and the following field with a default enum argument

```ex
field :projects, list_of(:project) do
  arg :order_by, non_null(:project_order_by), default_value: [asc: :inserted_at]
  resolve &Resolvers.Projects.list_projects/3
end
```

I get the following error

```
The default_value for an enum must be present in the enum values.

Could not use default value of `[asc: :inserted_at]` for :project_order_by.

Valid values are:

 * [desc: :inserted_at]
 * [asc: :inserted_at]
```

### Reason

https://github.com/absinthe-graphql/absinthe/blob/00601ffc3aac8f0ce989113330213ede70c66bbf/lib/absinthe/phase/schema/validation/default_enum_value_present.ex#L37-L38

The above call to `List.wrap` does nothing in case of `[asc: :inserted_at]`, but then it uses `Enum.all?` and ends up checking if `{:asc, :inserted_at}` belongs to values instead of `[asc: :inserted_at]`.

What is the reasoning behind using `List.wrap`?
*Edit: it was introduced in #762 to handle default value for `list_of(enum)`.*

### Environment

* Elixir version (elixir -v): Elixir 1.9.1 (compiled with Erlang/OTP 22)
* Absinthe version (mix deps | grep absinthe): 1.5.1

jonatanklosko:
An attempt to fix #941.

### The problem

The current `default_value` for enum looks like this:

```ex
default_valid? =
  List.wrap(default_value)
  |> Enum.all?(fn default -> default in values end)
```

It's quite limited as doesn't really take type structure into consideration. Most importantly it breaks when enum values are defined to be lists.

### Example

```ex
defmodule CorrectCatSchema do
  use Absinthe.Schema

  query do

    field :cats,
      type: non_null(list_of(non_null(:cat))),
      args: [
        order_by: [type: non_null(:cat_order_by), default_value: [{:asc, :name}]],
      ],
      resolve: fn
        %{order_by: _}, _ -> {:ok, []}
      end

  end

  enum :cat_order_by do
    value :name_asc, as: [{:asc, :name}]
    value :name_desc_inserted_at_asc, as: [{:desc, :name}, {:asc, :inserted_at}]
  end

  object :cat do
    field :name, :string
  end
end
```

What the mentioned code does is checking `{:asc, :name} in enum_values` instead of the expected `[{:asc, :name}] in enum_values`.

### Solution

With type definition at hand we can analyze the given `default_value` and figure out which inner values *should actually* be enum values.

In the example above we immediately see it's the enum type `non_null(:cat_order_by)`, so we check if `[{:asc, :name}]]` is a valid enum value.

Let's consider a nested case:

```ex
args: [
  genres: [type: non_null(list_of(non_null(:movie_genre))), default_value: [:action, :OTHER]],
],
```

Here we first see a **list type** and a `default_value` of `[:action, :OTHER]`, so we can dive deeper into the structure and check:
- `non_null(:movie_genre)` against `:action`
- `non_null(:movie_genre)` against `:OTHER`

In both cases we reached the inner enum type, so we can simply check if `:action` is within valid values and if `:OTHER` is within valid values.

### DIFF
--- a/lib/absinthe/phase/schema/validation/default_enum_value_present.ex
+++ b/lib/absinthe/phase/schema/validation/default_enum_value_present.ex
@@ -33,20 +33,18 @@ defmodule Absinthe.Phase.Schema.Validation.DefaultEnumValuePresent do
         values = Enum.map(enum.values, & &1.value)
         value_list = Enum.map(values, &"\n * #{inspect(&1)}")
 
-        default_valid? =
-          List.wrap(default_value)
-          |> Enum.all?(fn default -> default in values end)
+        case value_conforms_to_enum(node.type, default_value, values) do
+          {:error, value} ->
+            detail = %{
+              value_list: value_list,
+              type: type,
+              default_value: value
+            }
 
-        if not default_valid? do
-          detail = %{
-            value_list: value_list,
-            type: type,
-            default_value: default_value
-          }
+            node |> put_error(error(node, detail))
 
-          node |> put_error(error(node, detail))
-        else
-          node
+          {:ok, _} ->
+            node
         end
 
       _ ->
@@ -56,6 +54,25 @@ defmodule Absinthe.Phase.Schema.Validation.DefaultEnumValuePresent do
 
   def validate_defaults(node, _) do
     node
+  end
+
+  defp value_conforms_to_enum(%Blueprint.TypeReference.List{of_type: of_type}, value, enum_values)
+       when is_list(value) do
+    value
+    |> Enum.map(&value_conforms_to_enum(of_type, &1, enum_values))
+    |> Enum.find({:ok, value}, &match?({:error, _}, &1))
+  end
+
+  defp value_conforms_to_enum(%_{of_type: of_type}, value, enum_values) do
+    value_conforms_to_enum(of_type, value, enum_values)
+  end
+
+  defp value_conforms_to_enum(_, value, enum_values) do
+    if value in enum_values do
+      {:ok, value}
+    else
+      {:error, value}
+    end
   end
 
   defp error(node, data) do
--- a/test/absinthe/schema/rule/default_enum_value_present_test.exs
+++ b/test/absinthe/schema/rule/default_enum_value_present_test.exs
@@ -2,7 +2,7 @@ defmodule Absinthe.Schema.Rule.DefaultEnumValuePresentTest do
   use Absinthe.Case, async: true
 
   describe "rule" do
-    test "is enforced when the defaultValue is not in the enum" do
+    test "is enforced when the default_value is not in the enum" do
       schema = """
       defmodule BadColorSchema do
         use Absinthe.Schema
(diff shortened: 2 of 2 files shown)


lib/absinthe/phase/schema/validation/default_enum_value_present.ex
```diff
@@ -58,6 +56,25 @@ defmodule Absinthe.Phase.Schema.Validation.DefaultEnumValuePresent do
...
+    value_conforms_to_enum(of_type, value, enum_values)
+  end
+
+  defp value_conforms_to_enum(_, value, enum_values) do
+    if value in enum_values do
+      {:ok, value}
+    else
+      {:error, value}
+    end
+  end
```
jonatanklosko, line 76: Initially I started with a simpler version that would just return a boolean indicating if a value fits an enum. Such approach did however lead to quite misleading error messages, so I figured it's better to show which inner value actually failed to match the enum, and that's why this method returns it.

binaryseed:
Hi there! I'm a bit surprised by this usage of `:as`, and I suspect it wasn't intended to support complex data structures... @benwilson512 do you think Absinthe should support this kind of Enum value?

jonatanklosko:
@binaryseed I run into this issue while trying to use keyword list as value, so that I could directly pass it to `Ecto.order_by`. Specifically, having the following definition:

```ex
enum :project_order_by do
  value :inserted_at_asc, as: [asc: :inserted_at]
  value :inserted_at_desc, as: [desc: :inserted_at]
end
```

it's no longer necessary to have a mapping from each enum value to the proper keyword list (as we do it here). Not sure if there's anything particularly bad with this approach, but it seemed pretty appealing to me =)

jonatanklosko:
@benwilson512 any thoughts? =)

benwilson512:
I'm not the biggest fan of this pattern I think but the code you wrote is definitely better than the original implementation. I also think that Absinthe is best off treating the `as` value as opaque, which means supporting this sort of thing.

jonatanklosko:
Great, thanks! So in general you'd prefer keeping the values as atoms and then convert them to an order clause before the actual query?

benwilson512:
Well for example `:name_desc_inserted_at_asc` is not a pattern that extends well. The more columns you want to support, the more possible permutations there are. Instead of enumerating these permutations as individual enums you could support an input of `order: [{direction: ASC, attribute: NAME}, {direction: DESC, attribute: STARTS_AT}]` where direction is an enum of `ASC | DESC` and `attribute` is either a string input or an enum of all attributes you wish to allow ordering on.

jonatanklosko:
Alright, that makes perfect sense! One point that defends `name_desc_inserted_at_asc` is that you can create proper indices to optimize whatever you support, when using list it's less predictable. But I agree it easily becomes cumbersome, thanks =)

Issue #949, opened 2020-06-17
Order of arguments are reversed in schema.graphql generated by absinthe.schema.sdl task


ymtszw:
### Environment

* Elixir version (elixir -v):
```
Erlang/OTP 23 [erts-11.0.2] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe]
Elixir 1.10.3 (compiled with Erlang/OTP 21)
```

* Absinthe version (mix deps | grep absinthe):
```
* absinthe 1.5.1 (Hex package) (mix)
  locked at 1.5.1 (absinthe) b264eeb6
```
* Client Framework and version (Relay, Apollo, etc):
Not directly applicable, but we're using [dillonkearns/elm-graphql](/dillonkearns/elm-graphql) for code generation from `schema.graphql` file.

### Expected behavior

When we generate `schema.graphql` file using `mix absinthe.schema.sdl` task, order of `field` and `arg` appearance in source schema modules are preserved.

I do understand that currently in GraphQL schema, these order of appearances are NOT significant by spec. But in some cases, it is valuable to preserve them.

For instance, in our case with above mentioned `elm-graphql`, it generates functions to build up required and optional arguments as "Record" (which is much like struct in Elixir.)
In such functions order of arguments are derived from definitions in `schema.graphql` (or introspection result, if we choose to use them.)
So, if they are not preserved from source schema module in Elixir app, it is quite a surprise to developers' eyes in daily development.

Related discussion issue: https://github.com/graphql/graphql-spec/issues/577

### Actual behavior

* Order of `field` under `object` or `input_object` are preserved
* Order of `arg` under `query` or `mutation` are **reversed**

### Relevant Schema/Middleware Code

* Source module
```ex
defmodule Schema do
  use Absinthe.Schema

  object :object do
    field(:f1, :id)
    field(:f2, :string)
  end

  input_object :input_object do
    field(:f3, :boolean)
    field(:f4, :integer)
  end

  query do
    field :query1, :object do
      arg(:a1, :id)
      arg(:a2, :input_object)
      resolve(fn _, _, _ -> {:error, "TODO"} end)
    end
  end

  mutation do
    field :mutation1, :object do
      arg(:a3, :id)
      arg(:a4, :input_object)
      resolve(fn _, _, _ -> {:error, "TODO"} end)
    end
  end
end
```

* Generated `schema.graphql` file

```graphql
schema {
  query: RootQueryType
[...]

PR #950, opened 2020-06-18
[dev] Update mix_test_watch version, and correct options

ymtszw:
Stumbled upon this during patch development of #949

- `:fs` (dependency of `:mix_test_watch`) could not be compiled on macOS Catalina 10.15.5
  - What I did was fork, clone, `mix deps.get` and `mix test` on `master`.
- Actually, `:mix_test_watch` provides mix task for development so requring it on `:test` env is off. It should be `:dev`
  - Doc: https://github.com/lpil/mix-test.watch#usage
  - Also we should add `runtime: false` too.
  - With these changes `mix test` succeeds on macOS.
- While not directly related, `:mix_test_watch` is now released as `1.x` with `:file_system` as intermediate backend.
  Should be no harm to upgrade on this occasion since it is a development tool, not an application dependency. (Though I'm OK to revert.)


benwilson512:
Thank you!

PR #951, opened 2020-06-22
[SDL Renderer] Preserve order of arguments in schema generated via macros

### LINKED ISSUE #949: Order of arguments are reversed in schema.graphql generated by absinthe.schema.sdl task

ymtszw:
- I *think* this should cleanly fix #949
- As commented in #949, if we reverse order of arguments at the moment of
rendering SDL, it causes current behavior to change for schema generated
via importing SDL.
- So, I've traced code related to building blueprints for schema module
written in macros. And reversed order of `:arguments` in
`Absinthe.Blueprint.Schema.build/1` while they are accumulated under
`FieldDefinition` scope.
- This resulted in expected behavior with least changes and not affected
other parts. All other tests are passing (at least for me)

Reviews are appreciated. Thank you team in advance!

### DIFF
--- a/lib/absinthe/blueprint/schema.ex
+++ b/lib/absinthe/blueprint/schema.ex
@@ -177,6 +177,7 @@ defmodule Absinthe.Blueprint.Schema do
     field =
       field
       |> Map.update!(:middleware, &Enum.reverse/1)
+      |> Map.update!(:arguments, &Enum.reverse/1)
       |> Map.update!(:triggers, &{:%{}, [], &1})
       |> Map.put(:function_ref, {obj.identifier, field.identifier})
 
--- a/test/absinthe/schema/sdl_render_test.exs
+++ b/test/absinthe/schema/sdl_render_test.exs
@@ -219,10 +219,10 @@ defmodule SdlRenderTest do
 
              type RootQueryType {
                echo(
-                 timeInterval: Int
-
                  "The number of times"
                  times: Int
+
+                 timeInterval: Int
                ): String
                search: SearchResult
              }


benwilson512:
Thank you!

binaryseed:
thanks!

ymtszw:
Thanks for the review and merge!

PR #952, opened 2020-06-22
Preserve order of enum values in schema in SDL rendering

maartenvanvliet:
More or less a followup to https://github.com/absinthe-graphql/absinthe/pull/951 

The SDL rendering now uses the same order as the declaration of
the enum values in the macro schema.


benwilson512:
Thank you!

tjdam:
Hi, @maartenvanvliet 
Sorry if this is the wrong place to ask, but does this mean the values of an enum would appear in order they are declared instead of in alphabetic order everywhere? Like if you were navigating the types in GraphiQL?

maartenvanvliet:
This PR concerns rendering the schema as SDL. I don't think this affects its introspection representation, and that is what GraphiQL uses.

PR #954, opened 2020-06-26
Add DirectivesMustBeValid schema phase

maartenvanvliet:
I noticed in the SDL rendering output that a macro directive that did not
specify the `on` clause would render invalid SDL.

E.g.
```elixir
    directive :foo do
      arg :baz, :string
    end
```

Would render in SDL
```
directive @foo(baz: String) on
```
So the directive locations are missing making it invalid.

This PR adds a phase to the schema pipeline to check whether
the directive locations are set. I also check whether the locations
themselves are valid.


binaryseed:
Awesome!

maartenvanvliet:
Thanks for merging!

Issue #961, opened 2020-08-01
Inline string arguments appear to be unescaped twice


mattbaker:
### Environment

* Elixir version (elixir -v): `Elixir 1.10.1 (compiled with Erlang/OTP 22)`
* Absinthe version (mix deps | grep absinthe): tested against `1.5.0-rc.4` and `master`
* Client Framework and version (Relay, Apollo, etc): N/A

### Expected behavior

Given a query with an inline string argument like so:

```
{
  field(argument: "\\\\FOO")
}
```

We would expect the resolver to receive the string `\\FOO`.

### Actual behavior

The resolver will receive `\FOO`.

Similarly, providing `\\\\\\\\FOO` as the argument yields `\\FOO` in the resolver, but we expected `\\\\FOO`.

It's worth mentioning that expected matches actual in Absinthe v1.4.16, we just never noticed the change after the v1.5 upgrade until now.

### Relevant Schema/Middleware Code

* `Absinthe.Lexer`
* `:absinthe_parser.yrl`
* `Absinthe.Phase.Parse`

### Details

@jmschles and I have been looking into some odd string escaping behavior, specifically with sequences of many `\`'s. As mentioned above, given a query with an inline string argument like so:

```
{
  field(argument: "\\\\FOO")
}
```

We would expect the resolver to receive the string `\\FOO` after the escape characters have been processed. Instead the resolver will receive `\FOO`.

A [lexer test](https://github.com/mattbaker/absinthe/commit/3fdd4609a4095da7abdb69d38a73dbe19cc254a6#diff-9ce8af4958da92722e4fba7e0e27e2ddR30-R50) shows the lexer is performing as expected, producing two literal slashes after tokenization:

```elixir
@query ~S"""
{
  { foo(bar: "\\\\FOO") }
}
"""
test "multiple escaped slashes"
  assert Absinthe.Lexer.tokenize(@query) ==
            {:ok,
            [
              {:"{", {1, 1}},
              {:"{", {2, 3}},
              {:name, {2, 5}, 'foo'},
              {:"(", {2, 8}},
              {:name, {2, 9}, 'bar'},
              {:":", {2, 12}},
              {:string_value, {2, 14}, ~S("\\FOO") |> String.to_char_list()},
              {:")", {2, 23}},
              {:"}", {2, 25}},
              {:"}", {3, 1}}
            ]}
end
```

[...]

PR #962, opened 2020-08-01
Fix double escaped back slashes in string arguments

mattbaker:
@binaryseed suggested I just submit the branch I used to test for #961 as a PR with failing tests as an example. Happy to update the parser to remove the escape processing and add it to this PR if that's the right direction.

### DIFF
--- a/src/absinthe_parser.yrl
+++ b/src/absinthe_parser.yrl
@@ -361,36 +361,7 @@ put_description(Node, Description) ->
 % String
 
 extract_quoted_string_token({_Token, _Loc, Value}) ->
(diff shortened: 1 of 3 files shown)


benwilson512:
@bruce any chance you can weigh in on this one?

bruce:
@mattbaker Thanks for submitting this. I've made a commit that removes the escape handling from the parser code, relying instead on the lexer to handle it. Tests are passing locally. Please take a look.

mattbaker:
Makes sense to me @bruce 👍 The only thing on my mind is that this may actually be a breaking change, in case that impacts how the change is released or communicated.

I suspect it's rare for people to be escaping so heavily in a string, we just happen to have a couple specific cases in New Relic's API where folks are (which is for us to figure out).

PR #968, opened 2020-09-11
Escape single-lined description in SDL

barthez:
Hello!

I got a pretty strange error while generating types for typescript out of schema SDL. I have switched from rendering with https://github.com/maartenvanvliet/absinthe_sdl to `Absinthe.Schema.to_sdl`. It turned out I had a macro enum definition like this:

```elixir
enum :logic do
  value :and, description: "Apply \"and\" logic to query"
  value :or, description: "Apply \"or\" logic to query"
end
```

The type generator complained about duplicated values in enum. When I took a closer look at generated SDL it looked like:
```graphql
enum Logic {
  "Apply "and" logic to query"
  AND

  "Apply "or" logic to query"
  OR
}
```
Which, after stripping descriptions, is:
```graphql
enum Logic {
  and
  AND
  or
  OR
}
```

I got into GraphQL specs for strings: http://spec.graphql.org/June2018/#sec-String-Value

It seems some characters need to be escaped in a single line description string.

### DIFF
--- a/lib/absinthe/schema/notation/sdl_render.ex
+++ b/lib/absinthe/schema/notation/sdl_render.ex
@@ -356,7 +356,7 @@ defmodule Absinthe.Schema.Notation.SDL.Render do
     |> String.split("\n")
     |> case do
       [description] ->
-        [~s("), description, ~s("), line(), docs]
+        [~s("), escape_description(description), ~s("), line(), docs]
 
       description_lines ->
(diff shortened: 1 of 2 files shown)


barthez:
Hey @benwilson512, could you please give it a look?

benwilson512:
Great catch, thanks!

PR #972, opened 2020-09-25
Remove :inets dependency

hogjosh:
There's an implicit dependency on `:inets` when using `:httpd_util` while parsing documents with unicode characters.  Our distillery release does not include `:inets` by default so `:httpd_util` functions are unavailable at runtime. 

This PR *should* remove the dependency on `:inets` altogether.

### DIFF
--- a/lib/absinthe/lexer.ex
+++ b/lib/absinthe/lexer.ex
@@ -279,7 +279,7 @@ defmodule Absinthe.Lexer do
 
   defp unescape_unicode(_rest, content, context, _loc, _) do
     code = content |> Enum.reverse()
-    value = :httpd_util.hexlist_to_integer(code)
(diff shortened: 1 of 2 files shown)


lib/absinthe/lexer.ex
```diff
@@ -279,7 +279,7 @@ defmodule Absinthe.Lexer do
 
   defp unescape_unicode(_rest, content, context, _loc, _) do
     code = content |> Enum.reverse()
-    value = :httpd_util.hexlist_to_integer(code)
+    value = :erlang.list_to_integer(code, 16)
```
binaryseed, line 282: nice.

reference: https://github.com/erlang/otp/blob/master/lib/inets/src/http_lib/http_util.erl#L156-L157

benwilson512:
Thanks!

Issue #958, opened 2020-07-14
Inlined meta(s) do not work with field macro anymore


tlvenn:
Trying to upgrade our app / schema to the latest 1.5 release (still using 1.4 for now) and I got a compilation error in our schema regarding the usage of meta with field such as:

```
input do
  field(:country_id, :id, meta: [type: Jumpn.SupportedCountry])
end
```
 
The error:
```
KeyError) key :meta not found in: %Absinthe.Blueprint.Schema.FieldDefinition{__private__: [], __reference__: %{location: %{file: ".../activity/schema.ex", line: 29}, module: Jumpn.Schema.Activity}, arguments: [], complexity: nil, config: nil, default_value: nil, deprecation: nil, description: nil, directives: [], errors: [], flags: %{}, function_ref: nil, identifier: :country_id, middleware: [], module: Jumpn.Schema.Activity, name: "country_id", source_location: nil, triggers: [], type: :id}
    (stdlib 3.13) :maps.update(:meta, [type: Jumpn.SupportedCountry], %Absinthe.Blueprint.Schema.FieldDefinition{__private__: [], __reference__: %{location: %{file: ".../activity/schema.ex", line: 29}, module: Jumpn.Schema.Activity}, arguments: [], complexity: nil, config: nil, default_value: nil, deprecation: nil, description: nil, directives: [], errors: [], flags: %{}, function_ref: nil, identifier: :country_id, middleware: [], module: Jumpn.Schema.Activity, name: "country_id", source_location: nil, triggers: [], type: :id})
    lib/absinthe/blueprint/schema/field_definition.ex:7: anonymous fn/2 in Absinthe.Blueprint.Schema.FieldDefinition.__struct__/1
```

Might be related with #928

jeroenvisser101:
It's for sure related to #927

tlvenn:
Hi @benwilson512 , is there any chance you would have a little bit of spare time to look into this ? Thanks a lot in advance.

tlvenn:
Fixed with #973

PR #973, opened 2020-09-28
Allow :meta on field

Hermanverschooten:
As mentioned on Slack and in issue #958, inline meta does not work.
I think this add that ability back.


tlvenn:
Thanks a lot for this PR @Hermanverschooten , @benwilson512 this seems to be fine, what do you think ?

benwilson512:
Looks great, thanks!

PR #975, opened 2020-10-01
Support Decimal v2

mattmatters:
Mostly minor changes

* match on the new return type for `Decimal.parse/1`
* Swap `Decimal.cmp/2` for `Decimal.compare/2` as former is deprecated
* Use decimal v2 for local development

---

I was going through updating our dependencies and sadly Absinthe was requiring v1. Not sure what your preference is on allowing multiple major versions of dependencies or which one is preferred in the `mix.lock` file, so I made a broad assumption to stick the latest in there.


benwilson512:
this is perfect, thanks!

PR #982, opened 2020-10-21
Fix compilation bug with persistent_term backend

binaryseed:
This PR fixes a compile time bug with the `persistent_term` schema backend.

`PopulatePersistentTerm.run/2` was only getting called during compilation. So if another file changes but not the schema module, it doesn't get re-compiled and thus it's not populated - this can cause compile-time references to the schema to fail.

This is fixed by populating the `persistent_term` backend upon `@on_load`. This also means we don't need to startup a GenServer in the user's app to populate, which paves the way for easier adoption of this in the future.

@benwilson512

### DIFF
--- a/lib/absinthe/schema.ex
+++ b/lib/absinthe/schema.ex
@@ -110,6 +110,18 @@ defmodule Absinthe.Schema do
 
       @schema_provider Absinthe.Schema.Compiled
 
+      @on_load :__on_load__
+      def __on_load__ do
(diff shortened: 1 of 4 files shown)


benwilson512:
This is awesome, thank you!

binaryseed:
I think this PR is flawed, so I have a PR that fixes the original failure we were encountering https://github.com/absinthe-graphql/absinthe_plug/pull/234

PR #999, opened 2020-12-05
Add support for repeatable directives

maartenvanvliet:
Adds support for repeatable directives as defined by the spec (https://spec.graphql.org/draft/#sec-Type-System.Directives). See https://github.com/graphql/graphql-spec/pull/472 where the RFC was merged for more information.

This PR does several things:
  * adds support for the `repeatable` keyword for directives  in the parser
  * directives by default have `repeatable` set to false
  * renders directives with `repeatable` in SDL
  * adds `isRepeatable` to the directive type available for introspection
  * adds `repeatable(true|false)` to the schema definition macro's placed under directive
  * adds validation phase in the document pipeline to check whether a non-repeatable phase is applied multiple times
  * changes the directive type to match the spec (other built-in types could also be tightened with `non_null` declarations to closer match the spec, but can be done in a separate PR)

In theory this could be a breaking change, i.e. if the built-in directives were applied multiple times to the same field in a document this will now trigger an error.

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -32,6 +32,7 @@ locals_without_parens = [
   object: 3,
   on: 1,
   parse: 1,
+  repeatable: 1,
   resolve: 1,
   resolve_type: 1,
   scalar: 2,
--- a/lib/absinthe/blueprint/schema/directive_definition.ex
+++ b/lib/absinthe/blueprint/schema/directive_definition.ex
@@ -12,6 +12,7 @@ defmodule Absinthe.Blueprint.Schema.DirectiveDefinition do
     directives: [],
     arguments: [],
     locations: [],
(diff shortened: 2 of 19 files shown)


benwilson512:
@maartenvanvliet this is fantastic, thank you! We will probably target this for the 1.6 release which we hope to go ahead with by the end of the year due to the potential for breakage you note. Will review implementation here shortly.

binaryseed (approved):
This looks awesome! From my read, it looks like everything is in place & tested.. Nice work!

binaryseed:
Followup that adds this to the `IntrospectionQuery` provided by Absinthe: https://github.com/absinthe-graphql/absinthe/pull/1008

PR #1006, opened 2020-12-22
Reduce amount of subscription data stored in Elixir registry

zoldar:
This PR is a follow-up to a proposal that was initially posted on [Elixir slack](https://elixir-lang.slack.com/archives/C0PR49P4P/p1608111062497700) and on [Elixir Forum](https://elixirforum.com/t/proposal-reduce-ets-memory-utilized-by-absinthe-subscriptions/36290).

Instead of storing precomputed `initial_phases` where options (along with context) are repeated several times over, we are storing just enough information to reconstruct them using `Absinthe.Pipeline.for_document/2` which comes down to building a list of phases with options put in a number of spots, which should not be computationally expensive.

If I haven’t missed anything obvious, that change should be transparent and non-breaking :sweat_smile:

I have run a series of simple tests with absinthe’s test suite as well as test suite of a larger app I’m working on that uses subscriptions extensively. For the purpose of the experiment I have added a line calculating byte size of a serialized map stored in elixir registry by `Absinthe.Subscriptions.subscribe`, like this:

```elixir
doc_value |> :erlang.term_to_binary() |> byte_size() |> IO.inspect(label: :PAYLOAD)
```

This is not the same amount of space it takes up in ETS memory but it gives an idea in relative terms. Then I have run tests for both absinthe and the app without and with the patch applied. Here are the results:

https://gist.github.com/zoldar/9f06e95bfced82cd25793c25f2e632c5

As can be seen, the memory usage can be lowered even >10x this way. Given, in our case, roughly ~90% of VM memory in production is taken up by those ETS entries for subscriptions (multiple subscriptions per user), that would substantially lower memory usage. What do you think?

_**UPDATE:** Here’s one more round of test runs against our app where the sum of memory occupied by the ETS tables for key partitions is checked (10648 is subtracted from each table because that seems to be the base size on this particular setup):_

https://gist.github.com/zoldar/d99c636a933725c73d2a4f7131fdd936

### DIFF
--- a/lib/absinthe/subscription.ex
+++ b/lib/absinthe/subscription.ex
@@ -31,6 +31,8 @@ defmodule Absinthe.Subscription do
   require Logger
   alias __MODULE__
 
+  alias Absinthe.Subscription.PipelineSerializer
+
   @doc """
   Add Absinthe.Subscription to your process tree.
   """
@@ -106,15 +108,14 @@ defmodule Absinthe.Subscription do
   defp fetch_fields(_, _), do: []
 
   @doc false
-  def subscribe(pubsub, field_key, doc_id, doc, options) do
+  def subscribe(pubsub, field_key, doc_id, doc) do
     registry = pubsub |> registry_name
 
     doc_value = {
       doc_id,
       %{
-        schema: doc.schema,
-        source: doc.source,
-        options: options
+        initial_phases: PipelineSerializer.pack(doc.initial_phases),
+        source: doc.source
       }
     }
 
@@ -140,8 +141,12 @@ defmodule Absinthe.Subscription do
     pubsub
     |> registry_name
     |> Registry.lookup(key)
-    |> Enum.map(&elem(&1, 1))
-    |> Map.new()
+    |> Enum.map(fn match ->
+      {_, {name, doc}} = match
+      doc = Map.update!(doc, :initial_phases, &PipelineSerializer.unpack/1)
+
+      {name, doc}
+    end)
   end
 
   @doc false
--- a/lib/absinthe/subscription.ex
+++ b/lib/absinthe/subscription.ex
@@ -106,14 +106,15 @@ defmodule Absinthe.Subscription do
   defp fetch_fields(_, _), do: []
 
   @doc false
-  def subscribe(pubsub, field_key, doc_id, doc) do
+  def subscribe(pubsub, field_key, doc_id, doc, options) do
     registry = pubsub |> registry_name
 
     doc_value = {
       doc_id,
       %{
-        initial_phases: doc.initial_phases,
-        source: doc.source
+        schema: doc.schema,
+        source: doc.source,
+        options: options
       }
     }
 
--- a/lib/absinthe/subscription.ex
+++ b/lib/absinthe/subscription.ex
@@ -142,10 +142,10 @@ defmodule Absinthe.Subscription do
     |> registry_name
     |> Registry.lookup(key)
     |> Enum.map(fn match ->
-      {_, {name, doc}} = match
+      {_, {doc_id, doc}} = match
       doc = Map.update!(doc, :initial_phases, &PipelineSerializer.unpack/1)
 
-      {name, doc}
+      {doc_id, doc}
     end)
   end
 
--- a/lib/absinthe/subscription/phase_serializer.ex
+++ b/lib/absinthe/subscription/phase_serializer.ex
@@ -1,0 +1,68 @@
+defmodule Absinthe.Subscription.PipelineSerializer do
+  @moduledoc """
+  Serializer responsible for packing and unpacking pipeline stored in the Elixir registry.
+
+  The purpose of this logic is saving memory by deduplicating repeating options - (ETS
+  backed registry stores them flat in the memory).
+  """
+
+  alias Absinthe.{Phase, Pipeline}
+
+  @type options_label :: {:options, non_neg_integer()}
+
+  @type packed_phase_config :: Phase.t() | {Phase.t(), options_label()}
+
+  @type options_map :: %{options_label() => Keyword.t()}
+
+  @type packed_pipeline :: {:packed, [packed_phase_config()], options_map()}
+
+  @spec pack(Pipeline.t()) :: packed_pipeline()
+  def pack(pipeline) do
+    {reverse_pipeline, options_reverse_map} =
+      pipeline
+      |> List.flatten()
+      |> Enum.reduce({[], %{}}, fn phase, {pipeline, options_reverse_map} ->
+        {phase, options_reverse_map} = maybe_pack_phase(phase, options_reverse_map)
+
+        {[phase | pipeline], options_reverse_map}
+      end)
+
+    pipeline = Enum.reverse(reverse_pipeline)
+    options_map = Map.new(options_reverse_map, fn {options, label} -> {label, options} end)
+
+    {:packed, pipeline, options_map}
+  end
+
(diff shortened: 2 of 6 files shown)


benwilson512:
Hey @zoldar this is really great! The current structure of some of this stuff definitely causes issues when the structures are flattened inside of `:ets`. The changes you're making here are very small and look pretty safe, but I need to make sure that the phases are the correct ones. I'll pull this down today and take a look, thanks!

lib/absinthe/subscription.ex
```diff
@@ -106,14 +106,15 @@ defmodule Absinthe.Subscription do
...
 
   @doc false
-  def subscribe(pubsub, field_key, doc_id, doc) do
+  def subscribe(pubsub, field_key, doc_id, doc, options) do
     registry = pubsub |> registry_name
 
     doc_value = {
       doc_id,
       %{
-        initial_phases: doc.initial_phases,
```
benwilson512, line 115: The more I think about this the more I'm certain that this is not an equivalent change. `doc.initial_phases` is going to capture any and all changes that the end user made to the pipeline that was run, but when we throw that away and instead use ` pipeline = Absinthe.Pipeline.for_document(doc.schema, doc.options)` there is no real reason to believe it will return the same pipeline. All that `for_document` does is return the basic default pipeline, it won't re-run any pipeline modifiers that the end user has.

benwilson512 (changes requested):
I think the specific approach here can't work, `for_document/2` is simply not equivalent to `doc.initial_phases`. However, if there are a lot of memory savings to be had, then perhaps we can do some "compression" of sorts on the pipeline that is registered to minimize the space duplication that occurs when it is flattened in the `:ets` table. Let me take a look and see.

lib/absinthe/subscription.ex
zoldar: Ok, then indeed I might have misunderstood the flow here 😓 . I went off the assumption that `initial_phases` is usually populated at the very beginning (by `Init` phase) - and that `SubscribeSelf` uses that unchanged, initial pipeline instead of the actual flow adjusted by user. This does not hold though, right?

If that's the case, then perhaps, indeed, a "compression" approach, where occurrences of unique options kwlists could be stored in a separate map and the pipeline would hold only their placeholders, (`{:options, 1}`, `{:options, 2}` etc.). The pipeline could be easily reconstructed then when publishing it.

zoldar:
@benwilson512 You are right of course, I got hung up on `Absinthe.run` using that default pipeline 🤦 

I have made a PoC of an alternative approach in https://github.com/absinthe-graphql/absinthe/pull/1006/commits/187e65522e17c2ddadcf3498f09305cdf5f5264e - as described in https://github.com/absinthe-graphql/absinthe/pull/1006#discussion_r548699983.

I have run that version against our app's test suite again and the savings seem to be still pretty sizable:

https://gist.github.com/zoldar/f185d0896a217e36db13bc1f67f7f12e

I'm still going to verify it with additional tests, but when inspecting the serialized/deserialised data, it _seems_ to behave correctly. What do you think about this approach?

binaryseed:
I'm curious if we can look at the broader context of this... what is it that is winding up in `options` that is getting duplicated & taking up a lot of memory?

benwilson512:
@binaryseed that's a good question. Often one killer thing for folks is the context, a large context can explode things very fast. Large inputs (variables) can do the same thing.

In Absinthe 2.0 I want to build a proper `%Absinthe.Pipeline{}` struct that would contain a simple `phases` key which would be a list of just the modules, and then an `options` key, I think per phase options adds unnecessary complications.

I think the best option right now is to sort of re-create that structure by de-duplicating the options. Off the top of my head:

```
defp compress([], phases, opts_map, _index) do
  {Enum.reverse(phases), Map.new(opts_map, fn {k, v} -> {v, k} end)}
end

defp compress([phase | phases], compressed_phases, opts_map, index) do
  case phase do
    {phase, opts} ->
      {opts_map, index} = compress_opts(opts_map, opts, index)
      compress(phases, [{phase, index} | compressed_phases], opts_map, index)
    phase ->
      compress(phases, [phase, compressed_phases], opts_map, index)
  end
end

defp compress_opts(opts_map, opts, index) do
  if Map.has_key?(opts_map, opts) do
    {opts_map, index}
  else
    index = index + 1
    opts_map = Map.put(opts_map, opts, index)
    {opts_map, index}
  end
end
```

Basically, this turns a list of `{phase, opts}` tuples into a list of `{phase, integer}` tuples where the `integer` maps to a unique opts value. Haven't actually run this but I'm pretty sure it's roughly correct. Decompressing is as simple as:

```
Enum.map(compressed_phases, fn
  {phase, i} -> {phase, Map.fetch!(opts_map, i)}
  phase -> phase
end)
```

zoldar:
@benwilson512 Just to double-check - have you seen https://github.com/absinthe-graphql/absinthe/commit/187e65522e17c2ddadcf3498f09305cdf5f5264e ? 😅 (mentioned in https://github.com/absinthe-graphql/absinthe/pull/1006#issuecomment-751709632)

lib/absinthe/subscription/phase_serializer.ex
```diff
@@ -0,0 +1,68 @@
...
+  @type options_map :: %{options_label() => Keyword.t()}
+
+  @type packed_pipeline :: {:packed, [packed_phase_config()], options_map()}
+
+  @spec pack(Pipeline.t()) :: packed_pipeline()
+  def pack(pipeline) do
+    {reverse_pipeline, options_reverse_map} =
+      pipeline
+      |> List.flatten()
+      |> Enum.reduce({[], %{}}, fn phase, {pipeline, options_reverse_map} ->
```
benwilson512, line 24: Might as well use `map_reduce` here to avoid the reverse at the end.

lib/absinthe/subscription/phase_serializer.ex
```diff
@@ -0,0 +1,68 @@
...
+    pipeline
+  end
+
+  defp maybe_pack_phase({phase, options}, options_reverse_map) do
+    if Map.has_key?(options_reverse_map, options) do
+      options_label = options_reverse_map[options]
+
+      {{phase, options_label}, options_reverse_map}
+    else
+      new_index = map_size(options_reverse_map)
```
benwilson512, line 57: Ah clever, using the map size as an implicit indexer.

benwilson512 (approved):
Ah lol I missed that sorry. Looks great, and I'm glad to see there are still size savings.

zoldar:
@benwilson512 Cool! I'll get it into shape (apply feedback and serializer tests) soon.

lib/absinthe/subscription/phase_serializer.ex
zoldar: Addressed in https://github.com/absinthe-graphql/absinthe/pull/1006/commits/67acc31e133ee46349a84f24ef4b4680f7d17322

zoldar:
@benwilson512 The PR should be now fully ready for review 😅 

One observation though. I'm keeping a fork of an older tag of absinthe at the moment with the same patch applied. The fork does not contain this change: b407000f. Telemetry phases there are repeating the original options two times with customized `event` for each respective phase. For current version of absinthe that means - provided this default pipeline serves as a base - that even the packed version will always have at least 3 full copies of options stored from the get go - with a significant overlap.

I'm wondering if it would we worth the effort to make the serializer try to handle this, doing an extra processing step (diffing existing options and finding a common base perhaps)? Maybe there's a better way to handle this though.

zoldar:
If there's anything that needs addressing to make this patch acceptable, please let me know. Thanks!

benwilson512:
This looks great, thank you!

Issue #987, opened 2020-11-02
Nullable interface fields cannot have their implementing sub-type be non-nullable.


nickdichev:
### Environment

* Elixir version (elixir -v): Elixir 1.11.1 (compiled with Erlang/OTP 23)
* Absinthe version (mix deps | grep absinthe): absinthe 1.5.3 (Hex package) (mix)
* Client Framework and version (Relay, Apollo, etc):  N/A getting compilation error

### Expected behavior

The schema compiles and the implementing sub-type is valid.

### Actual behavior

Compilation error

```
Type "header_cell" does not fully implement interface type "manifest_cell" for fields [:text]
...
  * An object field type is a valid sub-type if it is a Non-Null variant of a
    valid sub-type of the interface field type.
```

### Relevant Schema/Middleware Code

```
  interface :manifest_cell do
    field :text, :string
  end

  object :header_cell do
    field :text, non_null(:string)

    interface :manifest_cell
    is_type_of &is_struct(&1, HeaderCell)
  end
```

PR #1009, opened 2020-12-25
Ensure interface nullable types can be implemented by non-nullable types

maartenvanvliet:
Fixes https://github.com/absinthe-graphql/absinthe/issues/987

Two cases are added to the schema:
 * a simple nullable interface -> non-nullable implementor.
 * a nullable list -> with implementors `list_of(non_null(type))` and
   `non_null(list_of(non_null(type)))`.

From my reading of the spec these should all validate.

https://spec.graphql.org/draft/#IsValidImplementationFieldType()

I did find that it is not validated whether the object implements
the field's arguments as specified by the interface. Also that the object
does not implement extra non-nullable arguments that are not specified in the interface.

From the spec
```
field must include an argument of the same name for every argument defined in implementedField.

    That named argument on field must accept the same type (invariant) as that named argument on implementedField.

field may include additional arguments not defined in implementedField, but any additional argument must not be required, e.g. must not be of a non‐nullable type.
```

I'll create an issue for this.

### DIFF
--- a/lib/absinthe/phase/schema/validation/object_must_implement_interfaces.ex
+++ b/lib/absinthe/phase/schema/validation/object_must_implement_interfaces.ex
@@ -164,6 +164,15 @@ defmodule Absinthe.Phase.Schema.Validation.ObjectMustImplementInterfaces do
     check_covariant(inner_type1, inner_type2, field_ident, types)
   end
 
+  defp check_covariant(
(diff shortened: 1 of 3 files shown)


benwilson512 (approved):
Great, thanks! Good point about arguments, definitely an area we should improve.

Issue #1003, opened 2020-12-20
Support TypeSystem Directives


binaryseed:
The latest GraphQL spec includes `TypeSystemDirective` (directives that annotate the type system) alongside the `ExecutableDirective` (directives that annotate query execution)

https://spec.graphql.org/draft/#TypeSystemDirectiveLocation

Absinthe has full support for `ExecutableDirective`s, but only partial support for `TypeSystemDirective`s

kdawgwilk:
Can I get more details on what support is missing for `TypeSystemDirective`s?  I am looking at diving into this issue and would love any guidance.

binaryseed:
One part of this (the SDL rendering) is accomplished in this PR: https://github.com/absinthe-graphql/absinthe/pull/1010

I've got more working, and will submit additional PRs.

Unfortunately, I don't have much guidance since I'm figuring this out as I go, and there's probably a bit of design work with the rest of the core team to get the rest of this in place.

binaryseed:
One example of folks expecting Type System Directives to be available inside the resolver function: https://github.com/absinthe-graphql/absinthe/issues/964

binaryseed:
FYI, there's a comment in ^^ that includes a bit of a workaround that would enable passing directive details via `meta` using `hydrate`

PR #1010, opened 2020-12-26
Type System Directives SDL rendering

binaryseed:
This PR is part of #1003 

This PR ensures that directives applied to types (ie: Type System Directives) are correctly rendered in SDL output

After this, we can add support for these directives actually showing up in the `Absinthe.Type.*` structs


lib/absinthe/phase/schema/apply_declaration.ex
```diff
@@ -44,7 +44,8 @@ defmodule Absinthe.Phase.Schema.ApplyDeclaration do
         %{
           schema_definition
           | type_definitions:
-              Enum.map(schema_definition.type_definitions, &maybe_mark_root(&1, root_mappings))
+              Enum.map(schema_definition.type_definitions, &maybe_mark_root(&1, root_mappings)),
+            schema_declaration: declaration
```
binaryseed, line 48: Storing the `schema_declaration` so it can be used by SDL rendering - it might contain directives that we want to render

lib/absinthe/schema.ex
```diff
@@ -613,7 +613,7 @@ defmodule Absinthe.Schema do
   def to_sdl(schema) do
     pipeline =
       schema
-      |> Absinthe.Pipeline.for_schema()
+      |> Absinthe.Pipeline.for_schema(prototype_schema: schema.__absinthe_prototype_schema__)
```
binaryseed, line 616: This was needed to get the directive definitions in place when applied with a prototype schema

PR #1011, opened 2020-12-28
Fix docs malformed links and html tags

jeepers3327:
First time to request for pull request in a project, hopefully I didn't screw it up.

I've made some corrections on absinthe documentation page like malformed links and some tags was not properly rendered.

Changes:
* Change <a> tags with markdown way
* Replace html tags such as <strong> to markdown equivalent


benwilson512:
Thank you!

PR #1014, opened 2021-01-01
Add phase to reformat descriptions

binaryseed:
This PR brings back the logic to trim descriptions, but this time as a simple schema phase rather than part of the schema compilation mechanism. This behavior was lost in #1005.


benwilson512 (approved):
Great idea!

PR #1015, opened 2021-01-01
Fix introspection bug with import_fields and input

### LINKED ISSUE #989: Import fields includes introspect fields

### Environment

* Elixir version (elixir -v): 1.10.2
* Absinthe version (mix deps | grep absinthe): 1.5.3
* Client Framework and version (Relay, Apollo, etc): none, tested on graphiql

### Expected behavior

```
type AnObjectInput {
  d: String!
}
```

### Actual behavior

```
type AnObjectInput {
  __typename: String
  d: String!
}
```

__typename is present in the type

```

  object :an_object do
    @desc """
    Meeting id.
    """
    field :id, non_null(:id)
    import_fields(:c)
  end

  input_object :an_object_input do
    import_fields(:c)
  end

  object :c do
    field :d, non_null(:string)
  end
```

binaryseed:
This PR fixes a bug that would include `__typename` in an `input` if it called `import_fields` on an object.

Fixes #989

### DIFF
--- a/lib/absinthe/phase/schema/field_imports.ex
+++ b/lib/absinthe/phase/schema/field_imports.ex
@@ -33,7 +33,7 @@ defmodule Absinthe.Phase.Schema.FieldImports do
     Enum.reduce(type.imports, type, fn {source, opts}, type ->
       source_type = Map.fetch!(types, source)
 
-      rejections = Keyword.get(opts, :except, [])
+      rejections = Keyword.get(opts, :except, []) ++ exclusions(type)
 
       fields = source_type.fields |> Enum.reject(&(&1.identifier in rejections))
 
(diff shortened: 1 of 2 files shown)


lib/absinthe/phase/schema/field_imports.ex
```diff
@@ -29,11 +29,14 @@ defmodule Absinthe.Phase.Schema.FieldImports do
...
   ]
+  @exclude_fields [
+    :__typename
+  ]
   def import_fields(%def_type{} = type, types) when def_type in @can_import do
     Enum.reduce(type.imports, type, fn {source, opts}, type ->
       source_type = Map.fetch!(types, source)
 
-      rejections = Keyword.get(opts, :except, [])
+      rejections = Keyword.get(opts, :except, []) ++ @exclude_fields
```
binaryseed, line 39: I chose to always exclude `__typename` since the current behavior was actually causing it to be included twice for `object` as well as once for `input_object`

PR #1016, opened 2021-01-01
Maintenance fixes

maartenvanvliet:
I found a couple of small improvements in the code base

 * Remove unused requires
 * Remove unused variables that exposed bugs in type imports/assertion helpers
 * Remove unused files
 * Rename test file so it gets run


lib/absinthe/pipeline/error_result.ex
```diff
@@ -1,20 +0,0 @@
-defmodule Absinthe.Pipeline.ErrorResult do
```
binaryseed, line 1: I wonder if some users are in fact using this even if Absinthe internals aren't using it... It was part of the docs

lib/absinthe/phase/schema/type_imports.ex
```diff
@@ -37,13 +37,14 @@ defmodule Absinthe.Phase.Schema.TypeImports do
...
 
-    case Keyword.fetch(opts, :only) do
-      {:ok, selections} ->
-        Enum.filter(types, &(&1.identifier in selections))
-
-      _ ->
-        types
-    end
+    types =
+      case Keyword.fetch(opts, :only) do
```
binaryseed, line 41: No test covered this feature? It seems like this one was covering it:

https://github.com/absinthe-graphql/absinthe/blob/master/test/absinthe/schema/notation/experimental/import_fields_test.exs#L100

lib/absinthe/schema/rule.ex
```diff
@@ -1,40 +0,0 @@
-defmodule Absinthe.Schema.Rule do
```
binaryseed, line 1: @benwilson512 Was this leftover from before Schema phases?

binaryseed (commented):
Thanks!

lib/absinthe/schema/rule.ex
benwilson512: Yes, it can be safely deleted.

lib/absinthe/phase/schema/type_imports.ex
maartenvanvliet: This test https://github.com/absinthe-graphql/absinthe/blob/master/lib/absinthe/phase/schema/type_imports.ex covers the :only/:except logic for type imports, but the test only asserts that :only/:except is set correctly, not whether the actual import happens. I've added assertions that should check this.

lib/absinthe/pipeline/error_result.ex
maartenvanvliet: Happy to leave this in, but easy enough for users to put back if they depend on it.

lib/absinthe/phase/schema/type_imports.ex
maartenvanvliet: Also added some docs for this functionality and added import_types/2 to locals_without_parens

lib/absinthe/pipeline/error_result.ex
binaryseed: Let's move forward and address it if anyone reports any issues

lib/absinthe/phase/schema/type_imports.ex
binaryseed: Great!

PR #1017, opened 2021-01-01
Conform built in types to spec

maartenvanvliet:
Changes the built in types to conform to the spec, see https://spec.graphql.org/draft/#sec-Schema-Introspection
Most notably adds the `__TypeKind` enum. 

`Absinthe.Introspection.Kind.kind` now returns an atom for the `__TypeKind` enum resolver.

### DIFF
--- a/lib/absinthe/introspection/kind.ex
+++ b/lib/absinthe/introspection/kind.ex
@@ -8,19 +8,20 @@ defmodule Absinthe.Introspection.Kind do
         __MODULE__
         |> Module.split()
         |> List.last()
-        |> Absinthe.Introspection.Kind.upcase()
+        |> Absinthe.Introspection.Kind.downcase()
+        |> String.to_existing_atom()
       end
 
       defoverridable kind: 0
     end
   end
 
-  def upcase(name) do
+  def downcase(name) do
     Regex.scan(~r{[A-Z]+[a-z]+}, name)
     |> List.flatten()
-    |> Enum.map(&String.upcase/1)
+    |> Enum.map(&String.downcase/1)
     |> Enum.join("_")
   end
 
-  @callback kind :: binary
+  @callback kind :: atom
 end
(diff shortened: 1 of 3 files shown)


lib/absinthe/introspection/kind.ex
```diff
@@ -8,19 +8,20 @@ defmodule Absinthe.Introspection.Kind do
         __MODULE__
         |> Module.split()
         |> List.last()
-        |> Absinthe.Introspection.Kind.upcase()
+        |> Absinthe.Introspection.Kind.downcase()
```
binaryseed, line 11: Given this is enumerated now, and not just a string, I wonder if this can be simplified (ie: not use all that module / string manipulation)

It seems like the goal is just to have a function `kind` that returns an atom... We could get rid of the macro altogether:

```elixir
defmodule Absinthe.Type.Object do
  @behaviour Absinthe.Introspection.Kind
  def kind, do: :object
end

defmodule Absinthe.Introspection.Kind do
  @type introspection_kind :: :enum | :object # ...
  @callback kind() :: introspection_kind()
end
```

binaryseed (commented):
Thanks, this is a good improvement!

lib/absinthe/introspection/kind.ex
binaryseed: I think I'll do this separately...

binaryseed (approved):

lib/absinthe/introspection/kind.ex
binaryseed: Actually while making this change, I found some further issues with our `__TypeKind`!

Further corrections here: https://github.com/absinthe-graphql/absinthe/pull/1019

+ 1 to being explicit!

maartenvanvliet: Thx, good catch!

PR #1018, opened 2021-01-01
Clean up test warnings

binaryseed:
This tweak will silence an expected compiler warning that is polluting our test output. Also inlines a valid schema test instead of using the mechanism that exists for invalid schema tests.


binaryseed:
Removes these:

Screen Shot 2021-01-01 at 3 55 33 PM

PR #1019, opened 2021-01-02
Corrections on __TypeKind

binaryseed:
Followup to #1017 

* Our `__TypeKind` was defined with one more type (`SCHEMA`) than the GraphQL spec, which is now corrected
* `Absinthe.Introspection.TypeKind` is more explicit, and no longer relies on module name parsing to define itself
* Directives aren't a Kind


lib/absinthe/type/built_ins/introspection.ex
```diff
@@ -94,15 +94,14 @@ defmodule Absinthe.Type.BuiltIns.Introspection do
 
   enum :__type_kind,
     values: [
-      :schema,
```
binaryseed, line 97: Not a `__TypeKind`

lib/absinthe/type/directive.ex
```diff
@@ -8,7 +8,6 @@ defmodule Absinthe.Type.Directive do
 
   alias Absinthe.Type
   alias Absinthe.Language
-  use Absinthe.Introspection.Kind
```
binaryseed, line 11: Not a `__TypeKind`

Issue #937, opened 2020-05-28
import_types breaking change


fuelen:
### Environment

* Elixir version (elixir -v): 1.10
* Absinthe version (mix deps | grep absinthe): absinthe 1.5.1

### Expected behavior
```
import_types(MyApp.Schema.Types.{Users, Posts})
```
imports types from MyApp.Schema.Types.Users and from MyApp.Schema.Types.Posts.
Absinthe 1.4.16 works fine with this kind of import
### Actual behavior
```
== Compilation error in file lib/my_app/schema/schema.ex ==
** (ArgumentError) module Elixir.MyApp.Schema.Types.Users is not available
```
```
import_types(MyApp.Schema.Types.Users)
import_types(MyApp.Schema.Types.Posts)
```
^
works correctly

marctrem:
I have experienced the same regression.

juddey:
As have I

binaryseed:
Does anyone have a minimal reproduction case?

dylan-chong:
This seems to be working for me at the moment on 1.5.5

binaryseed:
This was tracked down by @maartenvanvliet in #1022 !

I've published `1.6.0-rc.0` for folks to try out while we wrap up a few last items before a full release..

hazardfn:
I am getting a similar variant to this issue in 1.6.3, my current setup works fine in 1.5.x:

```elixir
Could not load module `Elixir.ApolloWeb.Schema.PostTypes`. It returned reason: `unavailable`.
```

I know for sure the module exists... I am not using the tuple layout as above just straight importing the types one after the other

binaryseed:
Can you share more context? When does that happen? Can you generate a small reproduction case?

bjunc:
I can replicate this in our app (not sure about a small / sharable repo), upgrading from `1.5.3` to `1.6.3`

On a full build, I will get an "unavailable" module error like above; which has a union type. If I replace the union type with a simple object of the same name (just an `id` field), then it compiles.  Seems union related. 

Oddly, if I do the full build with the dummy object, then replace with the union for subsequent (app only) compilation, it'll work.

sheharyarn:
Can confirm, we're facing a similar issue. Just tried upgrading absinthe from `1.5.5` to `1.6.x` and now the schema fails to compile with the same message:


```
== Compilation error in file lib/myapp_graphql/schema.ex ==
** (Absinthe.Schema.Error) Compilation failed:
[...]

PR #1022, opened 2021-01-02
Fix import_types/2 for A.{B, C}

maartenvanvliet:
Fixes https://github.com/absinthe-graphql/absinthe/issues/937

I can reproduce issue #937 with a repo with absinthe as a dependency but not from within a test in absinthe. Has to do with the modules not being compiled (yet)

Changing https://github.com/absinthe-graphql/absinthe/blob/master/lib/absinthe/schema/notation.ex#L1683 to `if Code.ensure_compiled(type_module) do` fixed it. (See https://github.com/absinthe-graphql/absinthe/pull/534 for dealing with this in an earlier version).

Or, removing the if-clause entirely and relying on the TypeImports phase to handle this. This is what happens for ordinary imports like `import_types(MyApp.Schema.Types.UnknownModule)`. I've done this in this PR and let failed imports add a phase error to the schema. 

This could also be extended that whenever non absinthe modules are imported another error is added e.g.

```elixir
defmodule Test do
end
defmodule Schema do
   use Absinthe.Schema
   import_types Test
end
```
The TypeImports phase could check  `function_exported(Test, :__absinthe_blueprint__, 0)` and add an error if it's not present.

### DIFF
--- a/lib/absinthe/phase/schema/type_imports.ex
+++ b/lib/absinthe/phase/schema/type_imports.ex
@@ -17,7 +17,9 @@ defmodule Absinthe.Phase.Schema.TypeImports do
     {Absinthe.Type.BuiltIns.Introspection, []}
   ]
   def handle_imports(%Schema.SchemaDefinition{} = schema) do
-    types = do_imports(@default_imports ++ schema.imports, schema.type_definitions)
+    {types, schema} =
+      do_imports(@default_imports ++ schema.imports, schema.type_definitions, schema)
+
     # special casing the import of the built in directives
     [builtins] = Absinthe.Type.BuiltIns.Directives.__absinthe_blueprint__().schema_definitions
     directives = schema.directive_definitions ++ builtins.directive_definitions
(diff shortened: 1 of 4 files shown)


lib/absinthe/phase/schema/type_imports.ex
```diff
@@ -26,25 +28,43 @@ defmodule Absinthe.Phase.Schema.TypeImports do
...
     end
+  end
 
-    do_imports(other_def.imports ++ rest, types ++ acc)
+  # Generate an error when loading module fails
+  @spec error(module :: module(), error :: :embedded | :badfile | :nofile | :on_load_failure) ::
+          Absinthe.Phase.Error.t()
+  defp error(module, reason) do
+    %Absinthe.Phase.Error{
+      message: "Could not load module `#{module}`. It returned reason: `#{reason}`.",
```
binaryseed, line 66: If you `inspect(module)` it'll get rid of the `Elixir.` prefix in the output...

binaryseed:
This is great!

PR #1023, opened 2021-01-03
Support dynamic Enum value

binaryseed:
This PR adds support for dynamic Enum values in macro based schemas using `unqoute` like in #1005

### DIFF
--- a/lib/absinthe/blueprint/schema.ex
+++ b/lib/absinthe/blueprint/schema.ex
@@ -145,7 +145,7 @@ defmodule Absinthe.Blueprint.Schema do
   end
 
   defp build_types([{:values, values} | rest], [enum | stack], buff) do
-    enum = Map.update!(enum, :values, &(values ++ &1))
+    enum = Map.update!(enum, :values, &(List.wrap(values) ++ &1))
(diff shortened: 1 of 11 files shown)


lib/absinthe/blueprint/schema.ex
```diff
@@ -145,7 +145,7 @@ defmodule Absinthe.Blueprint.Schema do
   end
 
   defp build_types([{:values, values} | rest], [enum | stack], buff) do
-    enum = Map.update!(enum, :values, &(values ++ &1))
+    enum = Map.update!(enum, :values, &(List.wrap(values) ++ &1))
```
benwilson512, line 148: What is the need to wrap these in a list?

binaryseed: Ahh, yes this is the sneaky bit. Now that `values` are `wrap_in_unquote`'d it can now be a list of atoms _or_ AST for a function call - so I just wrap everything in a list herer so that it can continue to collect in a list.

Then in the end after this all is injected into `__absinthe_blueprint__` it'll be some form of nested list, and I can call `List.flatten(type_def.values)` to get the final resulting list....

PR #1027, opened 2021-01-08
Copy logger metadata for async and batch

dnsbty:
Hi folks! Thanks for creating and maintaining an awesome library. I debated opening an issue for this first, but decided it's a simple enough change that I would just create a PR. I'm happy to rework it or scrap it if that's what's decided.

Currently whenever the async or batch middleware is used, any logger metadata is lost when the new async tasks are created. This means that the Phoenix request ID doesn't show in log lines if using Phoenix, and it means that a user ID or other metadata needs to be added in every resolver to be present for logs.

This copies the logger metadata from the parent process whenever a new async task is created so that any logs further down the stack will have access to it.

Thanks for reviewing! I look forward to any feedback 😄

### DIFF
--- a/config/test.exs
+++ b/config/test.exs
@@ -1,3 +1,5 @@
 use Mix.Config
 
 config :logger, level: :info
+
+config :logger, :console, metadata: [:request_id]
--- a/lib/absinthe/middleware/async.ex
+++ b/lib/absinthe/middleware/async.ex
@@ -50,8 +50,17 @@ defmodule Absinthe.Middleware.Async do
   # This function inserts additional middleware into the remaining middleware
   # stack for this field. On the next resolution pass, we need to `Task.await` the
   # task so we have actual data. Thus, we prepend this module to the middleware stack.
-  def call(%{state: :unresolved} = res, {fun, opts}) when is_function(fun),
-    do: call(res, {Task.async(fun), opts})
+  def call(%{state: :unresolved} = res, {fun, opts}) when is_function(fun) do
+    metadata = Logger.metadata()
+
+    task =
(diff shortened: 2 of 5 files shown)


binaryseed:
Hello there! Thanks for this, though I'm not quite sure that this is something we want to do...

I understand the need, but I'm not sure that manually copying the metadata into the right solution. I'd hope we have a more systematic way to manage context propagation across processes similarly across the Elixir/Erlang ecosystem, I know this is something that folks working on `telemetry` and OpenTelemetry have been thinking about.

@benwilson512 You have any thoughts on this?

FYI, a quick copy/pasta of these files & you can pretty easily run a modified version in your codebase given that they are just plugins in the first place...

hauleth:
One thing, if Absinthe would prefer to bump minimal supported version to 1.10, then it would be better to use `:logger.get_process_metadata/0` and `:logger.set_process_metadata/1` as it would do not need roundtrip `map -> keyword -> map` which is happening in `Logger.metadata/0` due to compatibility reasons.

benwilson512:
We have generally used Phoenix's Elixir support as a guide, and they still support 1.9

binaryseed:
I don't think this is something we should do in Absinthe. I am advocating for a general context propagation solution in `erlang`

I also realized that you can do this much easier yourself.. the Async middleware can take in a task you've defined yourself, so you can take this example and tweak it just a bit:

https://github.com/absinthe-graphql/absinthe/blob/d83a1ff439f9d52575e34940df492528738d038a/test/absinthe/middleware/async_test.exs#L41-L50

PR #1028, opened 2021-01-11
Allow directives on variable variable_definition_test

maartenvanvliet:
See https://github.com/graphql/graphql-spec/pull/510

Adds support for directives on variable definitions in the parser as is defined by the spec.


binaryseed:
Good find!

PR #1029, opened 2021-01-12
Update builtin types according to spec

maartenvanvliet:
Found some mismatches between the builtin types and the spec (https://spec.graphql.org/draft/#sec-Schema-Introspection), this PR fixes those.

Also adds the schema description to SDL rendering and introspection, see graphql/graphql-spec#466


benwilson512 (approved):
Excellent work as always, thank you!

PR #1030, opened 2021-01-15
Make tests async where possible

maartenvanvliet:
Some tests could be marked as async to speed up the test suite a little :)

In Elixir master it reports the async and sync test durations.
```
Finished in 11.5 seconds (8.6s async, 2.9s sync)
```


benwilson512:
Thank you!

PR #1032, opened 2021-01-16
Render null default values in SDL

maartenvanvliet:
Literal null default values in arguments in the SDL notation triggered errors.

 * A pattern match was missing for the `Language.NullValue` when converting to `Blueprint.Schema.InputValueDefinition`
 * In the SDL rendered the `Blueprint.Input.Null` was not matched when in render_value


binaryseed:
Thanks!

PR #1033, opened 2021-01-20
Improved serialization failure messages

benwilson512:
This changes the relatively unhelpful message we have today:

```
** (Absinthe.SerializationError) Value 1.0 is not a valid integer
```

into something that actually tells you what field to go look for to fix the issue:

```
** (Absinthe.SerializationError) Could not serialize term 1.0 as type Int
When serializing the field:
RootQueryType.bad_integer (/path/to/schema/module.ex:8)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.6.1
+
+- Feature: [Improved serialization failure messages](https://github.com/absinthe-graphql/absinthe/pull/1033)
+
 ## 1.6.0
 
 - Feature: [Interfaces can now implement Interfaces](https://github.com/absinthe-graphql/absinthe/pull/1012), matching the latest spec
--- a/lib/absinthe/phase/document/result.ex
+++ b/lib/absinthe/phase/document/result.ex
@@ -56,9 +56,9 @@ defmodule Absinthe.Phase.Document.Result do
           rescue
             Absinthe.SerializationError ->
               raise(Absinthe.SerializationError, """
-              Could not serialize: #{inspect(value)} as a #{schema_node.name}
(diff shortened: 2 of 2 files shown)


lib/absinthe/phase/document/result.ex
```diff
@@ -51,7 +51,19 @@ defmodule Absinthe.Phase.Document.Result do
     value =
       case Type.unwrap(emitter.schema_node.type) do
         %Type.Scalar{} = schema_node ->
-          Type.Scalar.serialize(schema_node, value)
+          try do
+            Type.Scalar.serialize(schema_node, value)
+          rescue
+            Absinthe.SerializationError ->
```
benwilson512, line 57: I'm not in love with how raising the choice to have serializers succeed or raise puts us in this fun position of having to raise / rescue / raise the same exception but I don't really want to go down the rabbit hole of changing how all of this works right now, and what we're doing here to improve the error message doesn't really box us in for any future changes.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.6.1
+
+- Feature: [Improved serialization failure messages](https://github.com/absinthe-graphql/absinthe/pull/1033)
```
benwilson512, line 5: @binaryseed I'm somewhat inclined to push for PRs to add their own changelog entries when we think one is applicable. This ensures they actually happen and involves the PR author in the description of the change.

binaryseed (approved):

CHANGELOG.md
binaryseed: The only tricky bit is the lack of a PR # before submitting :)

benwilson512: That is true, I had to add mine as a commit after opening the PR. Still, that doesn't seem like too painful of a requirement.

PR #1037, opened 2021-02-01
Hibernate Absinthe.Schema.Manager

binaryseed:
I noticed that the `Absinthe.Schema.Manager` process was using a lot of memory in our app, but it shouldn't since it doesn't do anything. I dug around and realized that it never gets GC'd since the process `init`s and then never does anything again, and therefore it never gets checked for memory usage.

This PR fixes that by having it `:hibernate` immediately which triggers a GC.


binaryseed:
For our large schema, this was ~60mb!

benwilson512:
Great catch!

Issue #1038, opened 2021-02-02
Slow compile times in a large app due to full schema recompilation since upgrading from 1.4


dmarkow:
I have a fairly large API using Absinthe (25,000 lines of code, around 400 objects, 300 queries, and 300 mutations). I've noticed that it's getting slower and slower to compile my schema, and since just about every module in my app is reachable from one of my resolvers, touching basically any file in the project results in a full recompile of my schema, even if none of my fields/queries changed. Compiling the schema adds around 15 seconds to every recompilation, making frequent updates and test runs painful.

I spent some time digging through and determined that most of the time is spent in [phase/schema/compile.ex](https://github.com/absinthe-graphql/absinthe/blob/master/lib/absinthe/phase/schema/compile.ex#L70), specifically at the `Module.create(...)` call, which takes around 10 seconds to generate the `ApiWeb.Schema.Compiled` module.

It seems to make no difference if I keep everything in one giant schema file versus splitting it by my contexts and using import_types/import_fields.  I found https://devonestes.herokuapp.com/trying-to-speed-up-compilation-with-absinthe but even trying the umbrella app experiment outlined there, my compile times only improved about 10% (and added a lot of complexity).

Out of curiosity I reverted back to absinthe 1.4.16 and everything is **way faster**. The `schema.ex` file is still recompiled, but it happens in around 750ms instead of 15 seconds. The bump up to 1.5 is where the recompilation issues started.

Is this just a downside to the convenience of the new features added since 1.5, or is there something that can be done either on Absinthe's end or my own to help with this?

### Environment

* Elixir version (elixir -v): Elixir 1.11.3 (compiled with Erlang/OTP 23)
* Absinthe version (mix deps | grep absinthe): 1.6.1 (Hex package) (mix)

### Expected behavior

Changes that don't actually affect my schema to not trigger a costly recompile.

### Actual behavior

Touching any file referenced by a resolver triggers a complete recompile of my schema.

### Relevant Schema/Middleware Code

Basic schema with many fields, queries, and mutations. No extra plugins/middleware.

benwilson512:
Hi @dmarkow, can you try using the persistent term backend found here: https://hexdocs.pm/absinthe/1.5.5/Absinthe.Schema.PersistentTerm.html ?

dmarkow:
[...]

PR #1039, opened 2021-02-02
Clarify placement of @schema_provider

dmarkow:
See #1038 -- `@schema_provider Absinthe.Schema.PersistentTerm` needs to be added after the `use Absinthe.Schema` line.


benwilson512:
Thanks!

PR #1045, opened 2021-02-10
Add failing test for macro deprecated field in SDL render

maartenvanvliet:
A field that is deprecated in the macro defined schema's is not rendered as deprecated in SDL. See this failing test case. This is a regression from Absinthe 1.5.5

In 1352d8fd6e520dabae3cd642c69ec663065d829c the deprecation directive was no longer special-cased, which works for SDL defined schema's, but not in macro defined schema's. The difference being that the macro schema's don't add a `deprecated` directive  to`Absinthe.Blueprint.Schema.FieldDefinition.directives`, so it only relies on the `FieldDefinition.deprecation` field. 

SDL schema's directives do add the `deprecated` directive to the `directives` field, so they are rendered correctly in SDL. They rely on [this directive expansion](https://github.com/absinthe-graphql/absinthe/blob/v1.6.3/lib/absinthe/schema/prototype/notation.ex#L47) to set the deprecation field. 

We could either special case the `deprecated` directive again in the SDL renderer, or use the macro schema to add a `deprecated` directive when deprecating a field (perhaps through a schema phase)

### DIFF
--- a/test/absinthe/schema/sdl_render_test.exs
+++ b/test/absinthe/schema/sdl_render_test.exs
@@ -214,7 +214,7 @@ defmodule Absinthe.Schema.SdlRenderTest do
       field :id, :id
       field :name, :string
       field :status, :order_status
-      field :other_status, :status
+      field :other_status, :status, deprecate: "See status field"
       import_fields :imported_fields
     end
 
@@ -275,7 +275,7 @@ defmodule Absinthe.Schema.SdlRenderTest do
                id: ID
                name: String
                status: OrderStatus
-               otherStatus: Status
+               otherStatus: Status @deprecated(reason: "See status field")
              }
              """
   end


binaryseed:
Thanks, good find. It seems like adding the directive is moving in the right direction...

aerosol:
I started looking at this, by it's quite difficult to figure out what needs to be done for someone unfamiliar with the internals. Is the idea to introduce a Phase that'd extend relevant nodes' `directives` with `Directive` instances in case the `deprecation` value is present? How would one go about instantiating the `Directive` struct given the blueprint input? It's also unclear to me how to provide the argument (reason) for later expansion there.

zoldar:
FWIW here's the workaround we have used to re-add directives back at the time https://gist.github.com/zoldar/d617f2037854ccfc26cd428e11df6ed7

chakeresa:
We are also seeing this issue (upgrading from 1.5.5 to 1.6.5 removed all the deprecation warnings from our introspected schema). Are there any updates on fixing this bug? I am hesitant to update, for this reason :-/.

ymtszw:
@binaryseed Hi, let me clarify the situation so that I could tackle on the issue:

* Currently Absinthe DOES support "defining" directives in macro schema, but DOES NOT support "calling/applying/using" defined directives as per https://elixirforum.com/t/how-to-apply-directives-on-fields-in-absinthe/20089/2
  Is this still correct?
* If the above holds, the most "proper" way to solve this issue is:
  * (1) introduce the whole directive mechanisms to support "applying" directives to various GQL locations as per https://spec.graphql.org/draft/#DirectiveLocation 
  * and (2) render SDLs based on actually applied directives in our schema
  * then (3) with (2), `@deprecated` and maybe-future-coming other "default directives" (such as `@skip`) are mere directives inside, so they are rendered alongside with user-defined directives
* "The whole directive mechanisms" is obviously a significant undertaking. My gut feeling says for now we should special-case `@deprecated` and use `:deprecations` in SDL rendering. What do you think on this?
  * Since deprecation mechanism is crucial for incremental type-safe development, I foresee many users will benefit from this.
  * For example, in our workflow we utilize [GraphQL Inspector](https://graphql-inspector.com/) for detecting deprecations/breaking changes, so properly rendering `@deprecated` in SDL greatly contributes to our type-based confidence!

---

BTW, I have found that `@deprecated` on args and input fields were rather new and still not merged to the official spec, to my surprise.
https://github.com/graphql/graphql-spec/pull/805

benwilson512:
@ymtszw that forum post is out of date. Absinthe does support calling directives in SDL.

see: https://github.com/absinthe-graphql/absinthe/blob/master/test/absinthe/schema/notation/experimental/import_sdl_test.exs#L567-L592

ymtszw:
I would still really love to tackle on this but couldn't have taken my time. Maybe soon...

FWIW our use case:

* We are auto-generating `schema.graphql` and committing it to git
* On GitHub Actions we call [graphql-inspector](https://github.com/kamilkisiela/graphql-inspector) to detect intentional/unintentional schema changes
* graphql-inspector can report removals of non-deprecated types and fields as errors

For this workflow, we compare `schema.graphql` in a PR branch against one from its base branch. But due to the issue, it cannot track deprecation => removal flow.
The inspector can be manually suppressed using predefined labels (such as `approved-breaking-changes`) so it isn't a show-stopper. But proper detection of removal-after-deprecation is definitely an improvement.

ymtszw:
Related attempt: https://github.com/absinthe-graphql/absinthe/pull/1110

maartenvanvliet:
I've created a PR to apply directives at the schema which also covers deprecations. This would also cover the `@deprecated` directive

chakeresa:
Awesome! Here’s the PR if anyone else is curious: https://github.com/absinthe-graphql/absinthe/pull/1140

PR #1050, opened 2021-03-21
Skip duplicates Github actions run

dolfinus:
Hello.

It looks like commiting to branches excluding `master` does not run Github Actions. So users which are working on feature/bugfix branches cannot see that their changes are leading to failed tests until they'll create a pull request to master branch.
So I've removed restriction against branch name for `push` events.

But there is an annoying fact about Github Actions - sometimes it runs actions twice. For example:
1. You just pushed a commit to a branch - testing workflow will run only once
2. Someone forked your repo and creates a pull request - same thing
3. You've created a new branch (feature, bugfix or release branch), pushed a commit here and also created a pull request to the main branch - two copies of workflow will be run

Running a workflow twice does not make any sense, it just takes twice the time without any useful side effect.

So in this PR I've also added a separated job of checking if another copy of workflow already started: https://github.com/fkirc/skip-duplicate-actions

This job allows to:
1. Skip the second copy of workflow in the case 3 (branch in original repo + PR to main branch). Cases 1 and 2 will work just the same as now
2. Skip workflow if someone changed only files which are not connected to package code, like README or CHANGELOG
3. Cancel already started workflow in the same PR or branch if someone pushed new commits to it.

Also I've found another quite annoying bug - step `actions/cache` does not update cache item if it was restored using a primary key.
For example, some dependencies like dialyzer or other ones store their files in `deps` or `_build` folders (which are cached in the workflow). But when these dependencies perform some updates of these folders, it does not lead to the updating of `mix.lock` file. File hasn't been changed -> hash is the same -> cache primary key is the same -> caching action will not update the existing cache.
This will cause the increase of workflow run time because the stored cache will not be used by some steps. Until `mix.lock` will be updated which will lead to updating the cache item.
Here I've appended a commit SHA to the cache primary key, so it will always be updated. But it will not affect cache reading.

What do you think about that?

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -2,16 +2,32 @@ name: CI
 
 on:
   push:
-    branches:
-      - master
   pull_request:
     branches:
       - master
 
 jobs:
+  check_duplicate_runs:
+    name: Check for duplicate runs
+    continue-on-error: true
+    runs-on: ubuntu-latest
+    outputs:
+      should_skip: ${{ steps.skip_check.outputs.should_skip }}
+    steps:
+      - id: skip_check
+        uses: fkirc/skip-duplicate-actions@master
+        with:
+          concurrent_skipping: always
+          cancel_others: true
+          skip_after_successful_duplicate: true
+          paths_ignore: '["**/README.md", "**/CHANGELOG.md", "**/CODE_OF_CONDUCT.md", "**/CONTRIBUTING.md", "**/LICENSE.md"]'
+          do_not_skip: '["pull_request"]'
+
   test:
     name: Elixir ${{matrix.elixir}} / OTP ${{matrix.otp}}
     runs-on: ubuntu-latest
+    needs: check_duplicate_runs
+    if: ${{ needs.check_duplicate_runs.outputs.should_skip != 'true' }}
 
     strategy:
       matrix:
@@ -34,44 +50,47 @@ jobs:
     - name: Restore deps cache
       uses: actions/cache@v2
       with:
-        path: deps
-        key: deps-${{ matrix.otp }}-${{ matrix.elixir }}-${{ hashFiles('**/mix.lock') }}
+        path: |
+          deps
+          _build
+        key: ${{ runner.os }}-deps-${{ matrix.otp }}-${{ matrix.elixir }}-${{ hashFiles('**/mix.lock') }}-git-${{ github.sha }}
         restore-keys: |
-          deps-${{ matrix.otp }}-${{ matrix.elixir }}
+          ${{ runner.os }}-deps-${{ matrix.otp }}-${{ matrix.elixir }}-${{ hashFiles('**/mix.lock') }}
+          ${{ runner.os }}-deps-${{ matrix.otp }}-${{ matrix.elixir }}
 
-    - name: Restore _build cache
-      uses: actions/cache@v2
-      with:
-        path: _build
-        key: _build-${{ matrix.otp }}-${{ matrix.elixir }}-${{ hashFiles('**/mix.lock') }}
-        restore-keys: |
-          _build-${{ matrix.otp }}-${{ matrix.elixir }}
+    - name: Create dializer plts path
+      run: mkdir -p priv/plts || true
 
     - name: Restore plts cache
       uses: actions/cache@v2
       with:
         path: priv/plts
-        key: plts-${{ matrix.otp }}-${{ matrix.elixir }}
+        key: ${{ runner.os }}-plts-${{ matrix.otp }}-${{ matrix.elixir }}-${{ hashFiles('**/mix.lock') }}-git-${{ github.sha }}
+        restore-keys: |
+          ${{ runner.os }}-plts-${{ matrix.otp }}-${{ matrix.elixir }}-${{ hashFiles('**/mix.lock') }}
+          ${{ runner.os }}-plts-${{ matrix.otp }}-${{ matrix.elixir }}
 
-    - name: Install deps
+    - name: Install package dependencies
       run: mix deps.get
+
+    - name: Remove compiled application files
+      run: mix clean
+
+    - name: Compile dependencies
+      run: mix compile
+      env:
+        MIX_ENV: test
 
     - name: Check Formatting
       run: mix format --check-formatted
 
     - name: Run unit tests
-      run: |
-        mix clean
-        mix test
+      run: mix test
 
     - name: Run unit tests with persistent_term backend
-      run: |
-        mix clean
-        mix test
+      run: mix test
       env:
         SCHEMA_PROVIDER: persistent_term
 
     - name: Run dialyzer
-      run: |
-        mkdir -p priv/plts
-        MIX_ENV=test mix dialyzer
+      run: mix dialyzer
--- a/mix.exs
+++ b/mix.exs
@@ -14,6 +14,9 @@ defmodule Absinthe.Mixfile do
       start_permanent: Mix.env() == :prod,
       package: package(),
       source_url: @source_url,
+      preferred_cli_env: [
+        dialyzer: :test
+      ],
       docs: [
         source_ref: "v#{@version}",
         main: "overview",


.github/workflows/elixir.yml
```diff
@@ -34,44 +50,51 @@ jobs:
     - name: Restore deps cache
       uses: actions/cache@v2
       with:
-        path: deps
-        key: deps-${{ matrix.otp }}-${{ matrix.elixir }}-${{ hashFiles('**/mix.lock') }}
+        path: |
+          deps
+          _build
+        key: ${{ runner.os }}-deps-${{ matrix.otp }}-${{ matrix.elixir }}-${{ hashFiles('**/mix.lock') }}-git-${{ github.sha }}
```
binaryseed, line 56: Seems like the `sha` is sub-ideal here, it's _always_ different.

I wonder if we should `hashFiles('deps/*')`

.github/workflows/elixir.yml
```diff
@@ -34,44 +50,51 @@ jobs:
     - name: Restore deps cache
       uses: actions/cache@v2
       with:
-        path: deps
-        key: deps-${{ matrix.otp }}-${{ matrix.elixir }}-${{ hashFiles('**/mix.lock') }}
+        path: |
+          deps
+          _build
```
binaryseed, line 55: I'd rather not combine these, `deps` changes infrequently, whereas `_build` will change with every single code change.

.github/workflows/elixir.yml
```diff
@@ -34,44 +50,51 @@ jobs:
...
       with:
         path: priv/plts
-        key: plts-${{ matrix.otp }}-${{ matrix.elixir }}
+        key: ${{ runner.os }}-plts-${{ matrix.otp }}-${{ matrix.elixir }}-${{ hashFiles('**/mix.lock') }}-git-${{ github.sha }}
+        restore-keys: |
+          ${{ runner.os }}-plts-${{ matrix.otp }}-${{ matrix.elixir }}-${{ hashFiles('**/mix.lock') }}
+          ${{ runner.os }}-plts-${{ matrix.otp }}-${{ matrix.elixir }}
+
+    - name: Install hex
+      run: mix local.hex --force
```
binaryseed, line 74: These are done by default in the `erlef/setup-elixir` action:

https://github.com/erlef/setup-elixir/blob/main/action.yml#L12-L17

.github/workflows/elixir.yml
```diff
@@ -34,44 +50,51 @@ jobs:
...
-        mix test
+      run: mix test
       env:
         SCHEMA_PROVIDER: persistent_term
 
     - name: Run dialyzer
-      run: |
-        mkdir -p priv/plts
-        MIX_ENV=test mix dialyzer
+      run: mix dialyzer
```
binaryseed, line 100: `MIX_ENV=test` was there to avoid a re-compile of the application

.github/workflows/elixir.yml
```diff
@@ -34,44 +50,51 @@ jobs:
...
+      run: mix local.rebar --force
+
+    - name: Install package dependencies
       run: mix deps.get
 
+    - name: Remove compiled application files
+      run: mix clean
+
+    - name: Compile dependencies
+      run: mix compile
```
binaryseed, line 86: We shouldn't need to compile the application in `dev` at all

mix.exs
```diff
@@ -14,6 +14,9 @@ defmodule Absinthe.Mixfile do
       start_permanent: Mix.env() == :prod,
       package: package(),
       source_url: @source_url,
+      preferred_cli_env: [
```
binaryseed, line 17: This isn't sufficient to avoid compiling the entire app again in `dev` since it needs to compile to _run_ the `mix dialyzer` task itself

binaryseed:
So I haven't found the need for running CI on all pushes / skipping branches to be honest, I can run the tests locally pretty trivially, so I don't depend on that. What is the problem that led to you needing this part?

The issues you point out about the cache are interesting, looks like there's some improvements that can be made there for sure...

.github/workflows/elixir.yml
dolfinus: `t's always different.`
And that's the main point here.

When `action/cache` is looking up for cache, it will firstly try `key`, then if nothing was found I'll try all the items from `restore-keys` until something will be found.
Also if it downloaded cache by primary key (`key`), it will not upload the updated files to cache storage.

If you have only `key` without `restore-keys`, `action/cache` will download it but will never update. In this case every update of mix.lock will lead to generating new primary key, and thus using completely new cache. But it can be a very long operation for large projects.

If you have `key: {{otp}}-{{elixir}}-{{hash(mix.lock)}}` and `restore-keys: {{otp}}-{{elixir}}-`, it will work faster because even you've upgraded mix.lock, and file hash was changed, 'action/cache` will download the latest cache item which correspond to the same OTP and Elixir versions. Mix will left intact dependencies which were not changed, and will update only those which were updated.

But sometimes there are steps which change your cached files but not changing the mix.lock hash. It can be a `dialyzer` run (it stores `.plts` files in `deps` subfolder), or some other tool which saves data to `deps` or `_build` folders.
In this case `action/cache` decides that because cache was fetched by primary key (mix.lock wasn't changed), there is no need to update the existing cache in the cache storage. But without updating this cache dialyzer and other tools like that will rebuild their files every time, and sometimes it takes a lot of time.

So you basically have a cache, but because it works not as expected, it will increase time of every next CI job run. Until `mix.lock` file will be updated which leads to updating the cache.

That's why I propose to add `github.sha` into the primary key here - it will always be new, so `action/cache` step will update cache every time. And because there are items in `restore-keys`, this cache will still be downloaded on next commits, so it will not be disabled as you may expect.

You cannot put `hash(**/_build)` into the key because this directory does not exist in the commit.

.github/workflows/elixir.yml
dolfinus: `deps` contains the source code of dependencies. `_build` contains the compiled `.beam` files for these dependencies as well as for your project. And the project compiles files are usually just the small part of `_build` folder content. So there is no actual reason to store these files separately - updating files in `deps` will lead to updating the `_build` folder.

.github/workflows/elixir.yml
dolfinus: Thank you, I didn't know that.

.github/workflows/elixir.yml
dolfinus: I've set up default environment for dialyzer task in the mix.exs file, that's why I removed this line.

.github/workflows/elixir.yml
dolfinus: Well, fair enough.

mix.exs
dolfinus: No, running just mix itself does not lead to starting the compilation process. It firstly reads mix.exs file content, then decides that should be done (evaluating aliases, setting up proper environment and paths for the compiler), and only after that runs the desired action.

Because `dialyzer` step now have default `test` environment, mix will firstly resolve all configs and paths for this specific environment, then run `dialyzer` itself, which then run `compile` step internally. You can check CI job on my fork to see that there is no additional compilation step performed (https://github.com/dolfinus/absinthe/runs/2160694599?check_suite_focus=true, see `Run dialyzer` step).

dolfinus:
You can run tests locally, there is no doubt, but users who are forking the repo may be not familiar with it. 
For example, my case:
1. I want to fix some issue in the absinthe repo
2. I create a fork of the repo
3. Then I realize that there are 2 different issues here
4. So instead of pushing commits to a master branch of my fork, I create branches named first_issie and second_issue
5. But because there is restriction on the workflow, test will not be run on these branches until I create a pull request to a master
6. If you decide to remove this restriction, you could face the running copies of the same workflow then pushing commits to branches which are used as a source for some pull request (I already faced this issue in other repos, e.g. https://github.com/karolsluszniak/ex_check/pull/26). But it just takes more time to wait until CI will finish without any useful result. So I propose to add a skip test to avoid such a behavior.

.github/workflows/elixir.yml
binaryseed: But code updates in absinthe itself (which is far more frequent) change `_build` which would re-cache the `deps` w/o needing too

dolfinus: Yes, but saving cache takes less than 2 seconds even for such a large project like absinthe.

binaryseed:
I'm going to pull this into a feature branch and make some modifications...
Thanks!

mix.exs
binaryseed: hmm, looks like that's correct - i tried it locally before but must have done something wrong...

.github/workflows/elixir.yml
```diff
@@ -34,44 +50,47 @@ jobs:
...
+        restore-keys: |
+          ${{ runner.os }}-plts-${{ matrix.otp }}-${{ matrix.elixir }}-${{ hashFiles('**/mix.lock') }}
+          ${{ runner.os }}-plts-${{ matrix.otp }}-${{ matrix.elixir }}
 
-    - name: Install deps
+    - name: Install package dependencies
       run: mix deps.get
 
+    - name: Remove compiled application files
+      run: mix clean
```
binaryseed, line 77: The clean step only existed because we run the tests in 2 modes (`SCHEMA_PROVIDER=persistent_term`) so we need to keep that inline before each test

dolfinus: Ok, got you

PR #1056, opened 2021-03-26
Better solve deadlocks on Elixir v1.12+

### LINKED ISSUE #937: import_types breaking change

josevalim:
Closes #937.

### DIFF
--- a/lib/absinthe/phase/schema/type_imports.ex
+++ b/lib/absinthe/phase/schema/type_imports.ex
@@ -63,6 +63,7 @@ defmodule Absinthe.Phase.Schema.TypeImports do
   # This gives the Elixir compiler more information to address
   # deadlocks.
   # TODO: Remove the else clause once we require Elixir v1.12+.
+  @compile {:no_warn_undefined, {Code, :ensure_compiled!, 1}}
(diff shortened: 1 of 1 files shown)


josevalim:
There is a dialyzer warning because the function is not available in certain Elixir versions so we need an annotation somewhere to address it - but I am not sure which one. If someone can address that, it would be very appreciated.

binaryseed:
Thanks! I'll see if I can get the warnings resolved

PR #1057, opened 2021-03-26
Better solve deadlocks on Elixir v1.12+ 

binaryseed:
From https://github.com/absinthe-graphql/absinthe/pull/1056 with a `dialyzer` fix


benwilson512 (approved):
This is good. Hopefully we can obviate this by moving more things to persistent term and runtime calls, but this is good regardless.

I've actually _finally_ reproduced this locally, so I'll be able to test that runtime oriented approaches fix the issue.

PR #1058, opened 2021-03-29
Compress registry tables by default

josevalim:
@benwilson512 sent a PR to Elixir to support compressed
tables. This enables it by default on the Registry.

This is just a proposal, perhaps you want to make it opt-in
instead.


benwilson512:
I think this is a good default, and anyone operating at a scale where the speed hit from the compression matters is probably going to OOM without this anyway.

PR #1062, opened 2021-04-15
Fix string serialization error raising

emeryotopalik:
During serialization of the result, String.Chars.to_string raises the following error when given a map:
```
** (Protocol.UndefinedError) protocol String.Chars not implemented for %{} of type Map. This protocol is implemented for the following type(s): Decimal, Float, DateTime, Time, List, Version.Requirement, Atom, Integer, Version, Date, BitString, NaiveDateTime, URI
```
This trickles up to the end user when instead it should handle the serialization error more gracefully like the other scalars.

### DIFF
--- a/lib/absinthe/phase/document/result.ex
+++ b/lib/absinthe/phase/document/result.ex
@@ -54,15 +54,18 @@ defmodule Absinthe.Phase.Document.Result do
           try do
             Type.Scalar.serialize(schema_node, value)
           rescue
-            Absinthe.SerializationError ->
-              raise(Absinthe.SerializationError, """
-              Could not serialize term #{inspect(value)} as type #{schema_node.name}
+            _e in [Absinthe.SerializationError, Protocol.UndefinedError] ->
+              raise(
+                Absinthe.SerializationError,
+                """
+                Could not serialize term #{inspect(value)} as type #{schema_node.name}
 
-              When serializing the field:
-              #{emitter.parent_type.name}.#{emitter.schema_node.name} (#{
-                emitter.schema_node.__reference__.location.file
-              }:#{emitter.schema_node.__reference__.location.line})
-              """)
+                When serializing the field:
+                #{emitter.parent_type.name}.#{emitter.schema_node.name} (#{
+                  emitter.schema_node.__reference__.location.file
+                }:#{emitter.schema_node.__reference__.location.line})
+                """
+              )
           end
 
         %Type.Enum{} = schema_node ->
--- a/lib/absinthe/type/built_ins/scalars.ex
+++ b/lib/absinthe/type/built_ins/scalars.ex
@@ -54,8 +54,16 @@ defmodule Absinthe.Type.BuiltIns.Scalars do
     represent free-form human-readable text.
     """
 
-    serialize &String.Chars.to_string/1
+    serialize &__MODULE__.serialize_string/1
     parse parse_with([Absinthe.Blueprint.Input.String], &parse_string/1)
+  end
+
+  def serialize_string(n) when not is_map(n), do: String.Chars.to_string(n)
+
+  def serialize_string(n) do
+    raise Absinthe.SerializationError, """
+    Value #{inspect(n)} is not a valid string
+    """
   end
 
   scalar :id, name: "ID" do
--- a/lib/absinthe/type/built_ins/scalars.ex
+++ b/lib/absinthe/type/built_ins/scalars.ex
@@ -54,16 +54,8 @@ defmodule Absinthe.Type.BuiltIns.Scalars do
     represent free-form human-readable text.
     """
 
-    serialize &__MODULE__.serialize_string/1
+    serialize &String.Chars.to_string/1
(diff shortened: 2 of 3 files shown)


lib/absinthe/type/built_ins/scalars.ex
```diff
@@ -54,10 +54,18 @@ defmodule Absinthe.Type.BuiltIns.Scalars do
     represent free-form human-readable text.
     """
 
-    serialize &String.Chars.to_string/1
+    serialize &__MODULE__.serialize_string/1
     parse parse_with([Absinthe.Blueprint.Input.String], &parse_string/1)
   end
 
+  def serialize_string(n) when not is_map(n), do: String.Chars.to_string(n)
```
binaryseed, line 61: Hmm, I wonder what the ideal way to approach this is.. I suspect we want something more robust than a not-map check...

Lots of things can `to_string`, but they don't all make sense

```elixir
% [1, 2, 3] |> to_string
<<1, 2, 3>>

% ["one", "two"] |> to_string
"onetwo"

% [:foo, :bar] |> to_string
** (ArgumentError) cannot convert the given list to a string.
```

The question is how strict should we get w/o causing a breaking change..

There's a minimal note in the spec about coercing basic values into strings for results (not inputs)

http://spec.graphql.org/draft/#sec-String.Result-Coercion

> GraphQL services may coerce non-string raw values to String when reasonable without losing information, otherwise they must raise a field error. Examples of this may include returning the string "true" for a boolean true value, or the string "1" for the integer 1.

cc @benwilson512

benwilson512: It's a protocol, so there is no guard clause we can use. Instead we need to basically:

```
def try_to_string(value) do
    String.Chars.to_string(value)
  rescue
    ProtocolUndefinedError ->
          raise Absinthe.SerializationError, """
    Value #{inspect(n)} is not a valid string
    """
end
```

There's no good way to know if you can make a value into a string without trying. (note, I'm just guessing on the exception name, but you get the idea).

benwilson512: @binaryseed your note that you can make non utf8 binaries with `to_string` is a bit of an issue though, that's unfortunate. We can't tolerate the performance hit of calling `String.valid?` on every binary though, so there are limits to the guarantees Absinthe can providehere.

emeryotopalik: Yeah, I don't think it makes sense to coerce `[]` to `""`, for instance.

benwilson512: > Yeah, I don't think it makes sense to coerce [] to "", for instance.

This isn't really up to us though. I don't believe Absinthe should implement its own `to_string` concept different than the core language, it's going to be harder to maintain. This is the first concrete issue anyone has raised in the years that Absinthe has been around. Re-raising with the Absinthe serialization error is a good, non breaking change that normalizes the error API. Changing what we allow to be returned from string resolvers goes too far.

benwilson512: So, on further thought, in an Absinthe 2.0 I could see an argument for requiring that the `:string` scalar only accept binaries. That _still_ doesn't mean it can be serialized (they could be invalid binaries) but we could reasonably force people to do `to_string` in their resolvers if they want to return `[]` or `1` and have it be coerced. This would require a fair bit more discussion though, and would definitely have to be held off until a 2.0 because it would break existing resolvers for people.

benwilson512: @binaryseed we could maybe start warning, but I think changing the types we accept is too much of a breaking change for a patch release. It also is an issue because it breaks _reasonable_ protocol expectations eg: `{:ok, %Decimal{}}` or `{:ok, DateTime.utc_now()}`.

binaryseed: Fair enough, let's go with the `rescue Protocol.UndefinedError` for now!

lib/absinthe/phase/document/result.ex
```diff
@@ -58,6 +58,16 @@ defmodule Absinthe.Phase.Document.Result do
               raise(Absinthe.SerializationError, """
               Could not serialize term #{inspect(value)} as type #{schema_node.name}
 
+              When serializing the field:
```
emeryotopalik, line 61: I'm going to put this in a variable...

test/absinthe/integration/execution/serialization_test.exs
```diff
@@ -45,4 +49,13 @@ defmodule Absinthe.Integration.Execution.SerializationTest do
       Absinthe.run(@query, Schema)
     end)
   end
+
+  @query """
+  query { badString }
+  """
+  test "returning not a string for a string raises" do
```
binaryseed, line 56:
```suggestion
  test "returning a type that can't `to_string` for a string raises" do
```

binaryseed (approved):

binaryseed:
Thanks!

Issue #1063, opened 2021-04-21
returning {:error, nil} in resolve/3


chungwong:
For a mutation

```elixir
    payload field(:post_subscription) do
      input do
        field(:email, non_null(:string))
      end
        
      output do
        field(:result, :boolean)
      end
        
      resolve(&subscribe/3)
    end

```

If `subscribe/3` returns `{:error, nil}`, the response is
```json
{
  "data": {
    "postSubscription": null
  }
}
```
If `subscribe/3` returns `{:error, "I am an error"}`, the response is
```json
{
  "data": {
    "postSubscription": null
  },
  "errors": [
    {
      "locations": [
        {
          "column": 3,
          "line": 4
        }
      ],
      "message": "I am an error",
      "path": [
        "postSubscription"
      ]
    }
  ]
}
```

### Environment

* Elixir version (elixir -v): 1.11
* Absinthe version (mix deps | grep absinthe): 1.5

### Expected behavior

It should return an error resposne
### Actual behavior

It is returning a successful response

benwilson512:
@chungwong I do wonder what the expected behaviour should be here. Absinthe does not allow for example `{:error, message: nil}`. We generally treat `{:error, message}` as shorthand for `{:error, message: message}`.

I agree that Absinthe's current response is wrong. However, my inclination is that `{:error nil}` should simply raise an error.

chungwong:
according to the error message of resolving
```elixir
## For an error result

One or more errors for a field can be returned in a single `{:error, error_value}` tuple.

`error_value` can be:
- A simple error message string.
- A map containing `:message` key, plus any additional serializable metadata.
- A keyword list containing a `:message` key, plus any additional serializable metadata.
- A list containing multiple of any/all of these.
- Any other value compatible with `to_string/1`.

### Examples

A simple error message:

    {:error, "Something bad happened"}

Multiple error messages:

    {:error, ["Something bad", "Even worse"]

Single custom errors (note the required `:message` keys):

    {:error, message: "Unknown user", code: 21}
    {:error, %{message: "A database error occurred", details: format_db_error(some_value)}}

Three errors of mixed types:

    {:error, ["Simple message", [message: "A keyword list error", code: 1], %{message: "A map error"}]}

Generic handler for interoperability with errors from other libraries:

    {:error, :foo}
    {:error, 1.0}
    {:error, 2}

## To activate a plugin

`{:plugin, NameOfPluginModule, term}` to activate a plugin.
```

[...]

PR #1068, opened 2021-05-06
Returns an error when `{:error, nil}` is given to resolve/3 (#1063)

cheerfulstoic:
Works in the same way that `{:error, ""}` would

I realize that maybe there isn't a conclusion on #1063 but I though I would at least hunt down the place where it's happening and make a test.  It should be easy enough to change the test if different behavior is desired


benwilson512:
I think I am going to close this and say that we should really just remove the `anything that to_string supports` and say that it should just be a string.

PR #1072, opened 2021-05-14
[WIP] failing case for non null list of non null error handling

benwilson512:


jtomaszewski:
Hey, any status on this? I see you've made a repro, but no fix? If we'd make a fix for it (so that the tests pass), would you merge it then?

benwilson512:
Yes I have not had time to look into this further, you're welcome to take a crack at it. Null propagation happens in https://github.com/absinthe-graphql/absinthe/blob/master/lib/absinthe/phase/document/execution/resolution.ex#L43-L75

benwilson512:
This has been handled by #1223

PR #1073, opened 2021-05-19
Document  behavior on Absinthe.Type.Object

Nickforall:
This documents the behavior of `is_type_of`. I couldn't find how to implement it aside from its tests.


benwilson512:
Great, thanks!

PR #1074, opened 2021-05-20
otp 24 and elixir 1.12

benwilson512:

### DIFF
--- a/lib/absinthe/phase/schema/type_imports.ex
+++ b/lib/absinthe/phase/schema/type_imports.ex
@@ -67,7 +67,12 @@ defmodule Absinthe.Phase.Schema.TypeImports do
   @dialyzer {:nowarn_function, [ensure_compiled: 1]}
   defp ensure_compiled(module) do
     if function_exported?(Code, :ensure_compiled!, 1) do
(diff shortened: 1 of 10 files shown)


benwilson512:
I need to figure out how to get the formatter check to only run on elixir v1.12 since it is different now.

binaryseed:
You have access to `matrix.elixir` that you could interpolate into the `mix format` command

lib/absinthe/phase/schema/type_imports.ex
```diff
@@ -67,7 +67,12 @@ defmodule Absinthe.Phase.Schema.TypeImports do
   @dialyzer {:nowarn_function, [ensure_compiled: 1]}
   defp ensure_compiled(module) do
     if function_exported?(Code, :ensure_compiled!, 1) do
-      {:module, Code.ensure_compiled!(module)}
+      try do
+        {:module, Code.ensure_compiled!(module)}
```
josevalim, line 71: You should not rescue ensure_compiled!, as it will throw the compiler on a hoop. My suggestion is for you to have “ensure_compiled!” When you need the module and have a non raising one and call them accoedingly.

Issue #1077, opened 2021-06-02
key :is_type_of not found on Interface


maennchen:
If submitting a bug, please provide the following:

### Environment

* Elixir version (elixir -v):

```
Erlang/OTP 24 [erts-12.0] [source] [64-bit] [smp:6:6] [ds:6:6:10] [async-threads:1] [jit]

Elixir 1.12.0 (compiled with Erlang/OTP 22)
```

* Absinthe version (mix deps | grep absinthe):

```
* absinthe 1.6.4 (Hex package) (mix)
  locked at 1.6.4 (absinthe) 6e9c1cf3
* absinthe_error_payload 1.0.1 (Hex package) (mix)
  locked at 1.0.1 (absinthe_error_payload) 03b30013
* absinthe_phoenix 2.0.1 (Hex package) (mix)
  locked at 2.0.1 (absinthe_phoenix) e773adc8
* absinthe_plug 1.5.8 (Hex package) (mix)
  locked at 1.5.8 (absinthe_plug) bbb04176
* absinthe_relay 1.5.1 (Hex package) (mix)
  locked at 1.5.1 (absinthe_relay) 1fd2a355
```

* Client Framework and version (Relay, Apollo, etc): None yet

### Expected behavior

When defining *interface* root / * interface* intermediate / *object* child and defining a `is_type_of` on the object, everything should work fine.

### Actual behavior

```
[...]

PR #1078, opened 2021-06-02
Prevent key :is_type_of not found error on interfaces

maennchen:
Fix for https://github.com/absinthe-graphql/absinthe/issues/1077

Based on #1079, merge that one first.

### DIFF
--- a/lib/absinthe/type/interface.ex
+++ b/lib/absinthe/type/interface.ex
@@ -86,8 +86,6 @@ defmodule Absinthe.Type.Interface do
   def resolve_type(type, obj, env, opts \\ [lookup: true])
 
   def resolve_type(interface, obj, %{schema: schema} = env, opts) do
-    implementors = Schema.implementors(schema, interface.identifier)
-
     if resolver = Type.function(interface, :resolve_type) do
       case resolver.(obj, env) do
         nil ->
@@ -102,7 +100,8 @@ defmodule Absinthe.Type.Interface do
       end
     else
       type_name =
-        Enum.find(implementors, fn type ->
+        Schema.implementors(schema, interface.identifier)
+        |> Enum.find(fn type ->
           Absinthe.Type.function(type, :is_type_of).(obj)
         end)
 
--- a/lib/absinthe/type/interface.ex
+++ b/lib/absinthe/type/interface.ex
@@ -102,7 +102,9 @@ defmodule Absinthe.Type.Interface do
       end
     else
       type_name =
-        Enum.find(implementors, fn type ->
+        implementors
+        |> Enum.filter(&match?(%Type.Object{}, &1))
+        |> Enum.find(fn type ->
           Absinthe.Type.function(type, :is_type_of).(obj)
         end)
 
--- a/lib/absinthe/type/interface.ex
+++ b/lib/absinthe/type/interface.ex
@@ -102,9 +102,7 @@ defmodule Absinthe.Type.Interface do
       end
     else
       type_name =
-        implementors
-        |> Enum.filter(&match?(%Type.Object{}, &1))
-        |> Enum.find(fn type ->
+        Enum.find(implementors, fn type ->
           Absinthe.Type.function(type, :is_type_of).(obj)
         end)
 
--- a/lib/absinthe/blueprint/schema/interface_type_definition.ex
+++ b/lib/absinthe/blueprint/schema/interface_type_definition.ex
@@ -49,10 +49,8 @@ defmodule Absinthe.Blueprint.Schema.InterfaceTypeDefinition do
     }
   end
 
-  @interface_types [Schema.ObjectTypeDefinition, Schema.InterfaceTypeDefinition]
-
   def find_implementors(iface, type_defs) do
-    for %struct{} = obj when struct in @interface_types <- type_defs,
+    for %Schema.ObjectTypeDefinition{} = obj <- type_defs,
         iface.identifier in obj.interfaces,
         do: obj.identifier
   end
--- a/test/absinthe/schema/rule/object_must_implement_interfaces_test.exs
+++ b/test/absinthe/schema/rule/object_must_implement_interfaces_test.exs
@@ -99,7 +99,7 @@ defmodule Absinthe.Schema.Rule.ObjectMustImplementInterfacesTest do
     assert %{
              named: [:cat, :dog, :user],
              favorite_foods: [:cat, :dog, :user],
-             parented: [:cat, :dog, :named, :user]
+             parented: [:cat, :dog, :user]
            } ==
              Schema.__absinthe_interface_implementors__()
   end
@@ -117,7 +117,7 @@ defmodule Absinthe.Schema.Rule.ObjectMustImplementInterfacesTest do
       url: String
     }
 
-    interface Image implements Resource & Node {
+    type Image implements Resource & Node {
       id: ID!
       url: String
       thumbnail: String
@@ -125,14 +125,21 @@ defmodule Absinthe.Schema.Rule.ObjectMustImplementInterfacesTest do
 
     """
 
+    def hydrate(%Absinthe.Blueprint.Schema.InterfaceTypeDefinition{}, _) do
+      {:resolve_type, &__MODULE__.resolve_type/1}
+    end
+
+    def hydrate(_node, _ancestors), do: []
+
+    def resolve_type(_), do: false
+
     query do
     end
   end
 
   test "interfaces are set from sdl" do
     assert %{
-             image: [],
-             node: [:image, :resource],
+             node: [:image],
              resource: [:image]
            } ==
              InterfaceImplementsInterfaces.__absinthe_interface_implementors__()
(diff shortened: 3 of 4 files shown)


benwilson512:
Can you include a test case that would fail on master?

maennchen:
@benwilson512 Where was the current functionality of the `is_type_of` tested?

benwilson512:
test/absinthe/type/interface_test.exs mostly

lib/absinthe/type/interface.ex
```diff
@@ -102,7 +102,9 @@ defmodule Absinthe.Type.Interface do
       end
     else
       type_name =
-        Enum.find(implementors, fn type ->
+        implementors
```
benwilson512, line 105: I think it's actually this list which is wrong. The implementors list should be only concrete implementors. This change should occur in lib/absinthe/blueprint/schema/interface_type_definition.ex `find_implementors/1`

benwilson512 (changes requested):

lib/absinthe/type/interface.ex
benwilson512: This way the filtering is accomplished at schema building time instead of every time the interface is evaluated at runtime.

maennchen:
@benwilson512 I added a test.

lib/absinthe/type/interface.ex
maennchen: I'm running out of time now, I'll nave a look later 👍

maennchen: @benwilson512 If I change it in `finde_implementors`, I break those two tests:

```
  1) test interfaces are propagated across type imports (Absinthe.Schema.Rule.ObjectMustImplementInterfacesTest)
     test/absinthe/schema/rule/object_must_implement_interfaces_test.exs:88
     Assertion with == failed
     code:  assert %{named: [:cat, :dog, :user], favorite_foods: [:cat, :dog, :user], parented: [:named]} == Schema.__absinthe_interface_implementors__()
     left:  %{favorite_foods: [:cat, :dog, :user], named: [:cat, :dog, :user], parented: [:named]}
     right: %{favorite_foods: [:cat, :dog, :user], named: [:cat, :dog, :user], parented: []}
     stacktrace:
       test/absinthe/schema/rule/object_must_implement_interfaces_test.exs:89: (test)

.

  2) test interfaces are set from sdl (Absinthe.Schema.Rule.ObjectMustImplementInterfacesTest)
     test/absinthe/schema/rule/object_must_implement_interfaces_test.exs:122
     Assertion with == failed
     code:  assert %{image: [], node: [:image, :resource], resource: [:image]} == InterfaceImplementsInterfaces.__absinthe_interface_implementors__()
     left:  %{image: [], node: [:image, :resource], resource: [:image]}
     right: %{image: [], node: [], resource: []}
     stacktrace:
       test/absinthe/schema/rule/object_must_implement_interfaces_test.exs:123: (test)
```

Do we consider those tests correct or wrong?

maennchen: @benwilson512 In which direction should I continue?

maennchen:
@benwilson512 Ping :)

maennchen:
@benwilson512 Iw ould really like to finish this PR. Can you give me some guidance on what to do to get it over the line?

lib/absinthe/type/interface.ex
maennchen: @benwilson512 I need some guidance to finish this PR.

arturs678:
Any updates on this ?

lib/absinthe/type/interface.ex
tlvenn: I just had a look at those 2 tests and I believe the schema for those is wrong. According to the [spec](https://spec.graphql.org/draft/#sel-GAHbhBHABmBBoBz3W) and my understanding of the following sentence:

>Transitively implemented interfaces (interfaces implemented by the interface that is being implemented) must also be defined on an implementing type or interface.

I believe that both `dog` & `cat` objects should declare the `parented` interface as well.

maartenvanvliet: yeah, I have a fix locally for #1080 and those schemas came up as being invalid. Just need to clean it all up for a PR for that issue

tlvenn: @maennchen I have merged @maartenvanvliet [PR](https://github.com/absinthe-graphql/absinthe/issues/1080) which fixed those 2 tests. Could you give it another try when you get the chance and let us know how it goes now ? Thanks in advance.

maennchen: @tlvenn I've rebased my commits and reverted my implementation changes. The tests are failing.

maennchen: @tlvenn 

> I believe the schema for those is wrong

What would you specifically change in the test?

tlvenn: The schema used by the test has been fixed so that `dog` `cat` & `user` are declaring the `parented` interface as they should be.

tlvenn: And the tests themselves have been updated to assert the right thing as well.

tlvenn: If you rebase on master your PR I believe the tests should work now.

maennchen: @tlvenn Can you please specify exactly what you mean I should change in my test? I can't follow what you think should be different.

I did rebase on master. It does not pass.

tlvenn: Interesting... Can you update your PR with the rebase and can you copy paste the results of those 2 tests please ?

tlvenn: Ha i see that you already pushed the rebase my bad, can you just update in the comment the results of those 2 tests ? Thanks !

tlvenn: Actually after you pushed your rebased PR, only the new test that is correctly failing now remains but the fix in `interface_type_definition` is gone or you never actually pushed it.

maennchen: @tlvenn I removed the fix and only left the test, since I thought it should already be fixed on master.

I'll reapply the fix then. :)

tlvenn: Sorry for the confusion, btw, It seems you pushed your initial fix and not the one targeting lib/absinthe/blueprint/schema/interface_type_definition.ex `find_implementors/1`

maennchen: @tlvenn Now I finally get it. :smile: 

I'll try that way then

maennchen: @tlvenn 

https://github.com/absinthe-graphql/absinthe/runs/4335029279?check_suite_focus=true

```
  1) test interfaces are set from sdl (Absinthe.Schema.Rule.ObjectMustImplementInterfacesTest)
Error:      test/absinthe/schema/rule/object_must_implement_interfaces_test.exs:132
     Assertion with == failed
     code:  assert %{image: [], node: [:image, :resource], resource: [:image]} == InterfaceImplementsInterfaces.__absinthe_interface_implementors__()
     left:  %{image: [], node: [:image, :resource], resource: [:image]}
     right: %{image: [], node: [], resource: []}
     stacktrace:
       test/absinthe/schema/rule/object_must_implement_interfaces_test.exs:133: (test)

.

  2) test interfaces are propagated across type imports (Absinthe.Schema.Rule.ObjectMustImplementInterfacesTest)
Error:      test/absinthe/schema/rule/object_must_implement_interfaces_test.exs:98
     Assertion with == failed
     code:  assert %{named: [:cat, :dog, :user], favorite_foods: [:cat, :dog, :user], parented: [:cat, :dog, :named, :user]} == Schema.__absinthe_interface_implementors__()
     left:  %{favorite_foods: [:cat, :dog, :user], named: [:cat, :dog, :user], parented: [:cat, :dog, :named, :user]}
     right: %{
              favorite_foods: [:cat, :dog, :user],
              named: [:cat, :dog, :user],
              parented: [:cat, :dog, :user]
            }
     stacktrace:
       test/absinthe/schema/rule/object_must_implement_interfaces_test.exs:99: (test)
```

You seem to know that code better. Should I just give you push access to my fork?

tlvenn: Ha great, thanks for this, I wanted to believe it would work but kinda suspected this could happen. Those 2 issues have been introduced by the original PR that added support for interface implementing interfaces because until now `find_implementors/1` was incorrectly returning interfaces instead of just concrete types.

Given that:
- This test suite is purposely for testing objects that use interfaces.
- We have other tests against interfaces implementing interfaces in `test/absinthe/schema/rule/object_interfaces_must_be_valid_test.exs`.
- Transitively implemented interfaces  must also be defined on implementing types.

We should simply get rid of interfaces implementing interfaces in those 2 tests and fix the assertions, we are mixing too many concerns otherwise and the purpose of those tests is lost.

We probably could / should introduce a specific test to assert that `find_implementors/1` is only returning concrete types in another place.

I am totally happy to take it from there if you want, totally up to you and in any case, thanks again for your time and effort on this. Have a great WE !

tlvenn: @maennchen can you please re send  the invitation to collab on your fork, i did not see it initially and it has expired now, thanks.

maennchen: @tlvenn I sent it again :)

maennchen:
@tlvenn Thanks :)

PR #1079, opened 2021-06-02
Fix Mix Format

maennchen:
Should make CI pass again


benwilson512:
Thanks! Formatting is going to get interesting since 1.12 made changes but this gets things inline properly for 1.11.

PR #1085, opened 2021-07-07
Fix union render with no UnionMemberTypes

### LINKED ISSUE #1084: SDL output for union with no types is invalid
### Environment

* Elixir version (elixir -v): 1.10 or 1.11
* Absinthe version (mix deps | grep absinthe): 1.6.4 or master branch

### Expected behavior

```graphql
union _Entity
```

### Actual behavior

```graphql
union _Entity = 
```

In a schema processing phase, I am dynamically creating this union type and in some cases the `types` list is empty but when I ask `Absinthe.Schema.to_sdl(MyApp.Schema)` I get `=` in the output with no types on the other side.  Tools like GraphiQL playground dont like this and throw errors

```elixir
%UnionTypeDefinition{
      __reference__: Notation.build_reference(__ENV__),
      description: "a union of all types that use the @key directive",
      identifier: :_entity,
      module: __MODULE__,
      name: "_Entity",
      types: []
    }
```

kdawgwilk:
Adds a unit test for a case that wasnt being tested for union with no types and fixes the issue.  Spec says that the UnionMemberTypes are optional https://spec.graphql.org/draft/#UnionTypeDefinition

Resolves #1084

### DIFF
--- a/lib/absinthe/schema/notation/sdl_render.ex
+++ b/lib/absinthe/schema/notation/sdl_render.ex
@@ -150,25 +150,33 @@ defmodule Absinthe.Schema.Notation.SDL.Render do
   end
 
   defp render(%Blueprint.Schema.UnionTypeDefinition{} = union_type, type_definitions) do
-    types =
-      Enum.map(union_type.types, fn
-        identifier when is_atom(identifier) ->
-          render(%Blueprint.TypeReference.Identifier{id: identifier}, type_definitions)
+    Enum.map(union_type.types, fn
+      identifier when is_atom(identifier) ->
+        render(%Blueprint.TypeReference.Identifier{id: identifier}, type_definitions)
 
-        %Blueprint.TypeReference.Name{} = ref ->
-          render(ref, type_definitions)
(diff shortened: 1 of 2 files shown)


MrGrinst (approved):

binaryseed:
Hmm, further down in the spec, I see this:

> A Union type must include one or more unique member types.

https://spec.graphql.org/draft/#sec-Unions.Type-Validation

kdawgwilk:
Interesting, I may be able to just not produce this type if there are no `UnionMemberTypes` but not sure what the right call here is

kdawgwilk:
If we want to go that direction we should prob add validation for that so that those unions won't be allowed in absinthe if they have no `NamedType`s

binaryseed:
I asked around about this, and the SDL syntax can define a "partial" type which can then be `extend`ed later... But eventually it needs to have `NamedType`s to be valid...

benwilson512:
Yeah I think that's the right call @binaryseed. I think this is good to go then?

PR #1086, opened 2021-07-08
Account for prototype schema when generating SDL

kdawgwilk:
The `Absinthe.Schema.to_sdl` function was taking this into account but the mix task was missing it see
https://github.com/absinthe-graphql/absinthe/blob/1948513d2103ac68182b64c62c0bf9fc35b96180/lib/absinthe/schema.ex#L616


benwilson512:
Good catch, thanks!

Issue #783, opened 2019-09-15
Eliminate bare atom identifiers


binaryseed:
> We should do a pass in the absinthe code base and eliminate all bare atom identifiers within blueprints and standardize on `%TypeReference.Name{} | %TypeReference.Identifier{}`

Noted in https://github.com/absinthe-graphql/absinthe/pull/780/files#r322457359

This will help us handle both Macro-based & SDL-based schemas better

PR #1091, opened 2021-07-30
Allow interface fields to reference other interfaces (#1089)

### LINKED ISSUE #1089: Possible difference in schema validation between DSL and import_sdl
The basic questions I'm trying to answer are:

1. Is there really is a discrepancy in schema validation between using import_sdl vs the DSL when using interfaces in a specific way?
2. If so, which one is right?

I provided a diff with a couple sample test cases in #1088 with details as well as a sample schema that should be representative of the "real one" we're working with internally.

### Env
Elixir 1.12.1
PR is using Absinthe master

### Expected behavior

It seems like the two modes should agree, but this is a weird case and I'll fully admit it's a bit of a tangle.

### Actual behavior

import_sdl will cause a schema validation failure, while expressing what appears to be the same schema using the DSL succeeds.

@jerelmiller — curious if my test case matches what you think you're seeing as well.

jerelmiller:
Closes #1089

It appears the DSL and the `import_sdl` differ in how interface types, with fields that themselves interfaces types, are compiled. When creating schemas via the DSL, compilation works as intended. I can create an interface that contains a field that is another interface type. When I do the same via `import_sdl`, compilation fails because it Absinthe thinks that the interface is not fully adhered to. This attempts to patch that behavior to ensure interface types can have fields that are themselves interfaces.

To ensure this use case is valid according to the GraphQL [spec](https://spec.graphql.org/June2018/#sec-Interface), I created a [minimal reproduction](https://github.com/jerelmiller/graphql-interface-reproduction) of this schema using GraphQL.js to validate whether the SDL or the DSL had the correct behavior. It does indeed appear that interface types are allowed as fields to other interface types.

### DIFF
--- a/lib/absinthe/phase/schema/validation/object_must_implement_interfaces.ex
+++ b/lib/absinthe/phase/schema/validation/object_must_implement_interfaces.ex
@@ -191,6 +191,18 @@ defmodule Absinthe.Phase.Schema.Validation.ObjectMustImplementInterfaces do
     :ok
   end
 
+  defp check_covariant(
+         %Blueprint.TypeReference.Name{name: iface_name},
+         %Blueprint.TypeReference.Name{name: type_name},
+         field_ident,
+         types
+       ) do
+    {_, itype} = Enum.find(types, fn {_, %{name: name}} -> name == iface_name end)
+    {_, type} = Enum.find(types, fn {_, %{name: name}} -> name == type_name end)
+
+    check_covariant(itype, type, field_ident, types)
+  end
+
   defp check_covariant(nil, _, field_ident, _), do: {:error, field_ident}
   defp check_covariant(_, nil, field_ident, _), do: {:error, field_ident}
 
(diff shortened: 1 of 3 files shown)


jerelmiller:
@binaryseed not sure if you're able to have a look, but I would love your input!

jpvajda:
It be cool to see this merged in soon! Looking forward to seeing this functionality.

lib/absinthe/phase/schema/validation/object_must_implement_interfaces.ex
```diff
@@ -191,6 +191,18 @@ defmodule Absinthe.Phase.Schema.Validation.ObjectMustImplementInterfaces do
     :ok
   end
 
+  defp check_covariant(
+         %Blueprint.TypeReference.Name{name: iface_name},
+         %Blueprint.TypeReference.Name{name: type_name},
```
binaryseed, line 196: Ahh, I suspect the difference between the SDL & DSL is that only SDL currently generates the `TypeReference` blueprint structs everywhere... DSL winds up with bare-atom identifiers

This is a related note from a while back: https://github.com/absinthe-graphql/absinthe/issues/783

binaryseed (approved):

lib/absinthe/phase/schema/validation/object_must_implement_interfaces.ex
jerelmiller: You're exactly right! Thats what I was seeing, just a bare atom in the case of the DSL.

mattbaker: ooh nice find

PR #1093, opened 2021-07-30
Filter log levels to prevent console bloating

kaaboaye:
## Problem I'm trying to fix

When having a lot of active subscriptions with a lot of messages server console becomes completely spammed by Absinthe debug logs.

The main offender is `Absinthe.Subscription.Local`.

https://github.com/absinthe-graphql/absinthe/blob/master/lib/absinthe/subscription/local.ex#L63-L68

Currently there is no way to disable Absinthe debug logs.

## Proposed solution

Enable `config :absinthe, log: boolean` to support log levels. As described below:

https://github.com/surferlocal/absinthe/blob/d353c593ab5d1b6b14ee96196d92664c09465f00/lib/absinthe/logger.ex#L40-L62

Additionally `absinthe_phoenix` would need to be updated after this path is released since subscription logging is spread between multiple apps.

https://github.com/surferlocal/absinthe_phoenix/commit/a374f9e54781f5de58c0fe48b0ebde1f9f2208e2#diff-3eb777c582eea111f7176ab1a3aa844fbe7fb49bcc3f17b0a6028e53be7eae64

Then it will become possible to set `config :absinthe, log: :info` and ignore any debug logs.

### DIFF
--- a/lib/absinthe/logger.ex
+++ b/lib/absinthe/logger.ex
@@ -37,7 +37,22 @@ defmodule Absinthe.Logger do
 
   The default is `#{inspect(@default_pipeline)}`.
 
-  ## Disabling
+  ## Log levels
+
+  ### Filtering
+
+  To filter out `:debug` level logs, set the `:log` configuration to `:info`
+  or any other level from `Logger.level`.
+
+      # logs info and above
+      config :absinthe,
+        log: :info
+
+      # only logs errors
+      config :absinthe,
+        log: :error
+
+  ### Disabling
 
   To disable Absinthe logging, set the `:log` configuration option to `false`:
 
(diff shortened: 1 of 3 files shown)


benwilson512:
Hi @kaaboaye. Newer version of Elixir actually do provide a way to deal with this without changes to Absinthe. Add the following to your config:

```
config :logger,
  compile_time_purge_matching: [
    [module: Absinthe.Subscription.Local]
  ]
```

Then make sure to recompile the dependency: ` mix deps.compile absinthe`.

Let me know if that doesn't work.

kaaboaye:
```ex
config :logger,
  compile_time_purge_matching: [
    [application: :absinthe, level_lower_than: :info]
  ]
```

It's perfect. Thanks

PR #1099, opened 2021-09-01
Use makeup_graphql to get GraphQL syntax highlighting in docs

Billzabob:
Docs before:
Screen Shot 2021-09-01 at 11 26 44 AM

Docs after:
Screen Shot 2021-09-01 at 11 26 04 AM


benwilson512:
Very cool! What remains to take this beyond draft?

Billzabob:
I'm using the master version of `ex_doc` in this because it doesn't work without the fix that @josevalim added [here](https://github.com/elixir-lang/ex_doc/commit/b25a0008f505258e42257d3c59af5ea7daaa5fae). So once there is a release for that I'll change this to actually use the latest version of `ex_doc` instead of using the `git` option and this will be good :)

josevalim:
0.25.2 is out :)

Billzabob:
This is good to go 👍

benwilson512:
Thank you!

PR #1104, opened 2021-09-24
Failing test for empty query sdl render bug

kdawgwilk:
Example for #1105

### DIFF
--- a/test/absinthe/schema/sdl_render_test.exs
+++ b/test/absinthe/schema/sdl_render_test.exs
@@ -285,4 +285,23 @@ defmodule Absinthe.Schema.SdlRenderTest do
              }
              """
   end
+
+  defmodule EmptyQueryTestSchema do
+    use Absinthe.Schema
+
+    query do
+    end
+  end
+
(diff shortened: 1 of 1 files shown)


binaryseed:
I don't think this is a valid schema - the root query type can't be empty. What did you do to encounter the inspect failure?

kdawgwilk:
Wouldn't the schema fail to compile in the tests if it was invalid?

binaryseed:
http://spec.graphql.org/draft/#sec-Objects.Type-Validation

> An Object type must define one or more fields

It might be that Absinthe doesn't currently have a compile time validation for this part of the spec

benwilson512:
@binaryseed I think this is related to that, but we may want to handle it a bit differently. There are certain schemas that are in valid in such a way where we can't meaningfully render them, but right now it tries to anyway if we `inspect`. We need some sort of fallback when inspecting because otherwise it is difficult to debug invalid schemas.

kdawgwilk:
Resolved by #1167

PR #1106, opened 2021-10-01
Remove unused Absinthe.Type.fetch

maartenvanvliet:
It publicly exposed `fetch/2` on a few Type modules
but it's not used internally, and don't expect it publicly either.


maartenvanvliet:
Tests are failing because `fetch/2` is a partial implementation of the access protocol. Closing for now

PR #1107, opened 2021-10-05
Add missing information in 1.5 changelogs

dorian-marchal:
Bumping from 1.4 to 1.6, I feel like this information was missing in changelogs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -93,6 +93,7 @@
 - Feature: Substantially lower subscription memory usage.
 - Documentation: Testing guide, numerous fixes and updates
 - Breaking Change: Scalar outputs are now type checked and will raise exceptions if the result tries to send the wrong data type in the result.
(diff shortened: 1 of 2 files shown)


benwilson512:
Hi Dorian, isn't that covered here: https://github.com/absinthe-graphql/absinthe/blob/master/CHANGELOG.md#v150-rc ?

dorian-marchal:
Hello, yes for the changelog, I didn't see it, but shouldn't we add this information in the upgrade guide, too?

benwilson512:
Yes, a comment about this in the upgrade guide is a good idea. Happy to re-open the PR with just that change pushed up.

dorian-marchal:
I did the change already but the PR didn't update (because it's closed, I
guess).

benwilson512:
Unfortunately github will not let me for some reason, can you simply open a new PR?

PR #1108, opened 2021-10-05
Rename root type names

kdawgwilk:
I realize this is a breaking change for some but seems to be going in the right direction, I am okay with waiting for this until we are ready to make breaking changes.  I ran into this difference while working on https://github.com/apollographql/apollo-federation-subgraph-compatibility/pull/35 thankfully you can override the name so I was able to.

https://spec.graphql.org/draft/#sec-Root-Operation-Types.Default-Root-Operation-Type-Names

## Default Root Operation Type Names

While any type can be the root operation type for a GraphQL operation, the type system definition language can omit the schema definition when the query, mutation, and subscription root types are named "Query", "Mutation", and "Subscription" respectively.

Likewise, when representing a GraphQL schema using the type system definition language, a schema definition should be omitted if it only uses the default root operation type names.

This example describes a valid complete GraphQL schema, despite not explicitly including a schema definition. The "Query" type is presumed to be the query root operation type of the schema.

```graphql
type Query {
  someField: String
}
```

### DIFF
--- a/lib/absinthe/phase/schema/introspection.ex
+++ b/lib/absinthe/phase/schema/introspection.ex
@@ -37,7 +37,7 @@ defmodule Absinthe.Phase.Schema.Introspection do
   def update_type_defs(type_defs) do
     for type_def = %struct_type{} <- type_defs do
       cond do
-        type_def.name in ["RootQueryType", "Query"] ->
+        type_def.name in ["Query", "Query"] ->
           type_field = field_def(:type)
           schema_field = field_def(:schema)
           typename_field = field_def(:typename)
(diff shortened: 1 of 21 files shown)


kdawgwilk:
Is this a change we want to make?  Or because we have the ability to override the type names is this unnecessary?

maartenvanvliet:
I think this is something for a 2.0 release. Anyone who needs to change the names right now can already do so but the `Query` etc. root types are a good new default.

Also, the SDL rendering should be adapted to omit the schema definition as mentioned in the spec you linked to.

benwilson512:
Yeah I'm gonna just mark this as something to look at in a v2 and close for now. Trying to get the open PR list under control.

PR #1110, opened 2021-10-08
Default values not rendered in the sdl

SvenW:
When specifying default values in input arguments they aren't rendered in the sdl, using `Absinthe.Schema.to_sdl`

### DIFF
--- a/lib/absinthe/schema/notation/sdl_render.ex
+++ b/lib/absinthe/schema/notation/sdl_render.ex
@@ -143,6 +143,7 @@ defmodule Absinthe.Schema.Notation.SDL.Render do
       directives(input_value.directives, type_definitions)
     ])
     |> description(input_value.description)
+    |> deprecation(input_value.deprecation)
   end
 
   defp render(%Blueprint.Schema.FieldDefinition{} = field, type_definitions) do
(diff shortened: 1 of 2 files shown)


binaryseed:
I think we should identify the source of the problem first... It seems like perhaps it's a difference between SDL & Macro schemas.

SDL schema does correctly render default values:
* https://github.com/absinthe-graphql/absinthe/blob/master/test/absinthe/schema/sdl_render_test.exs#L48

Macro schema doesn't render default values:
* https://github.com/absinthe-graphql/absinthe/blob/master/test/absinthe/schema/sdl_render_test.exs#L198
* https://github.com/absinthe-graphql/absinthe/blob/master/test/absinthe/schema/sdl_render_test.exs#L254

This is probably another situation where the SDL & Macro schemas generate different structs when they are compiled.

SvenW:
> I think we should identify the source of the problem first... It seems like perhaps it's a difference between SDL & Macro schemas.
> 
> SDL schema does correctly render default values:
> 
> * https://github.com/absinthe-graphql/absinthe/blob/master/test/absinthe/schema/sdl_render_test.exs#L48
> 
> Macro schema doesn't render default values:
> 
> * https://github.com/absinthe-graphql/absinthe/blob/master/test/absinthe/schema/sdl_render_test.exs#L198
> * https://github.com/absinthe-graphql/absinthe/blob/master/test/absinthe/schema/sdl_render_test.exs#L254
> 
> This is probably another situation where the SDL & Macro schemas generate different structs when they are compiled.

Ok, so what's the way forward here. Are you guys digging in to the issue? Can I help in some way?

Issue #723, opened 2019-04-30
Consider Logger option to mask based on argument name (not variable name)


dillonkearns:
Hello @benwilson512 @bruce! I'm the author of `elm-graphql` and I was wondering your opinion on something related to logging in Absinthe. In a nutshell, the problem that I'm running into is that Absinthe assumes that you use variables in order to mask sensitive data. But `elm-graphql` doesn't have GraphQL variables. Here's a little background as to why.

## Why there are no variables in `elm-graphql`
`elm-graphql` has a philosophy to use elm constructs rather than GraphQL constructs wherever possible. GraphQL variables are one of the most interesting cases of this. There's no need for variables because elm has constants, functions, modules, variables... so you can do anything you would want to do with a variable in your `elm-graphql` code. Adding GraphQL variables would add significant complexity from the developer ergonomics and the complexity of the type annotations, and for little benefit.

## Masking assumes variables in Absinthe
However, the Absinthe logger relies on variable names to decide which values to mask in the logs. The reasoning given in this issue is that it is more performant to filter based on the JSON that's passed in for variables. This is a totally reasonable decision to arrive at. I'd like to offer some reasons to consider supporting an option to filter based on argument name as well.

### Single source of truth
My thinking is that it would be less error-prone to have the knowledge of what is masked in one place, instead of split between client and server.

### Public-facing API instructions
Also, consider the case where there is a public-facing API. You wouldn't want to put a note in your docs that says "please be sure to pass in `password` args under a variable named `password`, otherwise it will be logged in our server logs."

## Conclusion

So it seems like it would be useful, at least for some standard use cases, to support the ability to mask based on argument names (and parse the GraphQL query if that option is set). It appears that there is currently no way to configure this or use a custom logger.

Here's the issue in the `elm-graphql` repo if you are curious to see more context: https://github.com/dillonkearns/elm-graphql/issues/27

Thanks for taking the time to read. I would love to hear any thoughts you have on this topic. Thank you!

benwilson512:
Hi @dillonkearns. I do think this is a useful feature. However, I also find the stance taken in that elm-graphql channel quite puzzling, and not just for the implications it has for input security.

[...]

PR #1114, opened 2021-10-23
Convert `Absinthe.Language.*` structs to graphql representation

maartenvanvliet:
This PR adds the inspect protocol for the Absinthe.Language.Document struct. 

A parsed graphql document can now be converted back to its graphql string representation. This gives the possible for a few new features:

  * In Elixir 1.13 formatting plugins can be used to format files or sigil delimited texts. This means .graphql documents can be formatted. See `Absinthe.Formatter` for an example.
  * Users could log incoming documents in a custom phase with interpolated arguments masked. (See https://github.com/absinthe-graphql/absinthe/issues/723). 
  * Programmatically create Absinthe.Language.Documents and convert them to graphql documents. 
  * Some instrumentation requires a canonical representation of a document. An incoming Absinthe.Language.Document can now be changed to become canonical (See https://www.apollographql.com/docs/studio/setup-analytics/#operation-signing)

### DIFF
--- a/lib/absinthe/formatter.ex
+++ b/lib/absinthe/formatter.ex
@@ -1,0 +1,34 @@
+defmodule Absinthe.Formatter do
+  @moduledoc """
+  Formatter task for graphql
+
+  Will format files with the extensions .graphql or .gql
+
+  ## Example
+
+      Absinthe.Formatter.format("{ version }")
+      "{\n  version\n}\n"
+
+
+  From Elixir 1.13 onwards the Absinthe.Formatter can be added to
+  the formatter as a plugin:
+
+      # .formatter.exs
+      [
+        # Define the desired plugins
+        plugins: [Absinthe.Formatter],
+        # Remember to update the inputs list to include the new extensions
+        inputs: ["{mix,.formatter}.exs", "{config,lib,test}/**/*.{ex,exs}", "{lib, priv}/**/*.{gql,graphql}"]
+      ]
+
+  """
+
+  def features(_opts) do
+    [sigils: [], extensions: [".graphql", ".gql"]]
+  end
+
+  def format(contents, _opts \\ []) do
+    {:ok, blueprint} = Absinthe.Phase.Parse.run(contents, [])
+    inspect(blueprint.input, pretty: true)
+  end
+end
--- a/lib/absinthe/language/render.ex
+++ b/lib/absinthe/language/render.ex
@@ -1,0 +1,279 @@
+defmodule Absinthe.Language.Render do
+  @moduledoc false
+  import Inspect.Algebra
+  import Absinthe.Utils.Render
+
+  @line_width 120
+
+  def inspect(term, %{pretty: true}) do
+    term
+    |> render()
+    |> concat(line())
+    |> format(@line_width)
+    |> to_string
+  end
+
+  def inspect(term, options) do
+    Inspect.Any.inspect(term, options)
+  end
+
+  defp render(bp)
+
+  defp render(%Absinthe.Language.Document{} = doc) do
+    doc.definitions |> Enum.map(&render/1) |> join("\n")
+  end
+
+  defp render(%Absinthe.Language.OperationDefinition{} = op) do
+    if op.shorthand do
+      concat(operation_definition(op), block(render_list(op.selection_set.selections)))
+    else
+      glue(
+        concat([to_string(op.operation), operation_definition(op)]),
+        block(render_list(op.selection_set.selections))
+      )
+    end
+  end
+
+  defp render(%Absinthe.Language.Field{} = field) do
+    case field.selection_set do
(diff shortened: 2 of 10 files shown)


lib/absinthe/formatter.ex
```diff
@@ -0,0 +1,34 @@
...
+        # Define the desired plugins
+        plugins: [Absinthe.Formatter],
+        # Remember to update the inputs list to include the new extensions
+        inputs: ["{mix,.formatter}.exs", "{config,lib,test}/**/*.{ex,exs}", "{lib, priv}/**/*.{gql,graphql}"]
+      ]
+
+  """
+
+  def features(_opts) do
+    [sigils: [], extensions: [".graphql", ".gql"]]
```
binaryseed, line 27: Is there a use case for a `:G` sigil here?

lib/absinthe/language/render.ex
```diff
@@ -0,0 +1,279 @@
...
+  end
+
+  def inspect(term, options) do
+    Inspect.Any.inspect(term, options)
+  end
+
+  defp render(bp)
+
+  defp render(%Absinthe.Language.Document{} = doc) do
+    doc.definitions |> Enum.map(&render/1) |> join("\n")
```
binaryseed, line 23: I think I used `join(line())` in the other render module..

binaryseed (commented):
This is looking really great!

Would you mind reviewing the test cases in this repo:
* https://github.com/newrelic/absinthe-schema-stitching-example/blob/master/apps/schema_stitch/test/query_generator_test.exs

At New Relic, we built a query renderer but it didn't use the Algebra. We did however, run into a variety of edge cases over time and fixed them each alongside tests. If all those test queries render nicely, I think we'll be good to go on this!

lib/absinthe/formatter.ex
maartenvanvliet: Since Absinthe does not define a sigil of its own I've left it out here. It's fairly simple for users to define their own formatter plugin and include any custom sigils.

lib/absinthe/language/render.ex
maartenvanvliet: Yes, in fact it uses `join([line(), line()])`. I'll change it to be consistent with the other module.

maartenvanvliet:
I've run the documents in https://github.com/newrelic/absinthe-schema-stitching-example/blob/master/apps/schema_stitch/test/query_generator_test.exs and they render nicely :) 

The absinthe parser also accepts SDL. Rendering the Language.* structs for SDL however is not implemented yet. For completeness I'll see if I can do that as a followup.

opsb:
This looks super useful. We've been wanting to sanitise queries for logging so I started down the road of creating a renderer for the blueprint structs. My plan is to allow fields to be whitelisted for logging in the schema e.g.

```
input_object :update_user_input do
  field :id, non_null(:id)
  field :email, :string, meta: [log: true]
  field :first_name, :string
  field :last_name, :string
end
``` 

The reason I'd thought to use the blueprint structs rather than the language structs is that they have the schema_nodes populated so it's possible to query the `meta` and determine if a field should be included in the rendered output or not. It seems excessive to have renderers both for the blueprint and language structs so I'm now pondering what the best path forwards is. Given this renderer exists, what's the best approach for log masking?

benwilson512:
> We've been wanting to sanitise queries for logging so I started down the road of creating a renderer for the blueprint structs.

Can you elaborate on this? Sensitive parameters from end users should always be passed in as variables, and this already supports sanitization.

opsb:
Hey @benwilson512, our absinthe API is client facing so we're not building the queries ourselves. Our position is that the spec supports inline arguments and variables so we should support them too. We do encourage using variables but for instance when exploring our API for the first time it's nicer to be able to just using arguments inline.

benwilson512:
>  Our position is that the spec supports inline arguments and variables so we should support them too

That's fair, but understand that there are reasonable limits to what this sport entails. For example, it is _much_ easier to have syntax errors in the GraphQL than in JSON, and if the GraphQL has syntax errors then it can't easily be stripped of sensitive information. This is particularly complex because Absinthe.Plug logs values _prior_ to parsing, so this is not a simple change.

@opsb I think this question should be followed up in an issue on Absinthe.Plug after this PR is merged, since I think it is largely orthogonal to getting this PR over the finish line.

opsb:
@benwilson512 that's an interesting point about syntax errors - possibly we'd just not log the query at all in that case. I think we'd probably skip the built in logger and introduce our own logging phase further on. But yeah, agree it make sense to explore this more in a follow on ticket.

binaryseed (approved):

binaryseed:
@benwilson512 think this is ready for merge & release?

benwilson512:
❤️ thank you @maartenvanvliet !

PR #1116, opened 2021-11-02
Update typespec for resolver_t

sega-yarkin:
As per code [of `Absinthe.Resolution.call/2`](https://github.com/absinthe-graphql/absinthe/blob/master/lib/absinthe/resolution.ex#L204-L212) three different forms of resolution function are supported, type spec `resolver_t` is updated accordingly to this.


benwilson512 (approved):
Thanks!

Issue #1081, opened 2021-06-07
default_value for an enum type doesn't work when the enum type is defined with the shorthand syntax 


x-ji:
If submitting a bug, please provide the following:

### Environment

* Elixir version (elixir -v): 
```
Erlang/OTP 23 [erts-11.2] [source] [64-bit] [smp:16:16] [ds:16:16:10] [async-threads:1] [hipe]

Elixir 1.11.4 (compiled with Erlang/OTP 23)
```
* Absinthe version (mix deps | grep absinthe): 1.6.3
* Client Framework and version (Relay, Apollo, etc): Not relevant as it's a backend bug

### Expected behavior

Define an enum type with the shorthand syntax:

`enum :example_type, values: [:A, :B, :C]`

In the schema, have a field with default value: `arg :example, :example_type, default_value: :A`.

It should work as expected.

### Actual behavior

A compilation error is thrown:

```
(ArgumentError) you attempted to apply :value on [:A, :B, :C]. If you are using apply/3, make sure the module is an atom. If you are using the dot syntax, such as map.field or module.function(), make sure the left side of the dot is an atom or a map
    :erlang.apply([:A, :B, :C], :value, [])
Stacktrace:
  │ (absinthe 1.6.3) lib/absinthe/phase/schema/validation/default_enum_value_present.ex:33: anonymous fn/1 in Absinthe.Phase.Schema.Validation.DefaultEnumValuePresent.validate_defaults/2
  │ (elixir 1.11.4) lib/enum.ex:1411: Enum."-map/2-lists^map/1-0-"/2
  │ (absinthe 1.6.3) lib/absinthe/phase/schema/validation/default_enum_value_present.ex:33: Absinthe.Phase.Schema.Validation.DefaultEnumValuePresent.validate_defaults/2
  │ (absinthe 1.6.3) lib/absinthe/blueprint/transform.ex:16: anonymous fn/3 in Absinthe.Blueprint.Transform.prewalk/2
  │ (absinthe 1.6.3) lib/absinthe/blueprint/transform.ex:109: Absinthe.Blueprint.Transform.walk/4
  │ (elixir 1.11.4) lib/enum.ex:1533: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
  │ (absinthe 1.6.3) lib/absinthe/blueprint/transform.ex:145: anonymous
```

I had to change the definition of the enum type like this:

```
  enum :example_type do
    value :A
    value :B
    value :C
  end
[...]

PR #1118, opened 2021-11-15
Fix default_value error on enum type

haodt:
Hello, this PR aiming to fix the issue #1081 for the time being before we could come up with anything better.

Currently i'm trying to understand how notation and schema work so i tried my best at fixing the validation which is desirable for the time being since we are having troubles trying to upgrade our absinthe from 1.5.5 to 1.6.6. This issue caused everything to explode ( we indeed used alot of default value).

Any guidances would be appreciated!

### DIFF
--- a/lib/absinthe/phase/schema/validation/default_enum_value_present.ex
+++ b/lib/absinthe/phase/schema/validation/default_enum_value_present.ex
@@ -30,7 +30,15 @@ defmodule Absinthe.Phase.Schema.Validation.DefaultEnumValuePresent do
 
     case Map.fetch(enums, type) do
       {:ok, enum} ->
-        values = Enum.map(enum.values, & &1.value)
+        values =
+          case enum.values do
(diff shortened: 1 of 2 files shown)


haodt:
hey @benwilson512 any thoughts on this one?

katafrakt:
I just wanted to bump this as I stumbled upon the bug it solves today. The fix looks correct to me. I'd be happy to help if there's something more required to get this merged.

Helge801:
Would love to see this merged in. 🙏

sraka1:
Bumping this as well. Tagging @benwilson512 again, this seems like a quick merge

anvox:
We came to another workaround that removing default value from enum params. Default value is checked and filled by resolver. It's a bit more code but unblock upgrading.

HarshBalyan:
@benwilson512 please, review this PR!

PR #1120, opened 2021-11-15
Changes misnamed `:__absinthe_type_import__` to `:__absinthe_type_imports__` in Schema.Notation

### LINKED ISSUE #1119: Module attribute name mismatch for `:__absinthe_type_import__` vs `:__absinthe_type_import__`?
During `Absinthe.Schema.Notation.__using__/1` the module attribute `:__absinthe_type_import__` is defined with `accumulate: true`, but later in the Notation module a very similar, but different module attribute `:__absinthe_type_imports__` is used 3 times.

The module attribute registration:

https://github.com/absinthe-graphql/absinthe/blob/d7837a7dc610c014bedae36409a36f515936f541/lib/absinthe/schema/notation.ex#L43

The first and second uses:

https://github.com/absinthe-graphql/absinthe/blob/d7837a7dc610c014bedae36409a36f515936f541/lib/absinthe/schema/notation.ex#L1678-L1684

The third use:

https://github.com/absinthe-graphql/absinthe/blob/d7837a7dc610c014bedae36409a36f515936f541/lib/absinthe/schema/notation.ex#L1757-L1763

Looks like potential the source of a bug.

elbow-jason:
This PR changes a (probably) misnamed module attribute registered during the compilation of `use Schema.Notation`.

Closes #1119

Signed-off-by: Jason Goldberger <jasongoldberger@gmail.com>

### DIFF
--- a/lib/absinthe/schema/notation.ex
+++ b/lib/absinthe/schema/notation.ex
@@ -40,7 +40,6 @@ defmodule Absinthe.Schema.Notation do
           batch: 4
         ]
 
-      Module.register_attribute(__MODULE__, :__absinthe_type_imports__, accumulate: true)
       @desc nil
       import unquote(__MODULE__), unquote(import_opts)
       @before_compile unquote(__MODULE__)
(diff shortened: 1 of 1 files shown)


benwilson512:
@elbow-jason hey thanks! I'll have to look at this more closely but it does look like tests are failing in the CI, do you see them passing locally?

elbow-jason:
@benwilson512 They are most definitely not passing locally. This PR needs a little more love before it's ready.

binaryseed:
I think the existing behavior was intended, and the fact that the accumulating attribute was mis-spelled was actually what made it work correctly, since the code seems to manually manage the "default" empty list case...

PR #1122, opened 2021-11-20
Remove unused transform walk functions

maartenvanvliet:
This was introduced in commit 7529731 but with the removal of
the flatten.ex in 0f33807 it's no longer needed.


benwilson512:
Thanks!

PR #1125, opened 2021-11-25
Invert the instructions for installing subscriptions with phoenix.

JonRowe:
Add the instructions for installing subscriptions into current versions of Phoenix as the default, and then explain how to use in older versions rather than the other way around. This is less confusing to newcomers, and people working with older Elixir are more likely to know they need to look for special instructions.


benwilson512:
Thanks!

Issue #1080, opened 2021-06-04
Absinthe doesn't conform to GraphQL spec for validating schemas with interfaces that implement other interfaces


stellanor:
If submitting a bug, please provide the following:

### Environment

* Elixir version (elixir -v): Elixir 1.11.4 (compiled with Erlang/OTP 23)
* Absinthe version (mix deps | grep absinthe): absinthe 1.6.4
* Client Framework and version (Relay, Apollo, etc): n/a

### Expected behavior

When defining a schema with a higher-order interface, for example, interface `Mammal` implements interface `Animal`, and object type `Dog` implements `Mammal` (and therefore `Dog` transitively implements `Animal`)...

I expect compilation of object types in that schema to enforce the [interface transitivity requirement of the GraphQL spec](https://github.com/graphql/graphql-spec/blob/main/spec/Section%203%20--%20Type%20System.md#interfaces):

> Transitively implemented interfaces (interfaces implemented by the interface that is being implemented) must also be defined on an implementing type or interface. For example, Image cannot implement Resource without also implementing Node

### Actual behavior

However, when defining an absinthe schema such as this `Dog`/`Mammal`/`Animal` example, if I only explicitly implement `Mammal` on `Dog`, I do not get a compilation error about the missing `Animal` interface declaration, and instead, I get runtime errors when resolving queries that spread on the various types. Here is an example.

My test schema:

```
defmodule AbsintheTestSchema do
  use Absinthe.Schema

  interface :animal do
    field :name, :string
    field :leg_count, :integer
    resolve_type fn
      %{breed: _}, _ -> :dog
      _, _ -> nil
    end
  end

  interface :mammal do
    interface :animal
    field :name, :string
    field :leg_count, :integer
    field :has_fur, :boolean

    resolve_type fn
      %{breed: _}, _ -> :dog
      _, _ -> nil
    end
  end

  object :dog do
    interface :mammal
    field :breed, :string
    field :name, :string
    field :leg_count, :integer
    field :has_fur, :boolean
  end

  query do
    field :my_animal, :animal do
      resolve fn _, _ ->
         {:ok, %{name: "Ponyo", leg_count: 4, has_fur: true, breed: "pitbull" }}
      end
    end
  end
end
```

Then I hop into iex with this test project, and define this query string:

```
{
  myAnimal {
    name
    ... on Mammal {
      hasFur
    }
    ... on Dog {
      breed
    }
  }
}
```

[...]

PR #1127, opened 2021-11-25
Validate object/interfaces implement all transitive interfaces

maartenvanvliet:
See https://github.com/absinthe-graphql/absinthe/issues/1080 for a full description

 * adds a type reference check to see whether interfaces on interfaces exist
 * adds a check to ensure that an object/interface implements all interfaces, also interfaces it transitively has

@tlvenn can you take a look

fixes https://github.com/absinthe-graphql/absinthe/issues/1080

### DIFF
--- a/test/absinthe/schema/rule/object_must_implement_interfaces_test.exs
+++ b/test/absinthe/schema/rule/object_must_implement_interfaces_test.exs
@@ -25,6 +25,9 @@ defmodule Absinthe.Schema.Rule.ObjectMustImplementInterfacesTest do
       field :another_parent, :named
 
       resolve_type fn
+        %{type: :dog}, _ -> :dog
+        %{type: :user}, _ -> :user
+        %{type: :cat}, _ -> :cat
         _, _ -> nil
       end
     end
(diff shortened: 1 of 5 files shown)


test/absinthe/schema/rule/object_must_implement_interfaces_test.exs
```diff
@@ -22,6 +23,10 @@ defmodule Absinthe.Schema.Rule.ObjectMustImplementInterfacesTest do
     interface :parented do
       field :parent, :named
       field :another_parent, :named
+
+      resolve_type fn
+        _, _ -> nil
```
tlvenn, line 28: Is there any reason why the resolve_type does not include `dog`, `cat` or `user` as in the other interfaces ?
Like:
```
resolve_type fn
  %{type: :dog}, _ -> :dog
  %{type: :user}, _ -> :user
  %{type: :cat}, _ -> :cat
  _, _ -> nil
end
```

tlvenn (commented):
Looks good to me, thanks for this @maartenvanvliet !

test/absinthe/schema/rule/object_must_implement_interfaces_test.exs
maartenvanvliet: Forgot about that one 🙂

tlvenn: Hehe no worry, as soon as you get the chance to fix this I will merge this PR, thanks again.

maartenvanvliet: 👍

PR #1128, opened 2021-11-28
Add `telemetry_span_context` following conventions

andrewhr:
Absinthe's telemetry events includes `id` metadata, as means to correlate start/stop/exception events. Telemetry defines a `telemetry_span_context` for this same reasons.

This patch publishes the same `id` value  as both keys, so we maintain backwards compatibility as well keep adherent to conventions, for tools that rely on that - like OpenTelemetry.

The `telemetry_span_context` is generated using `:erlang.make_ref/0`. Given typespecs describe as `term`, should be safe to use the same value as `id`.


benwilson512:
Thanks!

PR #1131, opened 2021-12-06
Add spec compliant Int type

### LINKED ISSUE #921: Max integer size in Absinthe is much larger than GraphQL's

### Environment

* Elixir version (elixir -v): 1.8.2
* Absinthe version (mix deps | grep absinthe): 1.4.16
* Client Framework and version (Relay, Apollo, etc): N/A -- using a custom-built client

### Expected behavior

When entering an integer bigger than the largest supported number by a 32-bit integer, Absinthe should return that it's an invalid integer  based on [GraphQL's typing](https://graphql.org/learn/schema/#scalar-types) and [GraphQL spec](https://spec.graphql.org/June2018/#sec-Int)

### Actual behavior

Absinthe validates numbers larger than 32-bit integers as integers which causes issues in other parts of the application that expect a 32-bit integer.

No relevant middleware code but I checked out the reasoning behind this in: https://github.com/absinthe-graphql/absinthe/blob/9f7350a6d4d7e31d553d436348d373e2b896db10/lib/absinthe/type/built_ins/scalars.ex

Where it mentions an adherence to the [IEEE754 standard](https://en.wikipedia.org/wiki/IEEE_754) standard for *double precision*. However, that standard directly clashes with the GraphQL spec.

IEEE 754 max integer is: 9,007,199,254,740,991
GraphQL max integer is: 2,147,483,647 

Anything over 2,147,483,647 should be a different type in a GraphQL schema (even if that doesn't map 1-to-1 in Elixir). The [GraphQL spec on Integers](https://spec.graphql.org/June2018/#sec-Int) suggests using a String or custom scalar instead.

tofran:
## Summary

This pull request closes #921.

Currently, as of `v1.6.6` and all previous versions, the Int scalar type provided by absinthe **is not compliant** with the [GraphQl spec](https://spec.graphql.org/October2021/#sec-Int). Thus [absinthe documentation](https://hexdocs.pm/absinthe/1.6.6/Absinthe.Type.Scalar.html)  is untrue, as seen in the [implementation](https://github.com/absinthe-graphql/absinthe/blob/master/lib/absinthe/type/built_ins/scalars.ex#L8-L11).

Since this bug is being used as a feature in the wild. We cannot just change the implementation, at least without a major version change. Thus this PR introduces a new, optional pipeline modifier `Absinthe.Phase.Schema.SpecCompliantInt`, that can be used with the helper `use_spec_compliant_int_scalar`:

```elixir
defmodule MySchema do
  use Absinthe.Schema,
    use_spec_compliant_int_scalar: true

    # ...
end
```

Users relying on the legacy implementation (Integers between `-(2^53 - 1)` and `2^53 - 1` specified by [IEEE 754](http://en.wikipedia.org/wiki/IEEE_floating_point)) should also be encouraged to migrate to the fixed type, and [create custom types](https://hexdocs.pm/absinthe/custom-scalars.html) instead for their needs.

## Todo
- [x] Add test for the new Int type;
       ~Should I make `Absinthe.Type.BuiltIns.ScalarsTest` synchronous (`async: alse`) in order to mutate the `:use_legacy_non_compliant_int_scalar_type` config before and after the test? Or do you have a better alternative? The problem here is that the code follows two paths at compile time.~ EDIT: I just proposed a solution, check if it makes sense.
- [x] Edit changelg;
- [x] Ensure all documentation is up to date. Including mentioning the new `:use_spec_compliant_int_scalar` option.

## Requirements

- [x] The existing non compliant `Int` type must be kept by default (at least for now - it can only be changed in a major, breaking release);
- [x] Users should be able to change to the new, correct, Int just by changing some configuration on their side;
- [x] Tests for both versions need to exist;
- [x] All remaining features should be kept working as it.


lib/absinthe/type/built_ins/scalars.ex
```diff
@@ -3,23 +3,43 @@ defmodule Absinthe.Type.BuiltIns.Scalars do
...
-  def serialize_integer(n) when is_integer(n), do: n
+  def serialize_int(value) when is_integer(value) and value >= @min_int and value <= @max_int do
+    value
+  end
 
-  def serialize_integer(n) do
+  def serialize_int(value) do
     raise Absinthe.SerializationError, """
-    Value #{inspect(n)} is not a valid integer
+    Value #{inspect(value)} is not a valid integer.
```
kdawgwilk, line 42: This would point to the actual scalar type which may be helpful
```suggestion
    Value #{inspect(value)} is not a valid Int.
```

lib/absinthe/type/built_ins/scalars.ex
```diff
@@ -3,23 +3,43 @@ defmodule Absinthe.Type.BuiltIns.Scalars do
 
   @moduledoc false
 
-  scalar :integer, name: "Int" do
-    description """
-    The `Int` scalar type represents non-fractional signed whole numeric values.
-    Int can represent values between `-(2^53 - 1)` and `2^53 - 1` since it is
-    represented in JSON as double-precision floating point numbers specified
-    by [IEEE 754](http://en.wikipedia.org/wiki/IEEE_floating_point).
+  if Application.get_env(:absinthe, :use_legacy_non_compliant_integers, true) do
```
kdawgwilk, line 6: not sure where in the documentation this should live or possibly the changelog so that way others are aware of this setting

kdawgwilk:
I would be nice to optionally plug into the serialize or parse to be able to report metrics or log each time an invalid Int was used so you can get an idea of the impact flipping the switch on might have

lib/absinthe/type/built_ins/scalars.ex
tofran: I have added it to the installation instructions and also information in the changelog. Do you think it is enough?

lib/absinthe/type/built_ins/scalars.ex
```diff
@@ -3,23 +3,43 @@ defmodule Absinthe.Type.BuiltIns.Scalars do
 
   @moduledoc false
 
-  scalar :integer, name: "Int" do
-    description """
-    The `Int` scalar type represents non-fractional signed whole numeric values.
-    Int can represent values between `-(2^53 - 1)` and `2^53 - 1` since it is
-    represented in JSON as double-precision floating point numbers specified
-    by [IEEE 754](http://en.wikipedia.org/wiki/IEEE_floating_point).
+  if Application.get_env(:absinthe, :use_legacy_non_compliant_int_scalar_type, true) do
```
benwilson512, line 6: In my view the flag is backwards here. We can't make a change like this in a point release that might break people's working code. The default needs to be the non compliant int, and then you can opt into the compliant int. In a 2.0 release we can swap this.

tofran: @benwilson512 

> The default needs to be the non compliant int

The default is the non compliant Int.

---

Maybe the naming is not good? The idea is that the lib will **use the legacy non compliant int scalar type** by default, thus `true` is the default.

You can also see this explained in the documentation I added.

In the `Installation`:

> As explained in Absinthe.Type.Scalar, for legacy reasons, absinthe's default `Int` scalar type is not compliant with the GraphQl spec, thus it is recommended to configure the following for new installations:
> 
> ```elixir
> config :absinthe, use_legacy_non_compliant_int_scalar_type: :false
> ```

<details>
  <summary>How it looks in the doc</summary>
  image
</details>


In the `Absinthe.Type.Scalar`:

> `:integer` - Represents non-fractional signed whole numeric value. **By default it is not** compliant with the GraphQl Specification**, it can represent values between `-(2^53 - 1)` and `2^53 - 1` as specified by [IEEE 754](http://en.wikipedia.org/wiki/IEEE_floating_point). To use the [GraphQl compliant Int](https://spec.graphql.org/October2021/#sec-Int), representing values between `-2^31` and `2^31 - 1`, set `config :absinthe, use_legacy_non_compliant_int_scalar_type: :false` in your config. Note that Absinthe uses the full word `:integer` to identify this type, but its `name` (used by variables, for instance), is `"Int"`.

<details>
  <summary>How it looks in the doc</summary>
  Screen Shot 2021-12-07 at 13 27 58
</details>

`In the CHANGELOG`:

> Bug Fix: Add **optional** fix for non compliant built-in scalar Int type. Set `config :absinthe, use_legacy_non_compliant_int_scalar_type: :false` in your config to use the fixed Int type. It is also advisable to upgrade for custom types if you are leveraging the use of integers outside the standard. More information in the [pull request #1131](https://github.com/absinthe-graphql/absinthe/pull/1131).

tofran:
@benwilson512 You can also see that all the tests above still pass. This proves that the legacy implementation is the default.

lib/absinthe/type/built_ins/scalars.ex
kdawgwilk: I think he is going for something more along the lines of
```elixir
config :absinthe, use_spec_compliant_int_scalar: true
```
to enable

kdawgwilk: Which would then make this
```elixir
if Application.get_env(:absinthe, :use_spec_compliant_int_scalar, false) do
```

maartenvanvliet: This setting is now applied globally, across different schema's in the same application. 

Would it not be possible to apply it per schema by manipulating the scalar in schema phases? This way each schema can individually opt in to these changes.

tofran: @kdawgwilk Thank you, I see, reversing the name will do the trick then.

tofran: > Would it not be possible to apply it per schema by manipulating the scalar in schema phases?

@maartenvanvliet making it configurable across schemas would be cool. I'm not sure how would we do this nor what are the implications.
But I also think this is a nice to have that can be left for another pull request, if needed. 
Making this feature would most likely force us to fiddle with more components within the project, and we would be building something to mutate the standard built in types - that should never be done.

Users can always create their custom types if they want the legacy one. The only difference for consumers is the name of the scalar during introspection. 

What are your thoughts?

tofran:
@kdawgwilk

> I would be nice to optionally plug into the serialize or parse to be able to report metrics or log each time an invalid Int was used so you can get an idea of the impact flipping the switch on might have

Before this change is available, how would you know if there are non GraphQL compliant Ints bouncing around your applications?
I would expect that this investigation needs to be done before the migration. Thus you need to know metrics about current values that are actually valid but will not be once the change is applied.

Right now you could try to achieve this, for example with a middleware or a custom field.
If it makes sense to hook into the parsed/serialized values, it needs to be globally for the built-in types, including for reporting valid values that do not respect some pre-defined rule. And thus, in my opinion this is another feature.

lib/absinthe/type/built_ins/scalars.ex
maartenvanvliet: There's a `absinthe/test/absinthe/schema/hydrate_builtins_test.exs` test already that demonstrates how to override a builtin scalar. This relies on the hydrate functionality. 

A similar approach could be used to override the default int scalar. In line with the rest of Absinthe it would an extensible way to override builtin scalars. Arguably the building blocks are there already, and we'd need to find a way to expose it to support the legacy->compliant integer conversion.

tofran: @maartenvanvliet interesting, sorry I didn't know about the hydrate feature.

Right now you can override the Int with hydrate regardless of the configuration, it just works like it worked before, no existing functionality was changed.
The only things to note is that `serialize_int` (in `Absinthe.Type.BuiltIns.Scalars`) changes depending on the setting. Would having distinct `serialize_ieee_int` and `serialize_graphql_int` help if users want to use them separately?

Software cannot fulfill everyone's dreams and corner cases. Is the feature requested above something nice to have or it is it actually a blocker for this fix? This PR is an optional, opt-in fix, that does not affect any user that does not explicitly says they want it, keeping a all the existing functionality.
If this feature is going forwards, in this pull request, I will probably need more help.

tofran:
Hello again @kdawgwilk @benwilson512
Have you had time to take a look at the responses I left above?

I am new to this codebase, and just really want to provide an out of the box way to make this type compliment with the spec. What, in your opinion are the minimum requirements for this to be feasible?

Does your opinion still stand? You you understand my point of view? Do you consider that I being to resistant to the features proposed? Feel free to leave your feedback.

Also, could you please approve the workflow?

kdawgwilk (approved):
Sorry my review doesnt count but I like this change

lib/absinthe/type/built_ins/scalars.ex
maartenvanvliet: I'd be in favor of making it schema specific. Absinthe hardly relies on global configuration (except for some logging I think) and this gives the most control to the user.

Here is a rough version https://gist.github.com/maartenvanvliet/cc3005fdb5950059edc5e0e753bd7c0c where the default integer type is overriden with the spec compliant integer. Needs some cleaning up though.

This can be done per schema by setting the `@pipeline_modifier Absinthe.Phase.Schema.SpecCompliantInt`, so it's not necessary to rely on the global configuration and it remains opt in for now. 

It could be made more ux friendly by e.g. `use Absinthe.Schema, use_spec_compliant_int_scalar: true`, which sets the pipeline modifier for the user.

tofran: Thank you for the detailed suggestion and the sample code.
I will take a look as submit the necessary changes once I am available.

kdawgwilk (approved):

tofran:
@kdawgwilk I completely forgot about this.
Sorry for leaving this abandoned. I will experiment with it, and the above suggestions ASAP.

tofran:
Again, sorry for the delay. But here it is.

It is much better now, to use the spec compliant Int you just need to pass `use_spec_compliant_int_scalar` to Absinthe.Schema:

```elixir
use Absinthe.Schema, use_spec_compliant_int_scalar: true
```

You could also opt to use `@pipeline_modifier Absinthe.Phase.Schema.SpecCompliantInt`

@maartenvanvliet Thank you very much for your guidance, it wouldn't have been possible otherwise.

If needed, I can clean the history or we could squash all these commits on merge.

test/absinthe/type/built_ins/scalars_test.exs
```diff
@@ -25,27 +36,43 @@ defmodule Absinthe.Type.BuiltIns.ScalarsTest do
   end
 
   describe ":integer" do
-    test "serializes as an integer" do
+    test "can serilize a valid integer using default non standard Int (IEEE 754)" do
```
maartenvanvliet, line 39:
```suggestion
    test "can serialize a valid integer using default non standard Int (IEEE 754)" do
```

test/absinthe/type/built_ins/scalars_test.exs
```diff
@@ -25,27 +36,43 @@ defmodule Absinthe.Type.BuiltIns.ScalarsTest do
...
+
+      assert -1 == serialize(:integer, -1)
+      assert 0 == serialize(:integer, 0)
       assert 1 == serialize(:integer, 1)
+      assert @max_ieee_int == serialize(:integer, @max_ieee_int)
+      assert @min_ieee_int == serialize(:integer, @min_ieee_int)
     end
 
-    test "can be parsed from an integer within the valid range" do
+    test "cannot serilize an integer outside boundaries using default non standard Int (IEEE 754)" do
```
maartenvanvliet, line 49:
```suggestion
    test "cannot serialize an integer outside boundaries using default non standard Int (IEEE 754)" do
```

lib/absinthe/type/built_ins/spec_compliant_int.ex
```diff
@@ -0,0 +1,49 @@
...
+
+  def serialize_int(value) when is_integer(value) and value >= @min_int and value <= @max_int do
+    value
+  end
+
+  defp parse_int(value) when is_integer(value) and value >= @min_int and value <= @max_int do
+    {:ok, value}
+  end
+
+  defp parse_int(_) do
```
maartenvanvliet, line 28: For followup, when https://github.com/absinthe-graphql/absinthe/pull/1187 is merged a custom error could be returned when the int is out of range.

test/absinthe/type/built_ins/scalars_test.exs
```diff
@@ -11,8 +11,19 @@ defmodule Absinthe.Type.BuiltIns.ScalarsTest do
     end
   end
 
-  @max_int 9_007_199_254_740_991
-  @min_int -9_007_199_254_740_991
+  @max_ieee_int 9_007_199_254_740_991
+  @min_ieee_int -9_007_199_254_740_991
+
+  defp recompile(module) do
```
maartenvanvliet, line 17: Is the recompile/1 still needed?

lib/absinthe/phase/schema/spec_complient_int.ex
```diff
@@ -0,0 +1,54 @@
+defmodule Absinthe.Phase.Schema.SpecCompliantInt do
```
maartenvanvliet, line 1: Filename is `spec_complient_int.ex` should be `spec_compliant_int.ex`

maartenvanvliet (commented):
Looks good, couple of nitpicks

lib/absinthe/type/built_ins/spec_compliant_int.ex
tofran: Should this PR wait for this?
Would it just say _the input is out of range (expected an signed integer from XXX to XXX)_?

tofran:
> Looks good, couple of nitpicks

Thank you for the review, sorry I missed some typos. Should be fixed now.

lib/absinthe/type/built_ins/spec_compliant_int.ex
benwilson512: That PR is merged now, let's rebase on master and add a custom error.

lib/absinthe/type/built_ins/spec_compliant_int.ex
```diff
@@ -0,0 +1,49 @@
...
+    The `Int` scalar type represents non-fractional signed whole numeric
+    values between `-2^31` and `2^31 - 1`, as outlined in the
+    [GraphQl spec](ttps://spec.graphql.org/October2021/#sec-Int).
+    """
+
+    serialize &__MODULE__.serialize_int/1
+    parse parse_with([Absinthe.Blueprint.Input.Integer], &parse_int/1)
+  end
+
+  @min_int -2 ** 31
```
benwilson512, line 17:
```suggestion
  @min_int trunc(:math.pow(-2, 31))
```

lib/absinthe/type/built_ins/spec_compliant_int.ex
```diff
@@ -0,0 +1,49 @@
...
+  @max_int 2 ** 31 - 1
```
benwilson512, line 18:
```suggestion
  @max_int trunc(:math.pow(2, 31)) -1
```

benwilson512 (commented):
We need to support earlier Elixir versions than 1.13, so we can't use `**` yet.

benwilson512:
Great work everyone, thanks!

Issue #1133, opened 2021-12-14
Duplicate fields on type are currently allowed


kdawgwilk:
### Environment

* Elixir version (elixir -v): 1.12.3
* Absinthe version (mix deps | grep absinthe): 1.6.6
* Client Framework and version (Relay, Apollo, etc): N/A

### Expected behavior

Absinthe fails to compile my schema and gives an error message indicating that the same field is being defined more than once on a type

### Actual behavior

schema compiles without issue

### Relevant Schema/Middleware Code

```elixir
defmodule MySchema do
  use Absinthe.Schema

  query do
    field :foo, :string
    field :foo, :string
  end
end
```

This results in a malformed SDL file generated from the `mix absinthe.schema.sdl` task which is rejected by most tooling e.g. schema registries.

kdawgwilk:
This can commonly occur when sharing schema types between multiple graphs when using `import_types` and `import_fields`

Issue #1049, opened 2021-03-11
Confusing warning when fields duplicated in a schema


joeapearson:
This isn't really a bug per-se but a usability issue, and was entirely caused by my own mistake:

I suddenly found after a series of changes to my code that I was getting warnings like this:

```
warning: this clause cannot match because a previous clause at line 1 always matches
  lib/myapp_web/schema/my_types.ex:1
```

It turns out I had mistakenly added duplicate fields to one of my schema.  The only way I could track it down was by progressively commenting out bits of my schema until I isolated the warning.

### Environment

* Elixir version (elixir -v):
        Erlang/OTP 23 [erts-11.1.8] [source] [64-bit] [smp:12:12] [ds:12:12:10] [async-threads:1] [hipe]

        Elixir 1.11.3 (compiled with Erlang/OTP 23)

* Absinthe version (mix deps | grep absinthe):
        * absinthe 1.6.0 (Hex package) (mix)
          locked at 1.6.0 (absinthe) 99915841
        * absinthe_plug 1.5.4 (Hex package) (mix)
          locked at 1.5.4 (absinthe_plug) 80360cd8
* Client Framework and version (Relay, Apollo, etc): N/A

### Expected behavior

A warning or error of some sort that informed me that I had made a silly mistake!

### Actual behavior

I receive several warnings like this:

```
warning: this clause cannot match because a previous clause at line 1 always matches
  lib/myapp_web/schema/my_types.ex:1
```

The line number is always `1`.

### Relevant Schema/Middleware Code

Here's a minimal example that reproduces the behaviour:

```
object :my_object do
    field :my_field, :string
    field :my_field, :number
  end
```

Of course this is a contrived example because you can clearly see the mistake when there are but two properties.  If your `object` schema is larger (like mine is) then it's harder to see.  It'd be great if some sort of warning and or error was triggered (with the appropriate line number, instead of just line `1`).

PR #1135, opened 2021-12-28
Add unique field names validation

maartenvanvliet:
Fixes https://github.com/absinthe-graphql/absinthe/issues/1133
Also fixes https://github.com/absinthe-graphql/absinthe/issues/1049
as the schema's won't compile anymore, so the warning no longer happens

### DIFF
--- a/lib/absinthe/phase/schema/validation/unique_field_names.ex
+++ b/lib/absinthe/phase/schema/validation/unique_field_names.ex
@@ -1,0 +1,66 @@
+defmodule Absinthe.Phase.Schema.Validation.UniqueFieldNames do
+  @moduledoc false
+
+  @behaviour Absinthe.Phase
+  alias Absinthe.Blueprint
+
+  def run(bp, _) do
+    bp =
+      bp
+      |> Blueprint.prewalk(&handle_schemas(&1, :name))
+
+    {:ok, bp}
+  end
+
+  defp handle_schemas(%Blueprint.Schema.SchemaDefinition{} = schema, key) do
+    schema = Blueprint.prewalk(schema, &validate_types(&1, key))
+    {:halt, schema}
+  end
+
+  defp handle_schemas(obj, _) do
+    obj
+  end
+
+  defp validate_types(%type{} = object, key)
+       when type in [
+              Blueprint.Schema.InputObjectTypeDefinition,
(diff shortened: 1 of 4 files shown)


kdawgwilk (approved):
Thanks!

lib/absinthe/phase/schema/validation/unique_field_names.ex
```diff
@@ -0,0 +1,66 @@
...
+
+  defp validate_types(%type{} = object, key)
+       when type in [
+              Blueprint.Schema.InputObjectTypeDefinition,
+              Blueprint.Schema.InterfaceTypeDefinition,
+              Blueprint.Schema.ObjectTypeDefinition
+            ] do
+    fields =
+      for field <- object.fields do
+        if duplicate?(object.fields, field, key) do
```
benwilson512, line 32: I think the first arg here should be a Map of the `object.field` names, otherwise this becomes `O(n^2)`. We can do:

```
name_counts = Enum. frequencies_by(object.fields, &Map.get(&1, key))

def duplicate?(name_counts, field, key) do
  field_identifier = Map.get(field, key)
  Map.get(name_counts, field_identifier, 0) > 1
end
```

benwilson512 (changes requested):
Hey thanks for submitting this! Small performance tweak and then we're good to go!

lib/absinthe/phase/schema/validation/unique_field_names.ex
maartenvanvliet: Done :)

I copied the duplicate approach from other parts of the code. E.g. `Absinthe.Phase.Document.Validation.UniqueVariableNames`. These could probably also be optimized, esp since they are run in the document pipeline and not in the schema pipeline, so the cost is incurred for every request.

benwilson512: Ah that sounds like a great followup!

maartenvanvliet: For now this PR can be merged

benwilson512:
Oh, one last thing, can you add a CHANGELOG.md entry. I'm gonna start requiring that cause it's super hard to track down changes people have made in the future.

benwilson512:
Thank you!

PR #1140, opened 2022-01-09
Apply type system directives in macro based schemas

maartenvanvliet:
This PR allows schemas to apply type system directives in macro based schemas. This was already possible for SDL based schemas. 
It makes deprecations work through the `deprecated` type system directive. This directive is defined on the default prototype schema and this led to a problem with the several deprecated introspection fields. These were extracted to a separate schema phase so the prototype schema can skip them. In Absinthe 2.0 this phase can be removed altogether but I've kept it for now for backwards compatibility.

 * Some work for a followup is to validate the arguments passed into applied type system directives. This is currently also not done for SDL based schema afaict.

 * Schema level directives also are not possible yet e.g.
 ```
 schema @feature(name: ":schema") {
   query: Query
 }
```

Since macro based deprecations are no longer special cased it will also fix https://github.com/absinthe-graphql/absinthe/pull/1045

### DIFF
--- a/lib/absinthe/middleware.ex
+++ b/lib/absinthe/middleware.ex
@@ -310,6 +310,9 @@ defmodule Absinthe.Middleware do
       [{:ref, Absinthe.Type.BuiltIns.Introspection, _}] ->
         expanded
 
+      [{:ref, Absinthe.Phase.Schema.DeprecatedDirectiveFields, _}] ->
+        expanded
+
       _ ->
         schema.middleware(expanded, field, object)
     end
--- a/lib/absinthe/schema/notation.ex
+++ b/lib/absinthe/schema/notation.ex
@@ -195,6 +195,13 @@ defmodule Absinthe.Schema.Notation do
   end
 
   defmacro object(identifier, attrs, do: block) do
+    block =
+      for {identifier, args} <- build_directives(attrs) do
+        quote do
+          directive(unquote(identifier), unquote(args))
+        end
+      end ++ block
+
     {attrs, block} =
       case Keyword.pop(attrs, :meta) do
         {nil, attrs} ->
@@ -381,6 +388,13 @@ defmodule Absinthe.Schema.Notation do
       end
 
     block =
+      for {identifier, args} <- build_directives(attrs) do
+        quote do
+          directive(unquote(identifier), unquote(args))
+        end
+      end ++ block
+
+    block =
       case Keyword.get(attrs, :meta) do
         nil ->
           block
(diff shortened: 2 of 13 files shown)

### CODE REVIEW

File: lib/absinthe/middleware.ex [hunk 1]
```diff
@@ -310,6 +310,9 @@ defmodule Absinthe.Middleware do
       [{:ref, Absinthe.Type.BuiltIns.Introspection, _}] ->
         expanded
 
+      [{:ref, Absinthe.Phase.Schema.DeprecatedDirectiveFields, _}] ->
```
maartenvanvliet, line 313: Not really happy with this, but can be removed when these fields are no longer necessary.

lib/absinthe/schema/notation.ex
```diff
@@ -857,20 +856,34 @@ defmodule Absinthe.Schema.Notation do
...
+                  :object,
+                  :scalar,
+                  :union,
+                  :value
+                ]
+              ]}
+
   @doc """
-  Defines a directive
+  Defines a or applies directive
```
ymtszw, line 875: Could be:
```suggestion
  Defines or applies a directive
```

lib/absinthe/schema/notation.ex
```diff
@@ -879,16 +892,65 @@ defmodule Absinthe.Schema.Notation do
...
+
+  scalar :sweet_scalar do
+    directive :feature, name: ":scalar"
+    parse &Function.identity/1
+    serialize &Function.identity/1
   end
   ```
   """
-  defmacro directive(identifier, attrs \\ [], do: block) do
+  defmacro directive(identifier, attrs, do: block) when is_list(attrs) when not is_nil(block) do
```
maartenvanvliet, line 930: `directive/1,2,3` can be used for both defining directives as applying them to types in the schema. 

I think this is clear for the users but perhaps there's another function name we can use for applying the directive.

ymtszw (commented):
Amazing! As commented in #1045 I was also looking into the issue and saw it a huge undertaking to properly support type system directives in macro schema by my hand, even for only supporting `deprecated`.

As a workaround I was to adopt [@zoldar's gist](https://gist.github.com/zoldar/d617f2037854ccfc26cd428e11df6ed7) on userland (which works for now, at least for `deprecated`) but this patch should nicely remove the extra work. Love it!

ymtszw:
A step to move https://github.com/absinthe-graphql/absinthe/issues/1003 forward (by a lot!)

test/absinthe/schema/type_system_directive_test.exs
```diff
@@ -91,8 +91,137 @@ defmodule Absinthe.Schema.TypeSystemDirectiveTest do
...
-    assert Absinthe.Schema.to_sdl(TypeSystemDirectivesSchema) ==
-             TypeSystemDirectivesSchema.sdl()
+  defmodule TypeSystemDirectivesMacroSchema do
+    use Absinthe.Schema
+
+    @prototype_schema WithTypeSystemDirective
+
+    query do
+      field :post, :post do
+        directive :feature, name: ":field_definition"
```
kdawgwilk, line 101: I am a bit confused about these strings that look like atoms?  are they arbitrary strings and you just happened to use strings that look like atoms or is there some significance to this format?

maartenvanvliet: It's arbitrary, in the same test file there's an SDL schema that follows the same format (https://github.com/absinthe-graphql/absinthe/blob/2390500573c08d882143b48bc02f455ae5d0fd86/test/absinthe/schema/type_system_directive_test.exs). The macro-schema is a near copy of that.

kdawgwilk:
❤️  Really looking forward to this since it will greatly simplify the https://github.com/DivvyPayHQ/absinthe_federation library

Issue #1134, opened 2021-12-17
Nullability of operation variables missing validation


kdawgwilk:
If submitting a bug, please provide the following:

### Environment

* Elixir version (elixir -v): 1.12
* Absinthe version (mix deps | grep absinthe): 1.6.6
* Client Framework and version (Relay, Apollo, etc): N/A

### Expected behavior

A validation error that the operation argument nullability does not match the nullability of the argument type where its used.

### Actual behavior

Absinthe accepts the operation variable type without issue

### Relevant Schema/Middleware Code

```elixir
defmodule Schema do
  use Absinthe.Schema

  query do
    field :foo, :string do
      arg :bar, non_null(:integer)
    end
  end
end
```

```gql
query VariableValidation(bar: Int) {
  foo(bar: $bar)
}
```

the `bar` argument is defined as `non_null` but the operation defines the variable as `Int` which can be null.

PR #1141, opened 2022-01-09
Improve type checks on variables

maartenvanvliet:
Currently the check was whether the unwrapped type of the variable and e.g. the argument were the same. The spec is stricter than this. In particular this section: https://spec.graphql.org/October2021/#sec-All-Variable-Usages-are-Allowed

This PR adds checks to see if the wrapped types are the same or the variable is stricter than the argument. Also, when default values are given some of these checks can be relaxed.

This is a bugfix in respect to the spec but will have the effect that some previously valid documents are now marked as invalid.

This fixes https://github.com/absinthe-graphql/absinthe/issues/1134

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Bug Fix: [Validate field names are unique to an object/interface/input object](https://github.com/absinthe-graphql/absinthe/pull/1135)
+- Bug Fix: [Validate variable usage in according to spec](https://github.com/absinthe-graphql/absinthe/pull/1141)
 
 ## 1.6.6
 
--- a/lib/absinthe/blueprint/type_reference.ex
+++ b/lib/absinthe/blueprint/type_reference.ex
@@ -29,6 +29,22 @@ defmodule Absinthe.Blueprint.TypeReference do
     unwrap(inner)
   end
 
+  @doc """
+  Get the GraphQL name for a (possibly wrapped) type reference.
+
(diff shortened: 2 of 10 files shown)


benwilson512 (approved):
Dude you are on fire lately! Thanks for the contribution!

giddie:
Great to see improved alignment to the spec, but I also wasn't expecting previously working queries to be rejected following a minor version bump, even after glancing through the changelog. I'm not sure how best to handle that sort of thing. Maybe a warning on the changelog entry?

maartenvanvliet:
The changelog could definitely be improved to list these sort of spec compliance bug fixes. I've done a PR here https://github.com/absinthe-graphql/absinthe/pull/1143 but more improvements are welcome.

benwilson512:
@giddie I think that's a fair concern. We could retire 1.6.7 and do a 1.7.0 release. @maartenvanvliet thoughts?

maartenvanvliet:
I think a 1.7 release makes sense. There may be other spec violation fixes that would break backwards compatibility but don't think there are any as impactful in day to day graphql usage as this one.

Some others like https://github.com/absinthe-graphql/absinthe/issues/921 would warrant clear opt-in behaviour

benwilson512:
1.6.7 has been retired, and 1.7.0 has been released.

blakedietz:
If this is a potentially breaking change, why is this being pinned at 1.7.0 instead of 2.0.0?

PR #1142, opened 2022-01-18
Ensure input/output types are correctly placed for SDL schemas

maartenvanvliet:
The InputOutputTypesCorrectlyPlaced phase did not work for
SDL schema's since it would unwrap a TypeReference, which
yields the TypeReference and not the atom type identifier.

The subsequent lookup in the schema types would therefore fail, as
it expects atoms not typereferences, and it would not add errors.

It now correctly fetches the atom type identifier, though in a bit of
a roundabout way.

### DIFF
--- a/test/absinthe/schema/rule/input_output_types_correctly_placed_test.exs
+++ b/test/absinthe/schema/rule/input_output_types_correctly_placed_test.exs
@@ -1,7 +1,7 @@
 defmodule Absinthe.Schema.Rule.InputOutputTypesCorrectlyPlacedTest do
   use Absinthe.Case, async: true
 
-  describe "rule" do
+  describe "macro schema" do
     test "is enforced with output types on arguments" do
       assert_schema_error("invalid_output_types", [
         %{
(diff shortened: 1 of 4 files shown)


test/absinthe/schema/rule/input_output_types_correctly_placed_test.exs
```diff
@@ -56,4 +56,74 @@ defmodule Absinthe.Schema.Rule.InputOutputTypesCorrectlyPlacedTest do
...
+              line: 4
+            }
+          ],
+          phase: Absinthe.Phase.Schema.Validation.InputOutputTypesCorrectlyPlaced
+        }
+      ])
+    end
+
+    test "is enforced with input types on arguments" do
+      assert_schema_error("invalid_input_types_sdl", [
```
maartenvanvliet, line 96: This test asserts there are two errors, whilst the macro based schema only has one. This points to an issue with the macro based schema in that input object fields are `FieldDefinition` structs and not `InputValueDefinition` structs like there are in the SDL schema.

benwilson512:
Thank you!

PR #1145, opened 2022-01-19
Parse constant values separate from locations that permit variables

maartenvanvliet:
The graphql spec (https://spec.graphql.org/draft/#sec-Appendix-Grammar-Summary.Source-Text)
declares that in some location a value cannot be a variable.

This is especially the case in the schema definition language, where
setting the values of directive arguments to a variable is not permitted.

Also in executable documents, setting the default value of a variable definition
to a variable is not permitted.

See https://github.com/graphql/graphql-spec/issues/918#issuecomment-1010248904
for more on this.


kdawgwilk (approved):

benwilson512:
Great, thanks!

PR #1146, opened 2022-01-20
Update non_null.ex with details about adding and removing

alanvardy:
Hello and thank you for the project.

I was confused around adding and removing non_null now that the correctness is more strictly enforced with 1.7.0. I received clarification from @benwilson512 and hope that this edit helps. I can reword this if anyone can think of a better way of putting it.

### DIFF
--- a/lib/absinthe/type/non_null.ex
+++ b/lib/absinthe/type/non_null.ex
@@ -6,7 +6,9 @@ defmodule Absinthe.Type.NonNull do
   By default, all types in GraphQL are nullable. To declare a type that
   disallows null, wrap it in a `Absinthe.Type.NonNull` struct.
   
-  Adding non_null/1 to a type is a breaking change, removing is not.
+  Adding non_null/1 to a type is a breaking change, removing it is not. Client documents that specify non null on a 
+  variable eg `query ($id: ID!)` are allowed to be passed to arguments which allow null. If the argument however is
+  non_null, then the variable type MUST be non null as well.
 
   ## Examples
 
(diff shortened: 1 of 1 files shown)


lib/absinthe/type/non_null.ex
```diff
@@ -5,6 +5,8 @@ defmodule Absinthe.Type.NonNull do
 
   By default, all types in GraphQL are nullable. To declare a type that
   disallows null, wrap it in a `Absinthe.Type.NonNull` struct.
+  
+  Adding non_null/1 to a type is a breaking change, removing is not.
```
benwilson512, line 9:
```suggestion
  Adding non_null/1 to a type is a breaking change, removing it is not. Client documents that specify non null on a 
  variable eg `query ($id: ID!)` are allowed to be passed to arguments which allow null. If the argument however is
  non_null, then the variable type MUST be non null as well.
```

benwilson512:
Thank you!

maartenvanvliet:
There's a bit of nuance here. 

Adding non_null/1 to an **input** type is a breaking change, whereas adding non_null/1 to an **output** type is not. 
The inverse also holds, removing non_null from an **input** type is non-breaking, removing non_null/1 from an **output** type is breaking.

benwilson512:
@maartenvanvliet ah yes good point, I was coming at it from the angle of inputs given the recent bug fix. I'll make an update to handle both situations.

alanvardy:
Very interesting, thank you @maartenvanvliet and @benwilson512 !

PR #1148, opened 2022-01-24
Add `__private__` field to EnumValueDefinition struct

maartenvanvliet:
The EnumValueDefinition is the only schema definition struct that omits the `__private__` field.


benwilson512:
Thanks!

PR #1151, opened 2022-02-12
Return rest in post_traverse to avoid future deprecations

josevalim:


benwilson512:
Thanks @josevalim, what a nice thing to do :)

PR #1155, opened 2022-02-16
Test on new elixir versions

kdawgwilk:
This adds newer elixir and otp versions to testing matrix and fixes a failing test
Closes #1154

PR #1156, opened 2022-02-16
Configure formatter width

andersonmcook:
I wanted to adopt the new formatter for formatting string queries for tests but my team was unsatisfied with how the formatter favored long lines.  This adds some configurability to the width of the formatter.

I have some of the same failing tests locally as https://github.com/absinthe-graphql/absinthe/pull/1155 due to being on Elixir 1.13/OTP 24.

### DIFF
--- a/lib/absinthe/schema/notation/sdl_render.ex
+++ b/lib/absinthe/schema/notation/sdl_render.ex
@@ -5,13 +5,11 @@ defmodule Absinthe.Schema.Notation.SDL.Render do
 
   alias Absinthe.Blueprint
 
-  @line_width 120
-
-  def inspect(term, %{pretty: true}) do
+  def inspect(term, %{pretty: true, width: width}) do
     term
     |> render()
     |> concat(line())
-    |> format(@line_width)
+    |> format(width)
     |> to_string
   end
 
--- a/lib/absinthe/schema/notation/sdl_render.ex
+++ b/lib/absinthe/schema/notation/sdl_render.ex
@@ -5,11 +5,13 @@ defmodule Absinthe.Schema.Notation.SDL.Render do
 
   alias Absinthe.Blueprint
 
-  def inspect(term, %{pretty: true, width: width}) do
+  @line_width 120
+
+  def inspect(term, %{pretty: true}) do
     term
     |> render()
     |> concat(line())
-    |> format(width)
+    |> format(@line_width)
     |> to_string
   end
 
--- a/lib/absinthe/formatter.ex
+++ b/lib/absinthe/formatter.ex
@@ -23,14 +23,18 @@ defmodule Absinthe.Formatter do
     inputs: ["{mix,.formatter}.exs", "{config,lib,test}/**/*.{ex,exs}", "{lib, priv}/**/*.{gql,graphql}"]
   ]
   ```
+
+  ## Options
+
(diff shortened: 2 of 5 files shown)


lib/absinthe/schema/notation/sdl_render.ex
```diff
@@ -5,13 +5,11 @@ defmodule Absinthe.Schema.Notation.SDL.Render do
 
   alias Absinthe.Blueprint
 
-  @line_width 120
-
-  def inspect(term, %{pretty: true}) do
+  def inspect(term, %{pretty: true, width: width}) do
```
kdawgwilk, line 8: This will only match if a width is passed as an opt right?  Do you also need another clause that does not match on width?

andersonmcook: I believe we benefit here from the default [Inspect.Opts](https://hexdocs.pm/elixir/Inspect.Opts.html).

lib/absinthe/schema/notation/sdl_render.ex
```diff
@@ -5,13 +5,11 @@ defmodule Absinthe.Schema.Notation.SDL.Render do
...
     term
     |> render()
     |> concat(line())
-    |> format(@line_width)
+    |> format(width)
```
maartenvanvliet, line 12: Should this not be applied in the `Absinthe.Language.Render` module? You add the width to the Absinthe.Formatter but this is never passed on to the `SDL.Render` module.

There are two intermedate representations in Absinthe of schema/executable documents. The Language structs, which is the result of the lexer/parser. And the Blueprint structs, which is used to build schema's and execute them. 

Both representations can be rendered to the graphql string representation. The Language representation can be converted to executable documents. 
Blueprints can be converted back to the schema definition language.

andersonmcook: Ah, I overlooked that. I'm happy to make the change wherever it makes the most sense to you 😄 .

maartenvanvliet: https://github.com/absinthe-graphql/absinthe/blob/102360eb4dbc289f223cf1cbc64fc01f148721f8/lib/absinthe/language/render.ex#L12 this is probably what you're looking for.

andersonmcook:
I'd be interested in making multiple arguments more vertical as well, if the reviewers/maintainers were amenable.

e.g.

```graphql
query SomeQuery(
  $abc: Boolean
  $def: String!
  $ghi: Int
) {
  someQuery(abc: $abc) {
    aField
    aFieldWithArgs(
      def: $def
      ghi: $ghi
    ) {
      anotherField
    }
  }
}
```

maartenvanvliet:
You'll need to fiddle with the Inspect Algebra representation, specifically the render_list function.

andersonmcook:
> You'll need to fiddle with the Inspect Algebra representation, specifically the render_list function.

Maarten, I noticed your [PR](https://github.com/absinthe-graphql/absinthe/pull/1160) and it gave me inspiration 😄 . There will be an inevitable conflict, but have a look and let me know if this is an OK formatting style.

benwilson512:
Hi @andersonmcook I am open to merging this but there are now merge conflicts as you anticipated. Please reopen once those are fixed, thanks!

PR #1157, opened 2022-02-16
Add SDL/macro type extensions

maartenvanvliet:
Gave it a shot to add support for `extend` to SDL. Once that was running, adding macro support was not that difficult.
It follows the graphql spec for the most part, though schema declarations cannot be extended as of yet. The validations are done by the currently existing validation phases.

I looked at https://github.com/absinthe-graphql/absinthe/pull/821 but decided against the 
```elixir
extend :foo do
   field :name, :string
end
```
 syntax in favor of 
```elixir
extend do 
   object :foo do
      field :name, :string
   end
end
```
since it was easier to implement and is closer to the graphql syntax. I also think it will work better with imports.

Imports across modules work as well, using the `type_extensions` field. It uses the `TypeImports` phase, though it could be moved to a different module if necessary or even a separate `import_type_extensions` macro instead of the `import_types` macro.

Fixes https://github.com/absinthe-graphql/absinthe/issues/772

### DIFF
--- a/lib/absinthe/phase/schema/type_imports.ex
+++ b/lib/absinthe/phase/schema/type_imports.ex
@@ -24,11 +24,11 @@ defmodule Absinthe.Phase.Schema.TypeImports do
 
   def handle_imports(node), do: node
 
-  defp do_imports([], types, schema) do
-    {types, schema}
+  defp do_imports([], types, type_extensions, schema) do
+    {types, type_extensions, schema}
   end
 
-  defp do_imports([{module, opts} | rest], acc, schema) do
+  defp do_imports([{module, opts} | rest], types_acc, type_extensions_acc, schema) do
     case ensure_compiled(module) do
       {:module, module} ->
         [other_def] = module.__absinthe_blueprint__.schema_definitions
@@ -47,10 +47,22 @@ defmodule Absinthe.Phase.Schema.TypeImports do
               types
           end
 
-        do_imports(other_def.imports ++ rest, types ++ acc, schema)
+        type_extensions = other_def.type_extensions
+
+        do_imports(
+          other_def.imports ++ rest,
+          types ++ types_acc,
+          type_extensions ++ type_extensions_acc,
+          schema
+        )
 
       {:error, reason} ->
-        do_imports(rest, acc, schema |> put_error(error(module, reason)))
+        do_imports(
+          rest,
+          types_acc,
+          type_extensions_acc,
+          schema |> put_error(error(module, reason))
+        )
     end
   end
 
--- a/lib/absinthe/phase/schema/type_imports.ex
+++ b/lib/absinthe/phase/schema/type_imports.ex
@@ -6,23 +6,27 @@ defmodule Absinthe.Phase.Schema.TypeImports do
 
   alias Absinthe.Blueprint.Schema
 
-  def run(blueprint, _opts) do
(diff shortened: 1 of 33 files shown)


kdawgwilk:
Could this support this syntax? Since that is what `absinthe_relay` uses and may already be familiar to consumers of absinthe https://hexdocs.pm/absinthe_relay/Absinthe.Relay.Node.html#module-object
```elixir
extend object :foo do
    field :name, :string
end
```
absinthe_relay
```elixir
node object :person do
  field :name, :string
  field :age, :string
end
```

maartenvanvliet:
Ah interesting, forgot about that syntax usage in relay. I'll have a look.

maartenvanvliet:
Thx! Fixed it in [675808a](https://github.com/absinthe-graphql/absinthe/pull/1157/commits/675808abf03c6e2557c2d8af67c166a89f92cf47)

lib/absinthe/schema/notation.ex
```diff
@@ -253,6 +253,48 @@ defmodule Absinthe.Schema.Notation do
...
+
+  ## Examples
+
+  ```
+  object :user do
+    field :name, :string
+    # ...
+  end
+
+  extend do
```
kdawgwilk, line 274: update this doc with the new approach
```elixir
  extend object :user do
    field :nick_name, :string
    # ...
  end
```

maartenvanvliet:
The formatting failures in CI were pre-existing

benwilson512 (changes requested):
Excellent stuff, can you add a changelog entry?

maartenvanvliet:
Thx! I've added an entry 🎉

lib/absinthe/phase/schema/type_imports.ex
```diff
@@ -17,22 +17,35 @@ defmodule Absinthe.Phase.Schema.TypeImports do
...
 
     # special casing the import of the built in directives
     [builtins] = Absinthe.Type.BuiltIns.Directives.__absinthe_blueprint__().schema_definitions
     directives = schema.directive_definitions ++ builtins.directive_definitions
-    {:halt, %{schema | type_definitions: types, directive_definitions: directives}}
+
+    schema = %{
+      schema
+      | directive_definitions: directives,
+        type_extensions: type_extensions,
```
kdawgwilk, line 35: Do you still need this in this phase now that you added that other `ApplyTypeExtensions` phase?

maartenvanvliet: I've extracted the type extension imports to a separate phase, like with import_directives.

maartenvanvliet:
Moved the defaults to the pipeline configuration. This way the `DeprecatedDirectiveFields` phase is no longer needed as the same an be achieved with configuration. 

@kdawgwilk sorry, this clashes a bit with your PR. The same structure can be applied there.

benwilson512:
@maartenvanvliet What is the overall state of this PR?

maartenvanvliet:
> @maartenvanvliet What is the overall state of this PR?

Just fixed a merge conflict with the changelog. Now, it's good to go. 🚀

I have an idea around extending scalar types, so a type extension can be used to make the Int type spec compliant (see https://github.com/absinthe-graphql/absinthe/issues/921) It goes beyond the spec and can be left for another PR though.

benwilson512 (approved):

Issue #1054, opened 2021-03-22
Cannot generate SDL for :persistent_term schema


dolfinus:
### Environment

* Elixir version (elixir -v): 1.11.3
* Absinthe version (mix deps | grep absinthe): 1.6.3
* Client Framework and version (Relay, Apollo, etc): -

### Expected behavior

If I try to execute `absinthe.schema.sdl` or `absinthe.schema.json`, I expect to see a file output:
```bash
absinthe.schema.sdl --schema ArkenstonWeb.Schema graphql/schema.graphql
# graphql/schema.graphql is created and filled up with schema SDL

absinthe.schema.json --schema ArkenstonWeb.Schema --pretty graphql/schema.json
# graphql/schema.json is created and filled up with schema JSON
```

### Actual behavior

It is broken when I use a `:persistent_term` schema:
https://github.com/dolfinus/arkenston-backend/runs/2160087638
```
* creating graphql
* creating graphql/schema.graphql
** (ArgumentError) argument error
    :persistent_term.get({Absinthe.Schema.PersistentTerm, ArkenstonWeb.Schema})
    (absinthe 1.6.3) lib/absinthe/schema/persistent_term.ex:91: Absinthe.Schema.PersistentTerm.get/1
    (absinthe 1.6.3) lib/absinthe/schema/persistent_term.ex:51: Absinthe.Schema.PersistentTerm.__absinthe_type__/2
    (absinthe 1.6.3) lib/absinthe/phase/schema.ex:107: Absinthe.Phase.Schema.set_schema_node/4
    (absinthe 1.6.3) lib/absinthe/phase/schema.ex:71: anonymous fn/4 in Absinthe.Phase.Schema.set_children/3
    (absinthe 1.6.3) lib/absinthe/blueprint/transform.ex:16: anonymous fn/3 in Absinthe.Blueprint.Transform.prewalk/2
    (absinthe 1.6.3) lib/absinthe/blueprint/transform.ex:109: Absinthe.Blueprint.Transform.walk/4
    (elixir 1.11.3) lib/enum.ex:1533: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
```

### Relevant Schema/Middleware Code
https://github.com/dolfinus/arkenston-backend/blob/68b588fa0b56aaed256d4836a136e4cff1b6f25f/lib/arkenston_web/schema.ex#L1-L5
```elixir
defmodule ArkenstonWeb.Schema do
  use Absinthe.Schema

[...]

PR #1159, opened 2022-02-17
Start Absinthe.Schema for persistent term schema

andersonmcook:
Locally we ran into an error generating a JSON schema file. We're using `Absinthe.Schema.PersistentTerm`

```elixir
@schema_provider Absinthe.Schema.PersistentTerm
```

Error:
```
mix absinthe.schema.json --schema OurSchema --pretty path/to/schema.json
** (ArgumentError) errors were found at the given arguments:

  * 1st argument: no persistent term stored with this key

    :persistent_term.get({Absinthe.Schema.PersistentTerm, OurSchema})
    (absinthe 1.6.6) lib/absinthe/schema/persistent_term.ex:94: Absinthe.Schema.PersistentTerm.get/1
    (absinthe 1.6.6) lib/absinthe/schema/persistent_term.ex:54: Absinthe.Schema.PersistentTerm.__absinthe_type__/2
    (absinthe 1.6.6) lib/absinthe/phase/schema.ex:107: Absinthe.Phase.Schema.set_schema_node/4
    (absinthe 1.6.6) lib/absinthe/phase/schema.ex:71: anonymous fn/4 in Absinthe.Phase.Schema.set_children/3
    (absinthe 1.6.6) lib/absinthe/blueprint/transform.ex:16: anonymous fn/3 in Absinthe.Blueprint.Transform.prewalk/2
    (absinthe 1.6.6) lib/absinthe/blueprint/transform.ex:109: Absinthe.Blueprint.Transform.walk/4
    (elixir 1.13.0) lib/enum.ex:1715: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
```

I was able to get it to generate by wrapping the task
```elixir

defmodule Mix.Tasks.Wrapper do
  use Mix.Task

  def run(argv) do
    children = [{Absinthe.Schema, OurSchema}]
    Supervisor.start_link(children, strategy: :one_for_one, name: Our.Supervisor)
    Mix.Tasks.Absinthe.Schema.Json.run(argv)
  end
end
```

I apologize if there was already a built-in or a better way to do this :)

### DIFF
--- a/lib/mix/tasks/absinthe.schema.json.ex
+++ b/lib/mix/tasks/absinthe.schema.json.ex
@@ -80,6 +80,12 @@ defmodule Mix.Tasks.Absinthe.Schema.Json do
 
     opts = parse_options(argv)
 
+    # Start for Absinthe.Schema.PersistentTerm
+    Supervisor.start_link([{Absinthe.Schema, opts.schema}],
+      name: Absinthe.Schema.JsonTaskSupervisor,
+      strategy: :one_for_one
+    )
+
     case generate_schema(opts) do
       {:ok, content} -> write_schema(content, opts.filename)
       {:error, error} -> raise error
--- a/lib/mix/tasks/absinthe.schema.json.ex
+++ b/lib/mix/tasks/absinthe.schema.json.ex
@@ -80,11 +80,10 @@ defmodule Mix.Tasks.Absinthe.Schema.Json do
 
     opts = parse_options(argv)
 
-    # Start for Absinthe.Schema.PersistentTerm
-    Supervisor.start_link([{Absinthe.Schema, opts.schema}],
-      name: Absinthe.Schema.JsonTaskSupervisor,
-      strategy: :one_for_one
-    )
+    # Start Absinthe.Schema if using persistent term provider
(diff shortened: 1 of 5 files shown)


lib/mix/tasks/absinthe.schema.json.ex
```diff
@@ -80,6 +80,12 @@ defmodule Mix.Tasks.Absinthe.Schema.Json do
 
     opts = parse_options(argv)
 
+    # Start for Absinthe.Schema.PersistentTerm
```
kdawgwilk, line 83: Would this also need to be added to `absinthe.schema.sdl`?

andersonmcook: Great question! I'm unsure, we had no problem running `absinthe.schema.sdl` locally.  It was only when we tried `absinthe.schema.json`.

dolfinus: Could you please also add the same change to `absinthe.schema.sql`? This could fix #1054

andersonmcook:
Ah, I see there's CI for Elixir 1.10 but `tmp_dir` was introduced in 1.11.  I can work around that.

benwilson512:
Thank you!

PR #1160, opened 2022-02-17
Convert SDL Language.* structs to SDL notation

maartenvanvliet:
Followup to https://github.com/absinthe-graphql/absinthe/pull/1114

Converts the schema definition language structs back to the string representation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,9 @@
 # Changelog
 
 ## Unreleased
-- Bug Fix: [Add `__private__` field to EnumValueDefinition](https://github.com/absinthe-graphql/absinthe/pull/1148) 
+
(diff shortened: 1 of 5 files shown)


kdawgwilk:
Would this help resolve the issues like https://github.com/absinthe-graphql/absinthe/pull/1110 where behavior is different between SDL and macro schemas?

maartenvanvliet:
No, this just ensures that we can transform the Absinthe.Language intermediate representation to the string form. This is relevant for tooling mostly. The whole Absinthe.Language.* could be extracted to a separate package, so the lexer, parser and intermediate representation can live there and Absinthe can depend on that. 

The Absinthe.Language structs are converted to Blueprint structs in the `import_sdl` notation. Here there are some discrepancies between the macro and sdl notation which leads to issues in the sdl renderer.

benwilson512 (approved):
Looks good to me!

PR #1161, opened 2022-02-20
Fix bug in Schema.__absinthe_types__(:all) for Persistent Term

maartenvanvliet:
Small bug fix, the `Schema.__absinthe_types__(:all)` list contained references instead of names in the persistent term provider. It did not affect anything afaict but it triggered an error when the `test/absinthe/type/object_test.exs` used the persistent term schema provider.

### DIFF
--- a/test/absinthe/type/object_test.exs
+++ b/test/absinthe/type/object_test.exs
@@ -5,6 +5,7 @@ defmodule Absinthe.Type.ObjectTest do
 
   defmodule Schema do
     use Absinthe.Schema
+    use Absinthe.Fixture
 
     query do
       # Must exist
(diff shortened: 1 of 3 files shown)


test/absinthe/type/object_test.exs
```diff
@@ -5,6 +5,7 @@ defmodule Absinthe.Type.ObjectTest do
 
   defmodule Schema do
     use Absinthe.Schema
+    use Absinthe.Fixture
```
benwilson512, line 8: @maartenvanvliet Did this solve a warning or something?

maartenvanvliet: It causes a test failure when the tests are run with PERSISTENT_TERM=true

benwilson512:
Thanks!

PR #1162, opened 2022-03-01
Support types with string identifiers

pulkit110:
There was another PR in the past that tried to answer this (https://github.com/absinthe-graphql/absinthe/pull/483), but it was closed in favor of 1.5. Now that the schema overhaul is done, I would like to propose allowing string type identifiers again. 

When generating a schema at runtime, it is desired that all dynamically generated identifiers be strings instead of atoms to prevent exhausting the `atom` table limit.

This does not change the macros `object` and `input_object` which still require types to be atoms, but when someone overrides the schema to provide runtime type identifiers, it resolves them properly. 

Please let me know if you would like to support string types in the macros as well and I will try to update the PR (as far as I understand, this should be possible, but may not be desirable since they are evaluated at compile time anyway) .

### DIFF
--- a/lib/absinthe/type.ex
+++ b/lib/absinthe/type.ex
@@ -316,7 +316,7 @@ defmodule Absinthe.Type do
 
   @doc "Expand any atom type references inside a List or NonNull"
   @spec expand(reference_t, Schema.t()) :: wrapping_t | t
-  def expand(ref, schema) when is_atom(ref) do
+  def expand(ref, schema) when is_atom(ref) or is_binary(ref) do
     schema.__absinthe_lookup__(ref)
   end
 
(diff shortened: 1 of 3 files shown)


benwilson512:
Hey @pulkit110 supporting string identifiers is definitely something we want to support.

It does sort of look like this PR has a long way to go. Generating the `__absinthe_type__` functions isn't really what you want to do, there is a whole schema construction pipeline now with various backends, just adding clauses to those functions doesn't really address the issue the right way.

I definitely appreciate you taking the time to look into this, and if you're interested in pursuing this further I'd be happy to schedule some time to chat with you about this.

pulkit110:
Hi @benwilson512, thanks for the quick reply on this. I would be happy to chat about this to take this further.

Issue #1166, opened 2022-04-25
Adding an input_object without fields causes schema not to generate and fail silently


MikaAK:
### Environment

* Elixir version (elixir -v): 1.13.3 
* Absinthe version (mix deps | grep absinthe): 1.6.6

When generating the API if you specify an input_object with no fields:
```elixir
input_object :my_input do
end
```
The schema will no longer generate but there's no errors so the only way to track it down is start removing queries/mutations to see what generates

### Expected behavior
Probably some sort of error would be nice saying there's an empty schema

### Actual behavior
Schema just doesn't generate and there's an empty schema in `/graphiql`

### Relevant Schema/Middleware Code
```elixir
mutation do
  field :hello_world, :string do 
    arg :input, :my_input
  end
end

input_object :my_input do
end
```

Issue #1105, opened 2021-09-24
Rendering SDL fails on schema without any fields in RootQueryType


kdawgwilk:
If submitting a bug, please provide the following:

### Environment

* Elixir version (elixir -v): 1.11.3
* Absinthe version (mix deps | grep absinthe): 1.6.5
* Client Framework and version (Relay, Apollo, etc): N/A

### Expected behavior

SDL rendered without issue

### Actual behavior

output was
```
"%Inspect.Error{\n  message: \"got FunctionClauseError with message \\\"no function clause matching in Inspect.Algebra.concat/2\\\" while inspecting %{__struct__: Absinthe.Blueprint, adapter: Absinthe.Adapter.LanguageConventions, directives: [], errors: [], execution: %Absinthe.Blueprint.Execution{acc: %{}, adapter: nil, context: %{}, fields_cache: %{}, fragments: %{},
```

### Relevant Schema/Middleware Code

```elixir
defmodule TestSchema do
  use Absinthe.Schema

  query do
  end
end

inspect(TestSchema, pretty: true)
```

Failing test PR: https://github.com/absinthe-graphql/absinthe/pull/1104

PR #1167, opened 2022-04-29
Add "Objects must define fields" schema validation

maartenvanvliet:
This PR adds an extra schema validation to ensure any object, input_object or interface declares one or more fields. See https://github.com/graphql/graphql-spec/blame/October2021/spec/Section%203%20--%20Type%20System.md#L868 in the spec. Introspection fields are not counted.

It affects a lot of test files since an empty `query do, ... end` is no longer allowed. 

For current users, since this change would be noticed during compilation, the impact is minimal. Some users might encounter invalid schemas. 

fixes https://github.com/absinthe-graphql/absinthe/issues/1166 
fixes https://github.com/absinthe-graphql/absinthe/issues/1105 as schema won't compile


benwilson512 (approved):
This looks great thanks!

kdawgwilk (approved):

benwilson512:
Hey @maartenvanvliet this is getting test failures on the CI now, can you investigate?

maartenvanvliet:
👍 I pushed some fixes

benwilson512:
Thanks!

ryanwinchester:
I think this should have been `[Breaking Bugfix]`?

benwilson512:
Good callout, the CHANGELOG has been updated accordingly.

PR #1169, opened 2022-05-07
Add telemetry events to async middleware

andrewhr:
Introduce telemetry events to Async middleware to instrument new `Task.async/1` created inside the middleware.

The use-case for this event comes from OpenTelemetry and trace propagation. For OTel, we need to run instrumentation from inside the newly created Task to be able to connect spans. With this event in place, we can track those Tasks without introducing custom wrappers on user code.

This patch adds events only in the case of the middleware creating the task. For tasks created by user code, there is nothing the library can do, unfortunately.

### DIFF
--- a/test/absinthe/middleware/async_test.exs
+++ b/test/absinthe/middleware/async_test.exs
@@ -75,12 +75,39 @@ defmodule Absinthe.Middleware.AsyncTest do
     assert {:ok, %{data: %{"asyncBareThing" => "bare task"}}} == Absinthe.run(doc, Schema)
   end
 
-  test "can resolve a field using the normal test helper" do
+  test "can resolve a field using the normal test helper and emit telemetry event", %{test: test} do
     doc = """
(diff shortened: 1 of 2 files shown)


test/absinthe/middleware/async_test.exs
```diff
@@ -75,12 +75,39 @@ defmodule Absinthe.Middleware.AsyncTest do
     assert {:ok, %{data: %{"asyncBareThing" => "bare task"}}} == Absinthe.run(doc, Schema)
   end
 
-  test "can resolve a field using the normal test helper" do
+  test "can resolve a field using the normal test helper and emit telemetry event", %{test: test} do
```
andrewhr, line 78: I extended this existing test with extra telemetry. I'm happy to change to a separate test OR add telemetry assertions on the existing ones - when applicable.

benwilson512:
Great addition, thanks!

PR #1172, opened 2022-05-25
Split telemetry into parse, validate and execution

andrewhr:
This get a bit closer to what we see in other implementations, like the javascript one, where we have explicit phases for parse and validation.

This way, we can instrument Absinthe in a similar fashion, and break down how much work is done inside the framework (by parse/validation) and by the execution engine itself (and resolvers). Example of those phases in [Apollo Federation + JS impl][1].

[1]: https://www.apollographql.com/docs/federation/opentelemetry/#graphql-specific-spans


cschiewek:
Closing as stale.

PR #1174, opened 2022-05-31
Add end_time_mono to telemetry :stop events

cheerfulstoic:
The Spandex library needs a 'completion_time' and it's not possible to get the start time at the beginning of the span to calculate it based on the duration

### DIFF
--- a/lib/absinthe/middleware/batch.ex
+++ b/lib/absinthe/middleware/batch.ex
@@ -149,8 +149,9 @@ defmodule Absinthe.Middleware.Batch do
       timeout = Keyword.get(batch_opts, :timeout, 5_000)
       result = Task.await(task, timeout)
 
-      duration = System.monotonic_time() - start_time_mono
-      emit_stop_event(duration, metadata, result)
+      end_time_mono = System.monotonic_time()
+      duration = end_time_mono - start_time_mono
+      emit_stop_event(duration, end_time_mono, metadata, result)
 
       result
     end)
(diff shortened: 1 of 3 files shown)


andrewhr:
Hey @cheerfulstoic, giving that's not released yet, are you ok with renaming that to `monotonic_time` like in [:telemetry.span/3](https://hexdocs.pm/telemetry/readme.html#spans)?

Reason is, at https://github.com/absinthe-graphql/absinthe/pull/1170 I propose some changes and end up using the `telemetry.span/3` as implementation detail. Assuming changes at that PR are not an option, I still argue for using the same names just for consistency - sort of unofficial convention. 

What do you think?

cheerfulstoic:
So do you mean that `start_time_mono` and `end_time_mono` would both become `monotonic_time`?

I'm fine with it, though if `start_time_mono` is changing I think that would be a breaking change, and I'd love for my change to go out sooner than that 😄   But that's probably still possible

cheerfulstoic:
Yeah, that's fine, but @andrewhr was suggesting changing the key to `monotonic_time`.  Up to you, for sure 😄

andrewhr:
My apologies here, some unnecessary noise due some misunderstanding of mine.

The original intend was to leverage the :telemetry conventions. Both start and end events produce a `monotonic_time` measurement, which is equivalent to ours `start_time_mono` and `end_time_mono`. [You can see details here](https://github.com/beam-telemetry/telemetry/blob/deab9ae2b90242d031466030fb4c83bb5fc559f5/src/telemetry.erl#L312-L318).

The proposal was to use those attributes names. My wrong assumption was both `start_time_mono` and `end_time_mono` weren't released yet, so reverting them would not break any production code.

Sorry again for the noise, and thanks for pointing out the breaking change!

Issue #1064, opened 2021-04-22
Multiple pushes per client for subscriptions that have a `context_id`


mupkoo:
### Environment

* Elixir version (elixir -v): 1.11.3 (compiled with Erlang/OTP 23)
* Absinthe version (mix deps | grep absinthe): 1.6.4

### Expected behavior

Publishing to a subscription should result in just a single push per client.

### Actual behavior

Using a subscription with a `context_id` results in the same number of pushes as the number of connected clients, so for example if you have 5 clients connected and you publish to a given subscription, all 5 clients will receive the data 5 times each.

This is not true when there is no `context_id`

### Relevant Schema/Middleware Code

I have created a repository to showcase the problem.
https://github.com/mupkoo/absinthe-duplicate-subscriptions

Here is a quick view of `schema.ex`

```elixir
defmodule ShowcaseWeb.Schema do
  use Absinthe.Schema

  object :item do
    field :title, :string
  end

  query do
  end

  subscription do
    field :change, :item do
      arg :id, non_null(:id)
      config fn %{id: id}, _ -> {:ok, topic: id, context_id: "global"} end
    end

    field :change_private, :item do
      arg :id, non_null(:id)
      config fn %{id: id}, _ -> {:ok, topic: id} end
    end

    field :change_no_args, :item do
      config fn _, _ -> {:ok, topic: "*", context_id: "global"} end
    end
  end
end
```

This is the test case that showcases the behaviour.
https://github.com/mupkoo/absinthe-duplicate-subscriptions/blob/master/test/showcase_web/schema_test.exs#L5-L38

---

I have mentioned this problem on the `absinthe_phoenix` package as well
https://github.com/absinthe-graphql/absinthe_phoenix/issues/83#issuecomment-823925521

benwilson512:
Thank you so much for creating an example! I will look into this today.

youroff:
We observe the same behavior and even worse: turning `context_id: "global"` made messages from adjacent topics leak to the users who weren't subscribed.

v-anyukov:
In addition it looks like deduplication itself does not work anymore. Resolution happens for every subscribed client, which leads to performance issues.

mupkoo:
It seems like this commit is disabling the test that makes sure that de-duplications work as expected

https://github.com/absinthe-graphql/absinthe/pull/775/files#diff-2d1ba39315325b35cec6a60875bc11082cfbaf47c74482abbc9907735b79c90aR396

[...]

PR #1175, opened 2022-06-27
fix: multiple pushes per client for subscriptions that have a context_id

girishramnani:
Dedups the Registry entries if they all have same topics and keys.

Fixes https://github.com/absinthe-graphql/absinthe/issues/1064

Note - This is not the best solution.

Explained the bug here https://github.com/absinthe-graphql/absinthe/issues/1064#issuecomment-1166711616

### DIFF
--- a/lib/absinthe/subscription/local.ex
+++ b/lib/absinthe/subscription/local.ex
@@ -26,6 +26,9 @@ defmodule Absinthe.Subscription.Local do
           {topic, doc} <- get_docs(pubsub, field, mutation_result, key_strategy) do
         {topic, key_strategy, doc}
       end
+      |> Enum.dedup_by(fn {topic, key_strategy, _} ->
+        "#{topic}_#{:erlang.phash2(key_strategy)}"
+      end)
 
     run_docset(pubsub, docs_and_topics, mutation_result)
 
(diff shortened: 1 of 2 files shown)


halfdan (approved):

LetThereBeDwight:
Can we get this reviewed and brought in? We're using absinthe and would love to see it.

girishramnani:
And I can fix any comments on this PR, have the bandwidth to do so.

LetThereBeDwight:
Looks like @halfdan approved it prior to the test addition commit?

girishramnani:
So can we ping @halfdan to approve again?

benwilson512:
Halfdan is not part of this project so his approval is not particularly relevant here. I'll look at this now.

benwilson512 (changes requested):
So if you comment out these lines:

```
      |> Enum.dedup_by(fn {topic, key_strategy, _} ->
        "#{topic}_#{:erlang.phash2(key_strategy)}"
      end)
```

And then run the tests, everything passes anyway, which means the tests aren't actually testing the code here.

Please write some tests that first _fail_ without the changes, and then show that the changes fix the test.

benwilson512 (changes requested):
So if you comment out these lines:

```
      |> Enum.dedup_by(fn {topic, key_strategy, _} ->
        "#{topic}_#{:erlang.phash2(key_strategy)}"
      end)
```

And then run the tests, everything passes anyway, which means the tests aren't actually testing the code here.

Please write some tests that first _fail_ without the changes, and then show that the changes fix the test. You are welcome to reopen at that time.

PR #1176, opened 2022-06-29
Allow schema declaration directives

maartenvanvliet:
In https://github.com/absinthe-graphql/absinthe/pull/1140 support was added for applying type system directives in the macro syntax, and in https://github.com/absinthe-graphql/absinthe/pull/1157 type extensions became possible in both SDL and macro syntax.

One omission in the above is directives on schema declarations, and applying type extensions on schema declarations. The omission prevents a full implementation of Apollo Federation 2.0 on top of Absinthe. This PR aims to improve on adding the support.

In Absinthe macro schema's it's not necessary to use schema declarations. This stays the same. However, when a type extension is applied to the schema declaration it's necessary for a pre-existing declaration. Therefore, the schema declaration is now inferred from the existing query/mutation/subscription objects. Only in the case of an SDL schema, there can be a pre-existing declaration.

The SDL rendering of blueprint schema's created an ad-hoc schema declaration if no declaration was available. This is no longer necessary.

a97a327 also fixes an issue where the description of the schema declaration was the description of the `__schema` object and not of the schema itself. The `__absinthe_schema_declaration__` was added to help out here, and also expose the declaration after the schema is compiled. 

The `schema do ... end``` was added to the macro syntax, though it will only work in an `extend` block. In a followup a toplevel `schema do ... end` could also work.


test/absinthe/schema/type_system_directive_test.exs
```diff
@@ -179,8 +183,7 @@ defmodule Absinthe.Schema.TypeSystemDirectiveTest do
   end
 
   @macro_schema_sdl """
-  "Represents a schema"
-  schema {
+  schema @feature(name: ":schema") {
```
kdawgwilk, line 186: Does this need `extend`?  or for SDL is it support to declare this without extend?

maartenvanvliet: Turns out that SDL supports it without extend.

maartenvanvliet:
For completeness I've also added support for macro based `schema` declaration. Also expanded support for extending builtin types such as `query`

kdawgwilk (approved):

benwilson512:
Fantastic, thanks!

Issue #1177, opened 2022-06-30
Directives can be repeated even if they are not marked repeatable


kdawgwilk:
We likely will need to add a directive validation phase that ensures that a non-repeatable directive cannot be added more than once to a field 

```elixir
defmodule MyApp.MySchema do
  use Absinthe.Schema

  object :user do
    field :whole_name, :string, deprecate: "Use fullName instead please" do
      deprecate("Use fullName instead")
      ...
    end
  end
  ...
end
```
running `mix absinthe.schema.sdl --schema MyApp.MySchema` gives the following invalid schema
```graphql
type User {
  wholeName: String @deprecated(reason: "Use fullName instead please") @deprecated(reason: "Use fullName instead")
  ...
}
...
```
`@deprecated` is not repeatable so this should not be allowed

maartenvanvliet:
There's a validation phase to check this. It's probably only present in the document pipeline and not in the schema pipeline.

PR #1179, opened 2022-07-13
Validate repeatable directives on schema

maartenvanvliet:
This PR adds validations to check whether a directive can be applied multiple times on a node. This was already done in the document pipeline, but not in the schema pipeline.

Also includes a small fix for the `deprecate` macro requiring a deprecation reason. In the graphql spec the reason is not required, so we can also allow `nil` here.

Fixes https://github.com/absinthe-graphql/absinthe/issues/1177

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Breaking Bugfix: [Validate repeatable directives on schemas](https://github.com/absinthe-graphql/absinthe/pull/1179)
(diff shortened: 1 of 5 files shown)


benwilson512 (changes requested):
This looks great, can we add a `[Breaking Bugfix]` line to the `changelog` about this?

maartenvanvliet:
yes. done :)

kdawgwilk (approved):
Thanks for this! ❤️

benwilson512 (approved):

PR #1180, opened 2022-07-14
Raise informative error on batched resolver timeout

jordi-chacon:
When the Task that runs a batched resolver times out we currently raise a generic error like this:
```
** (exit) exited in: Task.await(%Task{owner: #PID<0.15353.21>, pid: #PID<0.15380.21>, ref: #Reference<0.2441152191.2237136897.192020>}, 5000)
```
This provides no context to the engineer debugging the error about the resolver that timed out, making debugging much harder.

This change raises a `RuntimeError` which includes the resolver that timed out along with the arguments it received:
```
Batch resolver timed out after 1 ms.
Batch fun: {Absinthe.Middleware.BatchTest.Schema, :slow_field_by_user_id}
Batch data: [3, 2, 1]
Batch opts: [timeout: 1]
```
This should make the debugging experience a lot better.

### DIFF
--- a/lib/absinthe/middleware/batch.ex
+++ b/lib/absinthe/middleware/batch.ex
@@ -147,7 +147,19 @@ defmodule Absinthe.Middleware.Batch do
     end)
     |> Map.new(fn {batch_opts, task, start_time_mono, metadata} ->
       timeout = Keyword.get(batch_opts, :timeout, 5_000)
-      result = Task.await(task, timeout)
+
+      result =
+        try do
+          Task.await(task, timeout)
+        catch
+          :exit, {:timeout, {Task, :await, _}} ->
+            raise """
+            Batch resolver timed out after #{timeout} ms.
+            Batch fun: #{inspect(metadata.batch_fun)}
+            Batch data: #{inspect(metadata.batch_data)}
+            Batch opts: #{inspect(metadata.batch_opts)}
+            """
+        end
 
       end_time_mono = System.monotonic_time()
       duration = end_time_mono - start_time_mono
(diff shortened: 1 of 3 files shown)


benwilson512:
This is a great change, can you add a changelog entry?

lib/absinthe/middleware/batch.ex
```diff
@@ -147,7 +147,19 @@ defmodule Absinthe.Middleware.Batch do
...
     |> Map.new(fn {batch_opts, task, start_time_mono, metadata} ->
       timeout = Keyword.get(batch_opts, :timeout, 5_000)
-      result = Task.await(task, timeout)
+
+      result =
+        try do
+          Task.await(task, timeout)
+        catch
+          :exit, {:timeout, {Task, :await, _}} ->
+            raise """
```
benwilson512, line 156: I think the main thing I need to sort out is whether turning this exit into a raise is going to be an issue.

`Task.await/2` emits an exit when it times out, instead of a `RuntimeError` that `raise` emits.

@maartenvanvliet can you think of any reason why this could be a problem?

rossvz:
Bumping an old PR to see what it would take to get this merged. We see hundreds of these errors and we use batch resolvers everywhere so its hard to track down! 

If there's hesitations around raise vs emitting exit - could the code be tweaked to `Logger.error` or something similar?

cschiewek:
@rossvz if you fix the conflicts I'd be happy to move this forward.

rossvz:
@jordi-chacon Since youre the original author - would you want to take a stab at resolving conflicts?

katafrakt:
Wasn't this addressed in https://github.com/absinthe-graphql/absinthe/pull/1298 already?

cschiewek:
> Wasn't this addressed in #1298 already?

Yep.  Thanks!  Closing!

PR #1182, opened 2022-07-25
update old specification links

addstar34:
I was reading through the tutorial and a link to the specification didn't exist so I've gone an updated all the old links to the new ones. I also updated a small snippet of text from the spec.


benwilson512:
Thanks!

Issue #856, opened 2020-01-25
Custom parse errors for scalars


evanreierson:
In [the custom scalars guide](https://github.com/absinthe-graphql/absinthe/blob/master/guides/custom-scalars.md) it says:

> It's important to note that---currently---the correct result from a scalar's parse function in the event of the error is a lone atom, :error, not an error tuple with a reason. In a future version of Absinthe, custom parse errors may be supported. 

I think this is an important feature because clients should be able to know why a scalar isn't parsing to make debugging easier for them.

I am interested in implementing this feature, but I will need a bit of guidance because I am fairly new to elixir, absinthe, and open source contributions in general.

benwilson512:
Hi @evanreierson the first place to start would be to sort out if the GraphQL spec allows for custom parsing error messages, or if the language is still standardized. When we first wrote this there was particular language that scalars which failed to parse were supposed to use. Maybe that's changed now, you'd have to check.

evanreierson:
Thanks @benwilson512, I did not find this standardized language for custom parsing in the most recently published or draft version of the spec. It just says to return a query error when input coercion goes wrong or return a field error when result coercion fails.

In the [Response Errors section](http://spec.graphql.org/June2018/#sec-Errors) though, there is some language that seems to suggest that providing the ability to be more descriptive with error messages is encouraged.

> If the data entry in the response is not present, the errors entry in the response must not be empty. It must contain at least one error. The errors it contains should indicate why no data was able to be returned.

> Every error must contain an entry with the key message with a string description of the error intended for the developer as a guide to understand and correct the error.

These quotes are referring to the top level error ( "... Expected type X, found Y.") but there is also the option to add extensions, where we could put more details on why the coercion failed:

> GraphQL services may provide an additional entry to errors with key extensions. This entry, if set, must have a map as its value. This entry is reserved for implementors to add additional information to errors however they see fit, and there are no additional restrictions on its contents.

I may be missing something but I believe this feature would comply with the spec as long as we put the additional info in extensions. How should I continue?

evanreierson:
I did some more digging into Absinthe errors and the GraphQL spec and it appears that Absinthe errors might be out of sync with the spec. This might need to be addressed in another issue or maybe it's not a problem, but:

![image](https://user-images.githubusercontent.com/31623887/74281036-ef530b80-4cda-11ea-9393-3c038c99d773.png)
[source](https://spec.graphql.org/June2018/#sec-Errors)

[...]

PR #1187, opened 2022-08-26
Custom scalar parser errors

maartenvanvliet:
See https://github.com/absinthe-graphql/absinthe/issues/856

The `parse/1` could only return `:error` in case of a failure when parsing a custom scalar. This PR fixes that by also allowing an error tuple as a return value. 

The error is stored in the flag instead of the `errors` field. This was more in line with how the current code was set up. The cases where the parse error handles in input objects and input lists are also handled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Feature: [Support error tuples when scalar parsing fails](https://github.com/absinthe-graphql/absinthe/pull/1187)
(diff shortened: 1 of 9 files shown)


benwilson512:
@maartenvanvliet I'm generally 👍 to this, what's the consensus amongst the other implementations with respect to this feature?

maartenvanvliet:
The graphql reference implementation supports custom errors, e.g.: https://github.com/graphql/graphql-js/blob/main/src/type/scalars.ts#L29

The ruby implementation as well: https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/types/iso_8601_date_time.rb#L49

benwilson512:
Great, thanks!

PR #1189, opened 2022-08-29
Expose type system directives in introspection

maartenvanvliet:
Previously type system directives such as `deprecated` or those created by users were not available in introspection. The spec itself is not really clear in this but afaict they should be exposed.

I've  checked out the reference ](https://44c20.sse.codesandbox.io/?query=query%20%7B%0A%20%20__schema%7B%0A%20%20%20%20directives%20%7B%0A%20%20%20%20%20%20name%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D%0A)`graphql` implementation and it does expose the type system directives.

This PR adds a phase to add the prototype directive list to the schema directive list so they are available through introspection. In the prototype schema itself this phase is skipped to prevent an infinite loop.


kdawgwilk (approved):

benwilson512:
Thank you!

Issue #1190, opened 2022-08-30
"understanding subscriptions" documentation needs to be updated


Qqwy:
The ['Understanding Subscriptions'](https://hexdocs.pm/absinthe/subscriptions.html) page in the documentation explains how to start using subscriptions.

However, a couple of parts seem out of date:
- The default PubSub settings in the config seem to use a different syntax than the current one.
- The page lists changes to make for phoenix 1.5 and 1.4 and in places 1.3. Not all of the steps that are accurate for 1.5 seem to translate to 1.6 however. The part about adding your socket only shows changes for v1.3/v1.4 and v1.2 and before.
- The page guides you to include `absinthe_phoenix ~> 1.5`. However, the current version is 2.0.2.

Qqwy:
After trying to follow the instructions best as possible, I get the following compile-time error. I think it might be related, but maybe it is not.

```elixir
== Compilation error in file lib/heartbeat_web/schema.ex ==
[...]

PR #1191, opened 2022-08-30
Raise for non-toplevel root object definitions

maartenvanvliet:
As seen in https://github.com/absinthe-graphql/absinthe/issues/1190#issuecomment-1231589935 a more informative error can be shown when root objects are note used in the toplevel.


Qqwy:
Wow, thanks a lot for the fast work! :green_heart:

benwilson512 (approved):
Great, thanks!

Issue #918, opened 2020-05-07
Support `specifiedBy` directive


binaryseed:
There's a new directive that will soon be in the spec for Custom Scalars called `specifiedBy`

https://github.com/graphql/graphql-spec/pull/649

Example:

```graphql
scalar UUID @specifiedBy(url: "https://tools.ietf.org/html/rfc4122")
scalar URL @specifiedBy(url: "https://tools.ietf.org/html/rfc3986")
```

PR #1193, opened 2022-09-07
Add the specifiedBy type system directive

maartenvanvliet:
Adds the specifiedBy type system directive to Absinthe. See the spec https://spec.graphql.org/October2021/#sec--specifiedBy for its definition.

Also encountered an issue with directive names not being transformed through the adapter in introspection. 

Fixes https://github.com/absinthe-graphql/absinthe/issues/918


kdawgwilk (approved):

lib/absinthe/type/built_ins/introspection.ex
```diff
@@ -54,7 +54,11 @@ defmodule Absinthe.Type.BuiltIns.Introspection do
   object :__directive do
     description "Represents a directive"
 
-    field :name, non_null(:string)
+    field :name,
+      type: non_null(:string),
+      resolve: fn _, %{adapter: adapter, source: source} ->
+        {:ok, adapter.to_external_name(source.name, :field)}
```
benwilson512, line 60: Good catch.

benwilson512 (approved):
This looks good to me!

PR #1197, opened 2022-09-19
Add support for OTP-25 and Elixir 1.14

maartenvanvliet:
This PR adds support for OTP-25 and Elixir 1.14 in the test matrix. In addition it removes support for Elixir 1.10. Support is kept for Elixir 1.11 which seems plenty.

One change in OTP-25 caused some [test failures](https://github.com/maartenvanvliet/absinthe/actions/runs/3051327865) in the SDL rendering. See f64dd3866162b316b41258c6558290e57ff2f601 for an explanation.

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -14,20 +14,25 @@ jobs:
     strategy:
       matrix:
         elixir:
-          - "1.10"
           - "1.11"
           - "1.12"
           - "1.13"
+          - "1.14"
         otp:
           - "22"
           - "23"
           - "24"
+          - "25"
         exclude:
-          - elixir: "1.10"
-            otp: "24"
+          - elixir: "1.11"
+            otp: "25"
+          - elixir: "1.12"
+            otp: "25"
+          - elixir: "1.14"
+            otp: "22"
         include:
-          - elixir: "1.13"
-            otp: "24"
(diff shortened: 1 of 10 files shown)


oliver-kriska:
hello, thanks for contribution. I would like to ask you about any timeline for merging this PR and release in hex? Thanks

kdawgwilk (approved):

oliver-kriska:
thanks for approving it. Any plans for merging and releasing new version?

maartenvanvliet:
There's an existing test to showcase it works transitively but I agree it can be impactful. I am not a heavy user of import_fields so I cannot judge how it will fare in schema's that use it a lot.

There's one thing that irks me is that I cannot find exactly when the `topsort/1` behaviour changed in OTP. I found https://github.com/erlang/otp/commit/c5e3d13f0325f8caa155bb7cb7298e749b594d50 which could have some impact but this commit was already in OTP-23. That's the only relevant change in digraph/digraph_utils I could find. Note, the results are not wrong just in a different order.

Another course of action is to revert the changes and go back to `topsort/1`. We could skip the impacted tests for OTP < 25. The SDL rendering will just be in a different order compared to OTP 25. An explanatory note in the documentation can be added to explain the difference.

oliver-kriska:
is it possible from my side somehow help with this PR? :)

PR #1199, opened 2022-09-30
Fix typos

afhammad:
Fixed some minor typos noticed while going through the code


lib/absinthe/adapter/underscore.ex
```diff
@@ -1,6 +1,6 @@
 defmodule Absinthe.Adapter.Underscore do
   @moduledoc """
-  Underscores external input and leaves external input alone. Unlike the
+  Underscores internal input and leaves external input alone. Unlike the
```
benwilson512, line 3: I don't think this is accurate. It underscores `external` input via `to_internal_name`.

benwilson512 (commented):
Hi I'm gonna just do the `customer` thing in master and leave the adapter description alone.

PR #1201, opened 2022-10-13
Allow `Registry` compression to be configured by users

c-brenn:
At the moment `Registry` is started with `compressed: true` hard coded.

This presented problems when we updated absinthe some time ago, we are running an application with a large number of subscriptions and so the added CPU utilization was quite large. To avoid having to scale up we pinned absinthe to a pre-compression version.

We'd like to be able to keep up to date, and so propose this change to allow users to opt-out of compression and would welcome your feedback.

This PR alters the type signature of `Absinthe.Subscription.child_spec/1` so users can pass more configuration options. The original argument (a single pub-sub) is still supported for backwards compatibility.

### DIFF
--- a/guides/subscriptions.md
+++ b/guides/subscriptions.md
@@ -36,7 +36,7 @@ line:
       {Phoenix.PubSub, name: :my_pubsub},
       # Start the endpoint when the application starts
       MyAppWeb.Endpoint,
-      {Absinthe.Subscription, MyAppWeb.Endpoint}
+      {Absinthe.Subscription, pubsub: MyAppWeb.Endpoint}
     ]
 
     # See https://hexdocs.pm/elixir/Supervisor.html
@@ -44,6 +44,9 @@ line:
     opts = [strategy: :one_for_one, name: MyAppWeb.Supervisor]
     Supervisor.start_link(children, opts)
 ```
+
+See `Absinthe.Subscription.child_spec/1` for more information on the supported
+options.
 
 In your `MyAppWeb.Endpoint` module add:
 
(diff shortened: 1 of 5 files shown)


guides/subscriptions.md
```diff
@@ -36,7 +36,7 @@ line:
       {Phoenix.PubSub, name: :my_pubsub},
       # Start the endpoint when the application starts
       MyAppWeb.Endpoint,
-      {Absinthe.Subscription, MyAppWeb.Endpoint}
+      {Absinthe.Subscription, pubsub: MyAppWeb.Endpoint}
```
benwilson512, line 39: Changing this is a breaking change. We can _add support_ for this style, but the implementation you have here removes support for just passing in the module.

benwilson512 (changes requested):
Hi @c-brenn this is a reasonable request, but we need to do it in a way that preserves support for the existing `{Absinthe.Subscription, MyAppWeb.Endpoint}` style children.

guides/subscriptions.md
c-brenn: The implementation of `child_spec/1` in this PR should maintain compatibility with "just passing  module" - it accepts either an atom or a keyword list:

https://github.com/absinthe-graphql/absinthe/blob/70bd17dcd443cc6b1edee1ce1ca0e74d0ffd5644/lib/absinthe/subscription.ex#L62-L76

However this isn't true for `Absinthe.Subscription.start_link/1` - it now expects a keyword list, but I can change it to also accept just a module if you'd like 🙇

PR #1202, opened 2022-10-17
Fix broken links and documentation improvements

tofran:
It was raised up by a college of mine that the feature I added a while ago via the PR #1131 had a broken URL in the documentation.
I didn't want to create a PR just to add an `h`, so I went around the documentation and tried to fix/improve a few things:

- 85c5138ddf4f2d8410e8f5daa74c6ffa56052eac Fixed my typo and also found a legacy link in the changelog, updated it to the correct version at the time;
- e4d83b6e2280680558f992e51232a4dded777e9c Updated HTTP links to HTTP**S**. This reduces redirects and avoids having links in the documentation that downgrade the connection to HTTP;
- c652ae8c5954c6e1afb3761fb3a51354f9574704 Removed double spacing;
- d5af25848cedfc0875e12ef4466b089c04a55a60 Timex examples in the documentation were using the old syntax. Updated them to match the latest version (and made a couple naming adjustments);
- 6975b45621f19d5723bc4864ac112f988e7873c8 Removed duplicate module from the documentation index.

Hopefully this will make the documentation better. Please provide your feedback.

### DIFF
--- a/lib/absinthe/type/scalar.ex
+++ b/lib/absinthe/type/scalar.ex
@@ -19,7 +19,7 @@ defmodule Absinthe.Type.Scalar do
 
   ## Examples
 
-  Supporting a time format in ISOz format, using [Timex](http://hexdocs.pm/timex):
+  Supporting a time format in ISOz format, using [Timex](https://hexdocs.pm/timex):
 
   ```
   scalar :time do
--- a/lib/absinthe/type/scalar.ex
+++ b/lib/absinthe/type/scalar.ex
@@ -19,13 +19,13 @@ defmodule Absinthe.Type.Scalar do
 
   ## Examples
 
-  Supporting a time format in ISOz format, using [Timex](https://hexdocs.pm/timex):
+  Supporting datetime in the ISO8601 format for UTC only, using [Timex](https://hexdocs.pm/timex):
 
   ```
-  scalar :time do
-    description "Time (in ISOz format)"
-    parse &Timex.DateFormat.parse(&1, "{ISOz}")
-    serialize &Timex.DateFormat.format!(&1, "{ISOz}")
+  scalar :isoz_datetime do
+    description "UTC only ISO8601 date time"
+    parse &Timex.parse(&1, "{ISO:Extended:Z}")
+    serialize &Timex.format!(&1, "{ISO:Extended:Z}")
   end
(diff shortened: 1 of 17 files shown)


kdawgwilk (approved):

lib/absinthe/type/scalar.ex
```diff
@@ -19,13 +19,13 @@ defmodule Absinthe.Type.Scalar do
...
 
   ```
-  scalar :time do
-    description "Time (in ISOz format)"
-    parse &Timex.DateFormat.parse(&1, "{ISOz}")
-    serialize &Timex.DateFormat.format!(&1, "{ISOz}")
+  scalar :isoz_datetime do
+    description "UTC only ISO8601 date time"
+    parse &Timex.parse(&1, "{ISO:Extended:Z}")
+    serialize &Timex.format!(&1, "{ISO:Extended:Z}")
```
benwilson512, line 28: If we're updating this, we may as well switch this over to using the built in `DateTime` functions.

benwilson512 (changes requested):

lib/absinthe/type/scalar.ex
tofran: @benwilson512 I also considered the same thing. But since `datetime` it is already implemented in [Absinthe.Type.Custom](https://hexdocs.pm/absinthe/Absinthe.Type.Custom.html) I tried to maintain it as it was originally created: for solving a custom scenario that is not already included with absinthe.

Absinthe's `datetime` will do DateTime.from_iso8601(value), meaning it is fully ISO8601 compliment. In the example above it actually only accepts ISO8601 in the UTC time-zone, hence the "UTC only" mention.
IE: Will accept `2022-10-17T23:27:45.076302Z` but not `2022-10-17T23:27:45.076302+1000`.

I don't want to encourage users to re-implement the datetime type. Maybe I should add a note or redirect users to `Absinthe.Type.Custom`?

tofran: I've just made it clearer with f617cf40367bed55b4284ef97b2bb465d1c3841d
What do you think?

benwilson512 (approved):
Sure, I think that change makes sense.

benwilson512:
Thanks!

PR #1204, opened 2022-11-08
Move import_types etc to Absinthe.Schema 

maartenvanvliet:
Following a discussion https://elixirforum.com/t/can-you-create-circular-dependancies-when-using-import-types/51620/3
the `import_types` etc. macro's can be confusing as to where they can be placed.

In this PR they are moved to the Absinthe.Schema file so they cannot be used in the type modules. Instead an error is raised when they are used there. Since usage in the type modules was never supported this is a non-breaking change.

The `schema` macro is also moved to Absinthe.Schema, this also does not make sense to use in the type modules.

Furthermore, the `group: :notation` was added to all notation macro's to make it clearer in the docs.

Screenshot 2022-11-08 at 14 23 29

### DIFF
--- a/lib/absinthe/schema.ex
+++ b/lib/absinthe/schema.ex
@@ -180,6 +180,19 @@ defmodule Absinthe.Schema do
   @placement {:query, [toplevel: true, extend: true]}
   @doc """
   Defines a root Query object
+
+  ## Options
+    * `:name` - The name of the root query type. Defaults to `#{@default_query_name}`
+
+  ## Example
+
+  ```elixir
+  query do
+    field :posts, list_of(:post) do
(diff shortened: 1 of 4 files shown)


benwilson512:
I think we can emit a warning, but in theory this is a breaking change right? There are rare but valid uses for this today, eg:

```
defmodule SomeLib.DateTypes do
  use Absinthe.Schema.Notation
end
defmodule SomeLib.NumberTypes do
  use Absinthe.Schema.Notation
end
defmodule SomeLib.AllTypes do
  use Absinthe.Schema.Notation
  import_types SomeLib.DateTypes
  import_types SomeLib.NumberTypes
end
```

This would allow an end user to `import_types` either the specific sub modules or the main module.

PR #1205, opened 2022-11-08
Fix const parsing of empty lists

maartenvanvliet:
An empty list is a valid const value. This fixes the parsing of empty lists.


benwilson512 (approved):
Thank you!

Issue #1149, opened 2022-01-28
resolve_type second parameter is just some other resolved object?


marcinkoziej:
Hi!
Reporting a potential bug:

### Environment

Erlang/OTP 24 [erts-12.0.3] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [jit]
Elixir 1.12.0 (compiled with Erlang/OTP 23)
mix deps |grep absinthe
* absinthe 1.7.0 (Hex package) (mix)
  locked at 1.7.0 (absinthe) 566a5b55
* absinthe_phoenix 2.0.2 (Hex package) (mix)
  locked at 2.0.2 (absinthe_phoenix) d3691892
* absinthe_plug 1.5.8 (Hex package) (mix)
  locked at 1.5.8 (absinthe_plug) bbb04176
Client framework: GrapqiQL playground

### Expected behavior

When resolving a list of interfaces, I expect that `resolve_type(object, env)` receives a second argument that contains information about parent/location of object in the graph - consistently for all items. 

### Actual behavior

I have a problem when resolving type of a list of objects.

I have a following graph:
```
(  campaign  ) -has-many-> (  targets ) -has-many-> (target emails)
  / can be \                 / can be \
public     private        public     private 
campaign   campaign       target     target 
```

If you have a permission to see campaign (interface) as private campaign (object), you can also see the target (iface) as private target (object).
Target emails are only visible on private target.

In target interface definition, i want to check what the parent (campaign) was resolved to either `:public_campaign` or `:private_campaign`

```
  interface :target do
    field :id, non_null(:string)
    field :name, non_null(:string)
   # .... 
    resolve_type(fn
      _, %{parent_type: %{identifier: :private_campaign}} ->  :private_target
      _, %{parent_type: %{identifier: :public_campaign}} -> :public_target
      _, %{parent_type: _other} -> :public_target 
    end)
  end
```
When doing a query like this:

```
campaign(id: 123) {
  ... on PrivateCampaign {
    targets {
      ... on PrivateTarget {
        emails {
          email
        } 
      }
    }
  }
}
```
When resolving all items in the target list for particular campaign, only the FIRST target is resolved as `:private_target` - and the resolve_type is given the object where I can match on: `%{parent_type: %{identifier: :private_campaign}}`.
The SECOND target receives something else as second argument - an object describing `target_email` object, it seems to be a child of FIRST target!
[...]

PR #1208, opened 2022-11-16
Fix bookkeeping of Resolution path when resolving

marcinkoziej:
Addresses issue #1149 
module `Absinthe.Phase.Document.Execution.Resolution` walks the result tree and tracks `path` using a function parameter. At the same time, a `Resolution` struct is passed, which `path` field is not consistently tracked.

The problem manifests itself in bogus Resolution path passed onto `resolve_type` function as described in function #1149.

The reason for not updating the resolution.path in the walk process can be explained by comment of @benwilson512: `Traditionally the reason we pass in the env has more to do with making the schema available and the context available, I don't know that the other values in that struct have been validated in this situation.` Perhaps path parameter was used in the beginning and later resolution paramter was added, but not used to track current tree walk in it's path field.

Nevertheless, the resolution is passed on to user code - into `resolve_type(value, resolution)` callback, which might use path to understand the context of value (eg. what is the type of parent?).

This PR fixes the bookkeeping of `resolution.path` field. I have not inspected the use of other field, I am not sure which ones would change during the value walk (`parent_type` maybe?).

One could consider to just use `resolution.path` instead of `path` parameter - so there is not double bookkeeping of this value. This is up to maintainers to decide, I did not make such a change.

### DIFF
--- a/lib/absinthe/phase/document/execution/resolution.ex
+++ b/lib/absinthe/phase/document/execution/resolution.ex
@@ -115,7 +115,8 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
     walk_results(values, bp_node, inner_type, res, [i + 1 | sub_path], [result | acc])
   end
 
-  defp walk_results([], _, _, res = %{path: [_ | sub_path]}, _, acc), do: {:lists.reverse(acc), %{res | path: sub_path}}
+  defp walk_results([], _, _, res = %{path: [_ | sub_path]}, _, acc),
+    do: {:lists.reverse(acc), %{res | path: sub_path}}
 
   defp resolve_fields(parent, res, source, path) do
     # parent is the parent field, we need to get the return type of that field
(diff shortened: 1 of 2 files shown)


benwilson512:
Hi @marcinkoziej thanks for taking a crack at this. I'm open to moving forward with this but it definitely needs a test case or two. I'll also need some time to get my head back into this module and make sure there aren't any unintended consequences of this change.

Thanks for taking a step towards fixing it!

marcinkoziej:
Should I add it to the Interface test or into a separate file ?

benwilson512:
I think an interface_test would be good, and then any other more general resolution test that could demo the behavior could be interesting.

Please also run `mix format` on the code so that the test suite will run.

marcinkoziej:
Ran mix format and added a test.

marcinkoziej:
@benwilson512 :wave:

benwilson512:
Thank you!

PR #1210, opened 2022-11-30
Limiting token count in parse phase to prevent DoS

stellanor:
Hi! I'm opening this PR to start a conversation about how best to address this vulnerability that could lead to a denial-of-service attack.

An attacker could craft a very long query (which doesn't need to be valid), mentioning thousands of undefined or real fields or directives, which could require a lot of resources to parse. [This same vulnerability was identified and fixed in graphql-java last year](https://github.com/graphql-java/graphql-java/pull/2549).

I saw in my benchmarking that Absinthe will spend a long time in the parse phase for such queries. The lexer's tokenization and the parser themselves are quite performant even on very long queries, but the lexer has an extra step, `convert_token_column`, that processes line number information (for error reporting later?) This routine is O(n^2) and performs poorly for long, single-line queries.

[Here is a sample app I made to demonstrate the issue with Absinthe 1.7.0](https://github.com/stellanor/absinthe_parser_perf_demo)

My Absinthe PR adds a safety check after the lexer tokenizes but before performing `convert_token_column`. The limit is, of course, configurable, with a default.

I don't intend for this to be merged as-is without discussion. I am seeking input on the following:
- The change itself, of course
- What the default token limit should be. This is a tradeoff between safety by default and making a breaking change. I chose a default limit of 15,000 to start the discussion, since [this is what graphql-java chose](https://github.com/graphql-java/graphql-java/pull/2549/files#diff-cc1fab9887f0cf77d8caf9afdf45a006206cf0a572a09b1c4992af7c06ce53f6R23). However, in my tests, it still took over 30 seconds to parse a 15k token single-line query, so we may want to discuss a lower limit.
- The error message used in the phase error when this limit is enforced (I chose a concise error message; should we consider something that is more verbose? I did mention the exact error string in the docs so that someone googling it should find it).
- The documentation: I folded a new section into the existing complexity analysis guide page, but I'm open to making this its own page.
- Coordinating this change with absinthe_plug ([I'm concerned about this file and its docs](https://github.com/absinthe-graphql/absinthe_plug/blob/624696372cdcf72f4cff5ddd8a6bb2f1c89de9df/lib/absinthe/plug.ex#L127))
 
As a separate change, I am also curious about the possibility of managing the line/column attribution differently. Could this somehow be delayed and only performed during error rendering? Perhaps the number of validation errors returned can also be limited? I haven't explored this possibility yet or whether that's even within spec. I'm unlikely to have time to work on that anytime soon, but I wanted to throw the idea out.

My change was pre-reviewed by my colleague @mattbaker . Given the potentially sensitive nature of this, I tried to reach out to @benwilson512 on the Elixir slack last week but we realized Ben's not active there, so we decided to proceed with opening the PR.

Thank you! Any feedback, ideas, or questions are welcome.

### DIFF
--- a/guides/complexity-analysis.md
+++ b/guides/complexity-analysis.md
@@ -121,8 +121,7 @@ calculated and maximum complexities.
 ## Token Limits
 
 To protect a service from invalid queries that could take considerable resources to parse, 
-Absinthe includes a maximum limit on tokens in the GraphQL request document. If the lexer
-encounters more tokens than this, it will stop the parse phase and return a phase error 
-with the message `"Token limit exceeded"`. This limit is 15,000 by default and can be
-overridden by providing the option `token_limit` to `Absinthe.run`. `token_limit` could
-be an integer or `:infinity` for no limit.
+Absinthe offers the option to configure a maximum limit on tokens in the GraphQL request document. 
+If the lexer encounters more tokens than this, it will stop the parse phase and return a phase error 
+with the message `"Token limit exceeded"`. This limit is `:infinity` by default (no limit) 
+and can be overridden by providing an integer to the option `token_limit` to `Absinthe.run`.
--- a/guides/complexity-analysis.md
+++ b/guides/complexity-analysis.md
@@ -1,11 +1,12 @@
-# Complexity Analysis and Token Limits
+# Safety Limits
 
 A misbehaving client might send a very complex GraphQL query that would require
 considerable resources to handle. There are two variations of this problem:
 
-- Complex queries that overwhelm resolution resources
-- Long, invalid queries that could take considerable resources to parse, even if 
-the response is only validation errors.
+- Complex queries that overwhelm resolution resources.
+- Extremely long queries that could take considerable resources to parse. 
+(For example, an attacker could craft a long query including thousands of 
+undefined fields or directives.)
 
 Either of these could be a vector for a denial-of-service attack in any GraphQL 
 service. Absinthe includes mechanisms to protect services each of these.
--- a/guides/complexity-analysis.md
+++ b/guides/complexity-analysis.md
@@ -122,6 +122,6 @@ calculated and maximum complexities.
 To protect a service from invalid queries that could take considerable resources to parse, 
 Absinthe includes a maximum limit on tokens in the GraphQL request document. If the lexer
 encounters more tokens than this, it will stop the parse phase and return a phase error 
-with the message `"Token limit exceeded"`. This limit is 10,000 by default and can be
+with the message `"Token limit exceeded"`. This limit is 15,000 by default and can be
 overridden by providing the option `token_limit` to `Absinthe.run`. `token_limit` could
 be an integer or `:infinity` for no limit.
--- a/guides/complexity-analysis.md
+++ b/guides/complexity-analysis.md
@@ -1,7 +1,18 @@
-# Complexity Analysis
+# Complexity Analysis and Token Limits
 
 A misbehaving client might send a very complex GraphQL query that would require
-considerable resources to handle. In order to protect against this scenario, the
+considerable resources to handle. There are two variations of this problem:
+
+- Complex queries that overwhelm resolution resources
+- Long, invalid queries that could take considerable resources to parse, even if 
+the response is only validation errors.
+
+Either of these could be a vector for a denial-of-service attack in any GraphQL 
+service. Absinthe includes mechanisms to protect services each of these.
+
+## Complexity Analysis
+
+In order to protect against queries that could overwhel scenario, the
 complexity of a query can be estimated before it is resolved and limited to a
 specified maximum.
 
@@ -105,3 +116,12 @@ But this, at a complexity of `60`, wouldn't:
 If a document's calculated complexity exceeds the configured limit, resolution
 will be skipped and an error will be returned in the result detailing the
 calculated and maximum complexities.
+
+## Token Limits
+
+To protect a service from invalid queries that could take considerable resources to parse, 
+Absinthe includes a maximum limit on tokens in the GraphQL request document. If the lexer
+encounters more tokens than this, it will stop the parse phase and return a phase error 
+with the message `"Token limit exceeded"`. This limit is 10,000 by default and can be
+overridden by providing the option `token_limit` to `Absinthe.run`. `token_limit` could
+be an integer or `:infinity` for no limit.
--- a/lib/absinthe/lexer.ex
+++ b/lib/absinthe/lexer.ex
@@ -42,6 +42,7 @@ defmodule Absinthe.Lexer do
   comment =
     string("#")
     |> repeat_while(any_unicode, {:not_line_terminator, []})
+    |> post_traverse({:check_token_limit, []})
 
   # Comma :: ,
   comma = ascii_char([?,])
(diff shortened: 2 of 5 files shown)


benwilson512:
Hi @stellanor thank you very much for the PR and raising this issue. As you saw I am no longer active on the Elixir slack. Let me discuss this with some folks at CargoSense and try to get back to you promptly tomorrow (Wed Nov 29).

I do think this should also get paired with `Plug.Parser` length limits, but that doesn't create protection in every case. We should make sure that `Absinthe.Plug` supports that kind of limit as well.

guides/complexity-analysis.md
```diff
@@ -1,7 +1,19 @@
...
+- Extremely long queries that could take considerable resources to parse. 
+(For example, an attacker could craft a long query including thousands of 
+undefined fields or directives.)
+
+Either of these could be a vector for a denial-of-service attack in any GraphQL 
+service. Absinthe includes mechanisms to protect services each of these.
+
+## Complexity Analysis
+
+In order to protect against queries that could overwhel scenario, the
```
kdawgwilk, line 16:
```suggestion
To protect against queries that could overwhelm available resources, the
```

maartenvanvliet:
This looks good! Hardening Absinthe is something I've been wanting to explore but haven't had time for.

One thing that can be optimised in this PR is that the count of the tokens is done after parsing has completed. This can be moved to inside the tokenizer. Currently, the tokenised does a post_traverse call on each token. In this call we can track a token count in the context and short-circuit the parsing when the limit is hit.

guides/complexity-analysis.md
stellanor: Whoops, not sure how that happened! 😵‍💫 Thank you!

stellanor:
@benwilson512 Thank you! No particular hurry on my end.

> I do think this should also get paired with Plug.Parser length limits, but that doesn't create protection in every case. We should make sure that Absinthe.Plug supports that kind of limit as well.

I'm glad you mentioned this--I've been exploring request length limiting as well, although that's easier for my service to solve outside the libraries, so that's where I started with it. Even with a reasonable token limit (I'm exploring between 3,000-5,000 for my service), I was still able to craft a malicious query with lots of whitespace between tokens because this routine is scanning a line for each token.

My service will be exploring these three limits in combination:
- the Absinthe token limit
- a limit on the query portion of a GraphQL request payload, excluding variables
- an overall HTTP request body length limit (for us this needs to be very high; some of our service's APIs accept very long strings as arguments, but since long literals in the variable payload don't impact the query parsing performance being discussed here this can be a different magnitude limit than the other two)

stellanor:
Thank you, @maartenvanvliet , that is a great suggestion. I will give that a try (later today if I have time).

benwilson512:
@stellanor your point about the HTTP body limits being unable to distinguish between large variables and a large query is a very good one. That seriously limits the ability for Plug.Parsers to mitigate this specific issue.

benwilson512:
@stellanor as a very minor note can you run `mix format` on the PR please so that the CI will run?

stellanor:
Hi @maartenvanvliet , I explored your suggestion to short-circuit the tokenizer when the limit is reached. I am not sure this is possible within NimbleParsec's design--if you have some time, can you let me know if there's something I'm missing?

The sticking point is how to pass the configured limit value through so that it's included in the `context` map used by `post_traversal`. The configurable token limit is defined in the scope of [this line of code](https://github.com/absinthe-graphql/absinthe/pull/1210/files#diff-13c1cdaba0d4a0087248cb40f906f24e0fcc2c219034092c1a98d6cf51826c53L234) which calls `do_tokenize`. The limit is among the options passed in when running the parse phase. But `do_tokenize` is a function generated by a `defparsec` macro call. That appears to be a problem because the `defparsec` macro call happens at compile-time, and it has to conform to a specific signature as well ([docs here](https://hexdocs.pm/nimble_parsec/NimbleParsec.html#defparsec/3)). So I don't see a way that I can pass the runtime limit configuration through so that it is included in the generation of the combinators at compile-time.

maartenvanvliet:
It's a bit of a hidden feature but you can pass the initial context map to `do_tokenize`. See https://hexdocs.pm/absinthe/Absinthe.Lexer.html#do_tokenize/2-options

stellanor:
Hi folks, thanks for your patience! I'm back with an update to this PR that takes @maartenvanvliet 's suggested approach of short-circuiting in the lexer. 

Doing so is a bit more error-prone, since we have to add the counting and checking to several combinators, so I added some fairly meticulous tests and hand-inspected the results carefully today. But let's review with caution.

[One thing I noticed along the way is that `comment` was repeated](https://github.com/absinthe-graphql/absinthe/pull/1210/commits/594f4f3b0a6f066672853459ef470e9033d3d154#diff-13c1cdaba0d4a0087248cb40f906f24e0fcc2c219034092c1a98d6cf51826c53L274), once in the ignorable combinators and once outside it. Removing this line yielded the same results in my inspection of my test queries, and the test suite is still passing. Was this an error or am I missing something?

maartenvanvliet:
I think you're right about the `comment` being obsolete. It's already matched in the ignored branch so doesn't serve any purpose later on. 

The spec https://spec.graphql.org/draft/#sec-Appendix-Grammar-Summary.Ignored-Tokens also lists comments as ignored tokens so that's correct.

stellanor:
The CI is failing with errors like:

> Requested Erlang/OTP version (22) not found in version list (should you be using option 'version-type': 'strict'?)

That doesn't seem related to my changes, is there something I need to do?

benwilson512:
@stellanor this has been fixed in main, if you merge that in it should resolve.

stellanor:
In light of the optimization work #1220, I updated this PR to use a default token count limit of `:infinity`. I'd still like for it to be possible to configure a limit, but the default behavior is no longer as dangerous with the optimization.

Parsing queries with many tokens remains somewhat performance-intensive. I'd like to be able to configure a reasonable token limit for my service. Simply enforcing a limit on the raw request size coming in to my web server isn't enough on its own; I need to be able to support large string literals or variables which do not seem to be a performance concern to the same degree in my testing.

mattbaker:
@maartenvanvliet @benwilson512 — Is there anything else needed on this one? We're running on a fork with this fix but it would be nice to get it into Absinthe proper so we could get off the fork :) 

Also @stellanor can correct me if I'm wrong but I think we've been running this for about a month now if that builds any confidence. We're fielding around 6500 requests a second and haven't seen any issues. 

@stellanor looks like a merge conflict might have popped up since last month, is that an easy fix?

stellanor:
@mattbaker Thanks for checking back in on this--I'm also eager to merge this if we can. I'll fix the merge conflict Monday. (I also added you and some teammates to my fork if anyone wants to dig in tomorrow.)

We have not deployed this code in production. We have been running a fork that counts tokens more simply and instruments the results. 

One other step required is to PR into absinthe-plug to plumb through this configuration option. I'll do that if I have time but it might be something to delegate to a teammate.

benwilson512:
Hi @stellanor apologies I lost the thread on this. I am happy to move forward with this PR, we just need to handle the minor merge conflict.

stellanor:
Updated to fix the conflict (and a test I forgot to tweak.) Thanks, @benwilson512 ! Once this is merged we can look at adding this option to absinthe_plug

benwilson512 (approved):
Excellent, thanks for your continued work here!

PR #1211, opened 2022-12-04
Pass resolution to dataloader helper's callback

### LINKED ISSUE #1192: [Feature suggestion] Pass context to the callback function of Absinthe.Resolution.Helpers.dataloader/2 options
Hello everyone!

According to the docs, the callback function of Absinthe.Resolution.Helpers.dataloader options
> Can be used to e.g. compute fields on the return value of the loader.

This is a very useful way of transforming the data.

I think it this function would benefit from being passed the context, in case the data transformation needed to e.g. rely on session state.

Like this:
```
resolve dataloader(source, callback: fn (result, _parent, _args, context) ->
  {:ok, transform_data(result, context)}
end)
```

What do you think?

antonsatin:
Resolves #1192 

Hi folks 👋🏻 

This PR adds support for getting resolution inside a callback function that's passed to `dataloader` helper. Basically it allows to pass `callback/4` instead of `callback/3` to `dataloader` options, in that case last argument will be resolution struct. 

The `do_dataloader` function becomes a bit thick with its arguments list, but using `opts` or changing it completely didn't seem like a good option to me either.

### DIFF
--- a/lib/absinthe/resolution/helpers.ex
+++ b/lib/absinthe/resolution/helpers.ex
@@ -132,7 +132,9 @@ defmodule Absinthe.Resolution.Helpers do
     @type dataloader_opt ::
             {:args, map}
             | {:use_parent, true | false}
-            | {:callback, (map(), map(), map() -> any())}
+            | {:callback,
+               (map(), map(), map() -> any())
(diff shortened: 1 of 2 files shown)


benwilson512 (approved):
This looks great, thank you!

PR #1213, opened 2022-12-06
allow changing result phase for subscriptions

MortadaAK:
This PR will respect the `result_phase` from the pipeline for subscription resolution. 

An example of using `Absinthe.Phoenix.Controller.Result` as the result phase will allow using atoms as keys and returning structs

### DIFF
--- a/test/absinthe/execution/subscription_test.exs
+++ b/test/absinthe/execution/subscription_test.exs
@@ -3,6 +3,48 @@ defmodule Absinthe.Execution.SubscriptionTest do
 
   import ExUnit.CaptureLog
 
+  defmodule ResultPase do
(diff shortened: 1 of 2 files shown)


test/absinthe/execution/subscription_test.exs
```diff
@@ -3,6 +3,48 @@ defmodule Absinthe.Execution.SubscriptionTest do
 
   import ExUnit.CaptureLog
 
+  defmodule ResultPase do
```
benwilson512, line 6: This looks like a typo I think. No worries, I'll fix after I merge.

benwilson512 (approved):
This is a good idea, thanks!

PR #1214, opened 2022-12-14
Allow passthrough of opts to pipeline modifiers

maartenvanvliet:
This can be used to e.g. pass through options to custom schema
phases in SDL rendering, to change the schema just before rendering it.


benwilson512 (approved):
Looks good, thanks!

Issue #925, opened 2020-05-19
Error "extensions" handling doesn't match spec


dmarkow:
The [GraphQL Spec](https://spec.graphql.org/June2018/#sec-Errors) provides information on how errors should be structured, which includes the `message`, `locations`, and `path` keys. The spec then states that anything additional should be nested under an `extensions` key rather than added at the top level, the reasoning being that future revisions to the spec may add additional keys that could conflict with any custom data.

Right now errors seem to be processed in this way:

- Resolve with `{:error, message: "bad request", something: "something"}`
- During resolution, the error is converted into a `%Phase.Error{}` struct with all of the keys other than `message` being pushed into a nested `extra` field: `{:error, message: "bad request", extra: %{something: "something"}}`
- `Absinthe.Phase.Document.Result.format_error/1` later merges the `extra` key back into the top level.
-  The error passed back to the client then looks something like `%{errors: [%{message: "bad request", something: "something"}]`. 

In the mean time I can get around this by having my resolvers explicitly nest my extra info under an `extensions` key.

I'm not sure if this is something that should be changed in Absinthe to match the spec, or if we just need to update the Errors documentation to state that extra info should be in an `extensions` key.

### Environment

* Elixir version (elixir -v): 1.10.3
* Absinthe version (mix deps | grep absinthe): 1.5.1
* Client Framework and version (Relay, Apollo, etc): Apollo Client 2.x

### Expected behavior

Based on the spec, I would expect the errors to be returned as:

```elixir
%{
  errors: [
    %{
      message: "bad request",
      extensions: %{
        something: "something"
      }
    }
  ]
}
```

### Actual behavior

The errors are returned as:

```elixir
%{
  errors: [
    %{
      message: "bad request",
      something: "something"
    }
  ]
}
```

benwilson512:
Hi @dmarkow good catch. This behaviour exists from before extensions were part of the GraphQL spec. I'm fine aiming to change this in 1.6, with the aim to create some kind of warning in a 1.5.x release.

mathieuprog:
Any plans on adding the extensions entry? It seems important to have error codes for the client to use:

```javascript
switch (err.extensions.code) {
  case 'UNAUTHENTICATED':
    // ...
```

For now we can just add those custom data in the errors map though, but it's just a pity not to be able to do it the standard way.

sukinoverse:
I would like to vote for this issue to have more priority to catch-up with GraphQL spec. 👍🏻 

[...]

PR #1215, opened 2022-12-20
Place extra errors in extensions field

maartenvanvliet:
As per the spec (https://spec.graphql.org/October2021/#sec-Errors.Error-result-format) any extra errors are now placed in the `extensions` field. 

This only happens when the `spec_compliant_errors` opt is set to true. As this is a breaking change and some clients may relay on the current behaviour it is opt in. With Absinthe 2.0 we can switch this around and make the spec compliant behaviour opt-out and the legacy behaviour opt-in. 

Fixes https://github.com/absinthe-graphql/absinthe/issues/925


benwilson512 (approved):
Great addition, thanks!

PR #1218, opened 2022-12-23
Allow configurable docset runner for subscription documents

maartenvanvliet:
As the docs mention, currently when subscription documents are run
it provides a simple form of back pressure due to it blocking the mutation.
Furthermore there is no batching across subscriptions.

This PR gives the ability to configure a docset runner by implementing
the Absinthe.Subscription.DocSetRunner behaviour.

A custom runner could be implemented to provide batching,
or to provide a different form of back pressure. Both of these
are useful for scaling out subscriptions. They can also both be highly
specific to the project and so it makes sense to allow them to be configured.
Furthermore, it also gives the opportunity to adjust the pipeline
for subscription documents just before execution.

I've changed the docset runner to accept a struct instead of a tuple,
as it makes it easier to add new options in the future.

The default is the same as before, a simple runner that runs the docset
and blocks the mutation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Feature: [Configurable docset runner for subscription documents](https://github.com/absinthe-graphql/absinthe/pull/1218)
 - Bug Fix: [Validate field identifier uniqueness](https://github.com/absinthe-graphql/absinthe/pull/1200)
 - Bug Fix: [Validate type references for invalid wrapped types](https://github.com/absinthe-graphql/absinthe/pull/1195)
 - Breaking Bugfix: [Validate repeatable directives on schemas](https://github.com/absinthe-graphql/absinthe/pull/1179)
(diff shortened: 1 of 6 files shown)


michaelcaterisano:
@maartenvanvliet @benwilson512 I recently forked Absinthe because I needed to customize the `run_docset` function. It seems this PR would allow me do that without forking. Were there any objections to the direction of this PR? I'd be happy to help making any additional changes necessary to get something like this merged.

cschiewek:
👍

maartenvanvliet:
Glad you find it useful. I no longer have the need for this feature, so feel free to work on it. 

One thing I was thinking a while back is that instead of a behaviour and all that entails, it could be a simple callback function passed into `:docset_runner` in the context. It would default to the current behaviour. This because adding a behaviour for just one function may be a bit much.

benwilson512:
I think I'd be more open to a function callback / option based approach. I think document management and execution IS something we'll eventually want a behavior on, but it's a pretty big topic overall.

michaelcaterisano:
@benwilson512 Sounds good, my team and I are happy to work on a callback approach. We'll have a PR up soon.

maartenvanvliet:
Gonna close this in favour of https://github.com/absinthe-graphql/absinthe/pull/1295

PR #1220, opened 2023-01-06
Performance-optimize lexer column mapping (preventing DoS with maliciously long queries)

stellanor:
Hi! I’m back with another PR on the theme of lexer performance and protecting against Denial-of-Service attacks.

If you recall from my PR https://github.com/absinthe-graphql/absinthe/pull/1210 in December, I discovered that the byte-to-character column mapping routine in `Absinthe.Lexer` has quadratic time complexity and unacceptable performance on large inputs. This puts services at risk of DoS. In my first PR I introduced a configurable token count limit. I discovered after opening that PR that my service cannot mitigate this DoS risk with limits alone, because we have to support some legitimate queries that far exceed any safe limit (and are performing slowly today.)

So, I took a stab at optimizing this routine and eliminating the performance bottleneck. Results are looking good. Here are some benchmark results, testing `Absinthe.Lexer.tokenize("{ __typename #{String.duplicate("@abc ", n)} }”)` for different values of n:

Screen Shot 2023-01-06 at 10 29 31 AM

The original implementation is slow because it maps through each token, looks up the line the token belongs to, and counts the characters from the beginning of the line to the token's first byte. In the case of a long line with many tokens, this is a lot of redundant counting and O(n^2) time complexity.

My optimized implementation removes lines and/or characters after they have been processed and counted. This makes fetching lines by index faster, and most importantly eliminates redundant work counting characters before a token. Of course this algorithm is more complex and less concise--we must pass cursor and offset data through each iteration to represent what has already been counted.

I'd like to thank and acknowledge my colleague @zoevkay who rigorously reviewed my solution (in a private repo). Thanks to her collaboration with me, the end result is both faster and more clear than my first draft.


benwilson512:
This is a wonderful contribution, thank you so much! Reviewing now.

lib/absinthe/lexer.ex
```diff
@@ -242,12 +242,88 @@ defmodule Absinthe.Lexer do
...
+  defp convert_token_columns_from_byte_to_char(tokens, [first_line | next_lines]) do
+    initial_cursor_state = %{
+      line_num_cursor: 1,
+      current_line_substring: first_line,
+      current_line_char_offset: 1,
+      current_line_byte_offset: 1,
+      next_lines: next_lines
+    }
+
+    Enum.reduce(tokens, {[], initial_cursor_state}, fn current_token, {results, cursor_state} ->
```
maartenvanvliet, line 254: Wow, really nice improvement! 

One tiny nitpick is to use Enum.map_reduce here. Cuts down on the boilerplate of Enum.reduce a little.

stellanor: Oh! Great suggestion, thanks @maartenvanvliet ! Today I learned about `Enum.map_reduce`...

 I'll update the PR.

stellanor:
Updated with @maartenvanvliet 's suggestion to use `Enum.map_reduce`. Thanks again for the feedback!

stellanor:
Hi @benwilson512 , do you have time in the next couple of weeks to review this?

Just an update on my end: we've been running this optimization in my service's production environment for about a week and it's been going great. A nice side effect of this is that we saw some dramatic response time performance gains with some very large requests. Outliers to be sure, but it's nice to see that this did more for our service than just reduce risk.

benwilson512 (approved):
This is a wonderful contribution, thanks for following up!

PR #1221, opened 2023-01-12
Add pipeline modifier option to Absinthe.run

bernardd:
One of the limitations of testing with `Absinthe.run/3` is the inability to add custom pipeline phases. Doing so requires going through the full HTTP stack to test. That means that if there's a custom phase that is critical to your system's operation, it immediately renders `Absinthe.run/3` tests unusable.

This change allows the specification of a pipeline modifier callback function which takes the default pipeline and options and returns the pipeline to be used in the tests.


benwilson512 (approved):
Great idea, thanks!

Issue #1071, opened 2021-05-14
non_null(list_of(non_null(...))) doesn't propagate list of nil upwards


atomkirk:
If submitting a bug, please provide the following:

### Environment

* Elixir version (elixir -v): 1.10.4
* Absinthe version (mix deps | grep absinthe): 1.6.3
* Client Framework and version (Relay, Apollo, etc): elm-graphql

### Expected behavior

```
    field :users_who_can_sign, non_null(list_of(non_null(:user))) do
      resolve(fn %{id: field_id}, _, _ ->
        # …this returned [nil, nil, nil]
      end)
    end
```
It should have propogated `null` up to the next highest nullable field

### Actual behavior

the resulting payload was:

```
usersWhoCanSign: null
```

So the spec says, as I'm sure you know:

>If a List type wraps a Non-Null type, and one of the elements of that list resolves to null, then the entire list must resolve to null. If the List type is also wrapped in a Non-Null, the field error continues to propagate upwards.

So, it works correctly if it the field wasn't a non_null, but since its non_null, it should have propagated upward

benwilson512:
Thanks! So I've sort of reproduced this in https://github.com/absinthe-graphql/absinthe/pull/1072, but if you look at the test failure the `data` field is actually doing the right thing already today. It's missing the error message, which is bad, but I'm not reproducing the behavior you're seeing.

Can you create a reproduceable example?

ksmithut:
Here is a minimal example where I was able to reproduce the issue:

<details>
<summary>`server.ex`</summary>

```elixir
Application.put_env(:sample, Sample.Endpoint,
  http: [ip: {127, 0, 0, 1}, port: 4000],
  server: true
)

Mix.install([
  {:plug_cowboy, "2.6.1"},
  {:jason, "1.4.0"},
  {:phoenix, "1.7.2"},
  {:absinthe, "1.7.1"},
  {:absinthe_plug, "1.5.8"},
  {:req, "0.3.6"}
])

defmodule Sample.Schema do
  use Absinthe.Schema

  object :todo do
    field(:id, non_null(:id))
  end

  object :wrapper do
    field :todos, non_null(list_of(:todo)) do
      resolve(fn _, _ ->
        async(fn ->
          {:error, "error"}
        end)
      end)
    end
  end

  query do
    field :wrapper, :wrapper do
      resolve(fn _, _ -> {:ok, %{}} end)
    end
  end
end

defmodule Sample.Endpoint do
[...]

PR #1223, opened 2023-01-24
Propagate errors if a non-null list of non-null elements contains a null

jueberschlag:
If a non-null list of non-null elements contained a null element, the null value was correctly propagated to the parents but the error wasn't.
This is described in the issue https://github.com/absinthe-graphql/absinthe/issues/1071 and also in the WIP PR https://github.com/absinthe-graphql/absinthe/pull/1072

I am not sure it is the right way to fix this. According to the comment in `do_propagate_null_trimming/0`, the null element resolution should have failed earlier.


benwilson512 (approved):
Great catch, thanks! The message you removed that said it should have been handled is wrong I think.

Issue #1225, opened 2023-02-03
extend schema cannot parse maps in arguments


kzlsakal:
### Environment

* Elixir version (elixir -v): `Elixir 1.14.0 (compiled with Erlang/OTP 25)`
* Absinthe version (mix deps | grep absinthe): `1.7.1`
* Client Framework and version (Relay, Apollo, etc): `n/a`

### Expected behavior

Extending the schema with the link directive and its arguments should result the following SDL output:

```graphql
schema @link(url: "https:\/\/specs.apollo.dev\/federation\/v2.0", import: ["@key", "@tag", {name: "@override", as: "@replace"}]) {
  query: RootQueryType
}

type RootQueryType {
  hello: String
}
```

### Actual behavior

The map is eventually supplied to `Absinthe.Blueprint.Input.parse/1` but in AST format, therefore we get the following compile-time error:

```
(FunctionClauseError) no function clause matching in Absinthe.Blueprint.Input.parse/1    
    
    The following arguments were given to Absinthe.Blueprint.Input.parse/1:
    
        # 1
        {:quote, [closing: [line: 37, column: 36], line: 37, column: 10], [{:%{}, [closing: [line: 37, column: 35], line: 37, column: 17], [name: "@override"]}]}
    
    Attempted function clauses (showing 8 out of 8):
    
        def parse(-%struct{} = value-) when -struct === Absinthe.Blueprint.Input.Boolean- or -struct === Absinthe.Blueprint.Input.Enum- or -struct === Absinthe.Blueprint.Input.Field- or -struct === Absinthe.Blueprint.Input.Float- or -struct === Absinthe.Blueprint.Input.Integer- or -struct === Absinthe.Blueprint.Input.List- or -struct === Absinthe.Blueprint.Input.Object- or -struct === Absinthe.Blueprint.Input.String- or -struct === Absinthe.Blueprint.Input.Null-
        def parse(+value+) when -is_integer(value)-
        def parse(+value+) when -is_float(value)-
        def parse(+value+) when -value == nil-
        def parse(+value+) when -is_binary(value)-
        def parse(+value+) when -is_boolean(value)-
        def parse(+value+) when -is_list(value)-
        def parse(+value+) when -is_map(value)-

Stacktrace:
  │ (absinthe 1.7.1) lib/absinthe/blueprint/input.ex:37: Absinthe.Blueprint.Input.parse/1
[...]

Issue #1024, opened 2021-01-03
Meta attributes are macro quoted?


mmmries:
### Environment

* Elixir version: Elixir 1.11.2 (compiled with Erlang/OTP 23)
* Absinthe version): 1.5.5
* Client Framework and version (Relay, Apollo, etc): None

### Expected behavior

I can use maps, structs or other native data types in the `meta` attributes on a field

```elixir
field :paddles, list_of(:paddle) do
  meta(:proxy_to, %{url: "http://localhost:9002/graphql"})
  ..
end

# Later in a resolver

def resolve(_source, _args, resolution) do
  resolution.definition.schema_node.private.meta.proxy_to # => %{url: "http://localhost:9002/graphql"}
end
```

### Actual behavior

Only simple scalar types seem to work for meta attributes. From the example above when I end up getting in my resolver is

```elixir
def resolve(_source, _args, resolution) do
  resolution.definition.schema_node.private.meta.proxy_to # => {:%{}, [line: 8], [url: "http://localhost:9001/api/graphql"]}
end
```

I think this is because the meta attribute has been `quote`d without being `unquote`d.

### Relevant Schema/Middleware Code

I'm trying to pick up the torch of schema stitching and move it forward for the community.
After reading through [some issues](https://github.com/absinthe-graphql/absinthe/issues/387#issuecomment-739620900) it seemed like there was a desire for a community project to add this type of functionality.
I learned what I could from [open source examples](https://github.com/newrelic/absinthe-schema-stitching-example) and started it on trying to build something that was general enough to be re-usable by other teams https://github.com/SpiffInc/absinthe_compose.
I'm interested in tackling this problem because our team at Spiff is wanting to build individual services and provide a unified graphql interface. I would like to keep the unified interface as thin as possible and schema stitching seemed like a great approach.

As I played around with how to approach the problem I liked the idea of being able to mark certain parts of your schema with metadata and then have resolvers automatically create the downstream queries and pull the data back into the local schema. I'm hoping to support arbitrary extensibility by letting people do things like `meta(:proxy_to, {MyClient, [token: some_secret]})` where they can specify a module and pass some "init_args" style configuration. This way they could be calling graphql over some other transport, or using a custom HTTP client etc.

https://github.com/absinthe-graphql/absinthe/issues/928 seems to have a similar request, but it sounded like there were reasons that the project wanted to keep support for a more limited set of data types that could be passed in the `meta` attributes? I'm happy to change direction on this.

[...]

PR #1226, opened 2023-02-03
Allow maps as arguments in macros

kzlsakal:
This is one approach to possibly solve the issue https://github.com/absinthe-graphql/absinthe/issues/1225

My main goal is to support passing a map in the arguments for a directive that is wrapped in an `extend` macro.

### DIFF
--- a/lib/absinthe/blueprint/input.ex
+++ b/lib/absinthe/blueprint/input.ex
@@ -69,6 +69,11 @@ defmodule Absinthe.Blueprint.Input do
     }
   end
 
+  # Handle maps in AST format if they are not unquoted in macros
+  def parse({:%{}, _, map_key_values}) when is_list(map_key_values) do
+    map_key_values |> Enum.into(%{}) |> parse()
+  end
+
   def parse(value) when is_map(value) do
     %Input.Object{
       fields:
--- a/lib/absinthe/blueprint/input.ex
+++ b/lib/absinthe/blueprint/input.ex
@@ -69,11 +69,6 @@ defmodule Absinthe.Blueprint.Input do
     }
   end
 
-  # Handle maps in AST format if they are not unquoted in macros
-  def parse({:%{}, _, map_key_values}) when is_list(map_key_values) do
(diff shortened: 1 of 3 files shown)


maartenvanvliet (commented):
Good catch! 

I think the best place to convert the AST to maps would be in the Schema.Notation module. This to prevent the AST from leaking into other parts of Absinthe.

We can limit support to literal maps only for now. This would most closely match the SDL notation.

Note that this issue also happens elsewhere: https://github.com/absinthe-graphql/absinthe/issues/1024

kzlsakal:
Thanks for the feedback. I moved the logic to Notation as you both recommended.

benwilson512 (approved):
Looks great, thanks!

hauleth:
When this will be released?

hauleth:
This cause `Inspect.Error` to be written in generated SDL file while command happily succeeds.

benwilson512:
Hey @hauleth please file an issue with the relevant info on this.

hauleth:
I found more info - it works, but when the keys are binaries, not when keys are atoms. I will try to post more info when I will find time to spare.

maartenvanvliet:
```
diff --git a/lib/absinthe/schema/notation/sdl_render.ex b/lib/absinthe/schema/notation/sdl_render.ex
index baac5a24..35ea517d 100644
--- a/lib/absinthe/schema/notation/sdl_render.ex
+++ b/lib/absinthe/schema/notation/sdl_render.ex
@@ -408,7 +408,7 @@ defmodule Absinthe.Schema.Notation.SDL.Render do
   end
 
   defp render_value(%Blueprint.Input.Field{name: name, input_value: value}),
-    do: concat([name, ": ", render_value(value)])
+    do: concat([to_string(name), ": ", render_value(value)])
 
   defp render_value(%{value: value}),
     do: to_string(value)
```
The `Inspect.concat/1` expects binaries so it fails here when it encounters an atom. Above patch fixes it but perhaps it's better to cast to a string or fail in the `expand_ast` call as pretty much anything can be the key of a map.

PR #1227, opened 2023-02-08
Support propagating OpenTelemetry context

rewritten:
This change adds optional support for propagating `OpenTelemetry` contexts in tasks spawned by the `Async` and `Batch` middleware.

This allows to correlate spans in otel backends (like datadog and others), so something like

```elixir
field :posts, :list_of(:post) do
  resolve fn user, _, _ -> async(fn -> PostsResolver.posts_by_user(user) end) end
end

defmodule PostsResolver do
  def posts_by_user(user) do
    OpenTelemetry.Tracer.with_span("data access: posts by user") do
      # query posts
    end
  end
end
```

will create a child span in the existing trace, instead of creating a totally unrelated span.


kdawgwilk:
This is fantastic! 👏

kdawgwilk:
Does https://github.com/spandex-project/spandex have a similar Task you can delegate to?

rewritten:
@kdawgwilk Do you need anything else to make this accepted?

benwilson512 (approved):
Happy to merge this, thanks!

PR #1231, opened 2023-02-17
Validate input object not being an Enum

rewritten:
We have found out that in a query like this, when the `complexArg` argument is an input object:

```graphql
{ someField(complexArg: SEEMS_AN_ENUM) }
```

the argument is discarded instead of being marked as an error like if it was a scalar.

This PR adds a parse check for `Input.Enum` for this case, and tests for this and other cases.


benwilson512:
Oops, great catch!

rewritten:
Thanks to @ssaenz and @JamieMaclean for spotting the issue.

benwilson512 (approved):

PR #1235, opened 2023-03-22
Fix endpoint and docket setup documentation for subscriptions

JoseMPena:
The current documentation on Subscriptions indicates we should `use Absinthe.Phoenix.Socket` module in our `Endpoint` which raises an undefined function error, instead of using `Absinthe.Phoenix.Endpoint` in our endpoint module and then using the `Absinthe.Phoenix.Socket` on our socket modules.


benwilson512 (approved):
Thanks

PR #1236, opened 2023-03-28
add previous walk_results/6 as catch all for resolution without path

ucwaldo:
related question in slack: https://elixir-lang.slack.com/archives/C0PR49P4P/p1679996159340499

We have a custom pipeline with which we skip some validation phases to not break external clients when upgrading absinthe. We basically replace the validation result phase with a custom one and keep track of which validations we explicitly skip so we get some insights and can address them separately:

```
config
|> Absinthe.Plug.default_pipeline(pipeline_opts)
|> Absinthe.Pipeline.replace(
  Absinthe.Phase.Document.Validation.Result,
  {CustomValidationResult, phases: [Absinthe.Phase.Document.Validation.NoUnusedFragments]}
)
```

Since upgrading from 1.6 to 1.7 there is a specific scenario where this stopped working when using transport level batching and the @skip directive.

```
query MyQuery1 {
  foo @skip(if: true) {
    bar
  }
}
query MyQuery2 {
  biz {
    baz
  }
}

fragment Unused on SomeType {
  foo
  bar
}
```

This explodes with the following error:

```
Attempted function clauses (showing 2 out of 2):

defp walk_results([value | values], bp_node, inner_type, res, [i | sub_path] = path, acc)
defp walk_results([], _, _, res = %{path: [_ | sub_path]}, _, acc)

code: |> post(@gql_path, %{
stacktrace:
  (absinthe 1.7.1) lib/absinthe/phase/document/execution/resolution.ex:113: Absinthe.Phase.Document.Execution.Resolution.walk_results/6
  (absinthe 1.7.1) lib/absinthe/phase/document/execution/resolution.ex:93: Absinthe.Phase.Document.Execution.Resolution.walk_result/5
  (absinthe 1.7.1) lib/absinthe/phase/document/execution/resolution.ex:67: Absinthe.Phase.Document.Execution.Resolution.perform_resolution/3
  (absinthe 1.7.1) lib/absinthe/phase/document/execution/resolution.ex:24: Absinthe.Phase.Document.Execution.Resolution.resolve_current/3
  (absinthe 1.7.1) lib/absinthe/pipeline.ex:405: Absinthe.Pipeline.run_phase/3
  (absinthe 1.7.1) lib/absinthe/pipeline/batch_resolver.ex:87: Absinthe.Pipeline.BatchResolver.run_pipeline/3
  (absinthe 1.7.1) lib/absinthe/pipeline/batch_resolver.ex:74: Absinthe.Pipeline.BatchResolver.execute/5
  (absinthe 1.7.1) lib/absinthe/pipeline/batch_resolver.ex:46: Absinthe.Pipeline.BatchResolver.do_resolve/6
  (absinthe_plug 1.5.8) lib/absinthe/plug/batch/runner.ex:44: Absinthe.Plug.Batch.Runner.build_valid_results/2
  (absinthe_plug 1.5.8) lib/absinthe/plug/batch/runner.ex:17: Absinthe.Plug.Batch.Runner.run/4
  (absinthe_plug 1.5.8) lib/absinthe/plug.ex:510: Absinthe.Plug.run_request/4
```

Removing the skip directive or the validation skip resolves the error, but doesn’t result in the desired behaviour.

What doesn’t match is 

```
defp walk_results([], _, _, res = %{path: [_ | sub_path]}, _, acc)
```

because `res.path` is `nil`. It appears that this was ok in absinthe 1.6 but not in absinthe 1.7 anymore: https://diff.hex.pm/diff/absinthe/1.6.8..1.7.1#14238965-118-


benwilson512:
I think adding the clause makes sense but can you post a complete example I can run to test something? doesn't have to be an actual test case, just a complete example I can run.

ucwaldo:
Thanks for taking looking into this! I am not having the easiest time to get a real reproduction going as i am unsure how i need to formulate the queries so they actually go through the correct case of `walk_result` in the resolution execution phase. I assume that i need to formulate bigger/nested queries, or queries that retrieve the same results, as the comment above `walk_results` states:

> This function walks through any existing results. If no results are found at a given node, it will call the requisite function to expand and build those results

This should still give you a good idea of what we're doing in our app: https://github.com/ucwaldo/absinthe/pull/2/files
i am just not sure how i can make sure we go through the correct path and land here instead of the case above:
https://github.com/absinthe-graphql/absinthe/blob/master/lib/absinthe/phase/document/execution/resolution.ex#L92

ucwaldo:
Additionally, i'm not sure how i can test batched queries in absinthe, maybe absinthe_plug would be a better place to get a reproduction going? It looks similar to what i have here, just with actual batch queries. However, i have the same issue with not getting into the right case of `walk_result` in the document execution phase.

ucwaldo:
@benwilson512 i managed to reproduce the error in an `absinthe_plug` test: https://github.com/ucwaldo/absinthe_plug/pull/1/files

note that when you remove the `@skip` directive, the query works, but while it's there it raises the error from this PR description.

Sorry for jumping between repos, but since we encounter the error only during transport level batch requests, it was easier for me to reproduce it there.

benwilson512:
No worries about the other repo, that's super helpful thank you! I'll give it a look and see if I can dig into why this happens in more detail.

As I said I'm not opposed to the patch here, I just want to understand the circumstances where it is necessary better.

ucwaldo:
👋 @BenWilson really don’t want to push, have you had a chance to look into this? I was wondering if it would be possible to fix in the batch runner of absinthe_plug instead of going with this patch and would be interested in your opinion

aaronrenner:
Sorry to pester, but I've been seeing this issue as well since I upgraded to absinthe 1.7. I know it's a bit of an edge case, but I'm curious if there's been any further developments on resolving these errors.

benwilson512:
Hey @aaronrenner apologies on the delay.

I have not had time to look into it further, but given that the tests pass, that it can't possibly interfere with other clauses, and the bit where this fixes issues people are having I'm gonna just go ahead and merge.

benwilson512 (approved):

aaronrenner:
Thank you! ❤️

PR #1238, opened 2023-03-29
Fix Absinthe.Subscription parameter in Subscriptions guides

mbuffa:
Hello there,

There's a small inconsistency in the Subscriptions guide. It may have been valid for older `:absinthe_phoenix` versions but the latest version documentation states otherwise:
https://hexdocs.pm/absinthe_phoenix/2.0.2/readme.html


benwilson512 (approved):
Good catch, thanks!

PR #1239, opened 2023-04-05
Fix dynamic enums when used as arguments with default value

rewritten:
When an enum is defined dynamically, its `values` field will contain a list-of-one-list-of-atoms (see #1041) and the validator will fail.

This change allows the validator to be used for dynamic enums too.

(tests pending)


benwilson512 (changes requested):
Hey @rewritten I think the issue here is that the way this ought to be handled is that we ought to be normalizing the values of the enum so that they are the same shape no matter how they are in the schema.

If all of the code has to worry about it being both a list of atoms AND a list of structs it's going to get a lot more complicated.

Issue #1233, opened 2023-03-14
Schema compilation with absinthe 1.7.1 generates warnings with elixir version < 1.14


MaeIsBad:
### Environment

* Elixir version (elixir -v): 1.11.2
* Absinthe version (mix deps | grep absinthe): 1.7.1 (absinthe) c0c4dbd9
This issue doesn't occur with absinthe 1.7.0
* Client Framework and version (Relay, Apollo, etc): n/a

### Expected behavior

Absinthe doesn't generate warnings during schema compilation

### Actual behavior

During compilation of a schema 4 warnings are emitted

```
root@31064b1ea5b7:/code# mix compile --warnings-as-errors
Compiling 1 file (.ex)
warning: this clause cannot match because a previous clause at line 74 always matches
  deps/absinthe/lib/absinthe/phase/schema/compile.ex:74

warning: this clause cannot match because a previous clause at line 74 always matches
  deps/absinthe/lib/absinthe/phase/schema/compile.ex:74

warning: this clause cannot match because a previous clause at line 74 always matches
  deps/absinthe/lib/absinthe/phase/schema/compile.ex:74

warning: this clause cannot match because a previous clause at line 74 always matches
  deps/absinthe/lib/absinthe/phase/schema/compile.ex:74
```

since the warnings occur during code generation this will cause `mix compile --warnings-as-errors` to fail, which causes problems for CI environments

### Relevant Schema/Middleware Code

```elixir
if not Elixir.Version.match?(System.version(), ">= 1.11.0 and < 1.14.0") do 
  IO.warn("Elixir version 1.14 or newer will not emit a warning")
end

defmodule MyAppWeb.Schema do
  use Absinthe.Schema

  # Example data
  @items %{
    "foo" => %{id: "foo", name: "Foo"},
    "bar" => %{id: "bar", name: "Bar"}
  }

  object :item do
    field(:id, :id)
    field(:name, :string)
  end

  query do
    field :item, :item do
      arg(:id, non_null(:id))

      resolve(fn %{id: item_id}, _ ->
        {:ok, @items[item_id]}
      end)
    end
  end
end
```

MaeIsBad:
After git bisecting it seems like 7eec586ca47a17da265141ad44aad7d98cf3a355 is the first bad commit

MaeIsBad:
After bisecting 5f1b05b2878a1075b0b2ee875b3542df58e7c8f3 is the first bad commit

[...]

PR #1242, opened 2023-04-13
Deduplicate directives when building schema

MaeIsBad:
When importing directives from the prototype some of them were duplicated, which caused multiple functions with the same clause to be generated.
This should (partially?) fix https://github.com/absinthe-graphql/absinthe/issues/1233

Thanks @maartenvanvliet for looking into the original issue :blue_heart:

### DIFF
--- a/lib/absinthe/phase/schema/build.ex
+++ b/lib/absinthe/phase/schema/build.ex
@@ -5,12 +5,15 @@ defmodule Absinthe.Phase.Schema.Build do
     %{schema_definitions: [schema]} = blueprint
 
     types = build_types(blueprint)
-    directive_artifacts = build_directives(blueprint)
+
+    directive_artifacts =
+      (schema.directive_artifacts ++ build_directives(blueprint))
+      |> Enum.uniq_by(fn v -> v.identifier end)
 
     schema = %{
       schema
       | type_artifacts: types,
-        directive_artifacts: schema.directive_artifacts ++ directive_artifacts
+        directive_artifacts: directive_artifacts
     }
 
     blueprint = %{blueprint | schema_definitions: [schema]}
--- a/lib/absinthe/phase/schema/build.ex
+++ b/lib/absinthe/phase/schema/build.ex
@@ -5,15 +5,12 @@ defmodule Absinthe.Phase.Schema.Build do
     %{schema_definitions: [schema]} = blueprint
 
     types = build_types(blueprint)
-
-    directive_artifacts =
-      (schema.directive_artifacts ++ build_directives(blueprint))
-      |> Enum.uniq_by(fn v -> v.identifier end)
+    directive_artifacts = build_directives(blueprint)
 
     schema = %{
       schema
       | type_artifacts: types,
-        directive_artifacts: directive_artifacts
+        directive_artifacts: schema.directive_artifacts ++ directive_artifacts
     }
 
     blueprint = %{blueprint | schema_definitions: [schema]}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Bug Fix: [Object type extensions may be empty](https://github.com/absinthe-graphql/absinthe/pull/1228)
 - Bug Fix: [Validate input object not being an Enum](https://github.com/absinthe-graphql/absinthe/pull/1231)
+- Bug Fix: [Deduplicate directives when building schema](https://github.com/absinthe-graphql/absinthe/pull/1242)
 
 ## 1.7.1
 - Breaking Bugfix: [Validate repeatable directives on schemas](https://github.com/absinthe-graphql/absinthe/pull/1179)
--- a/lib/absinthe/schema/prototype/notation.ex
+++ b/lib/absinthe/schema/prototype/notation.ex
@@ -38,6 +38,7 @@ defmodule Absinthe.Schema.Prototype.Notation do
         pipeline
         |> Absinthe.Pipeline.without(Absinthe.Phase.Schema.Validation.QueryTypeMustBeObject)
         |> Absinthe.Pipeline.without(Absinthe.Phase.Schema.ImportPrototypeDirectives)
+        |> Absinthe.Pipeline.without(Absinthe.Phase.Schema.DirectiveImports)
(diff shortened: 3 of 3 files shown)


lib/absinthe/phase/schema/build.ex
```diff
@@ -5,12 +5,15 @@ defmodule Absinthe.Phase.Schema.Build do
     %{schema_definitions: [schema]} = blueprint
 
     types = build_types(blueprint)
-    directive_artifacts = build_directives(blueprint)
+
+    directive_artifacts =
+      (schema.directive_artifacts ++ build_directives(blueprint))
+      |> Enum.uniq_by(fn v -> v.identifier end)
```
benwilson512, line 11: Unless I am missing something, shouldn't directives with duplicate identifiers be flagged with errors just like objects with duplicate identifiers? Silently removing them doesn't seem good either because there is no guarantee that the directives are logically identical.

benwilson512 (changes requested):
Hey thanks for taking a look at this! I have some concerns about solving it this way that I have noted.

lib/absinthe/phase/schema/build.ex
MaeIsBad: if duplicate directives cause an error, currently it will cause the code generation to always fail, since absinthe imports duplicate directives from the prototype schema.

I'm not a fan for a few reasons but would using `{v.expand, v.identifier}` instead work for you?

MaeIsBad: Also note that this doesn't change the current behavior.
Absinthe already will overwrite directives if build is called with duplicates, all this does is remove a confusing elixir warning that the users can't really deal with themselves

maartenvanvliet: There's a check for duplicate directive identifiers, only this happens before the prototype directives are imported into the schema. 

I think alternatively we can prevent the issue from happening by not importing the default directives into the prototype schema. They are of no use there as include/skip only work at the runtime level of a document's execution.

```
--- a/lib/absinthe/schema/prototype/notation.ex
+++ b/lib/absinthe/schema/prototype/notation.ex
@@ -42,6 +42,10 @@ defmodule Absinthe.Schema.Prototype.Notation do
           Absinthe.Phase.Schema.TypeExtensionImports,
           {Absinthe.Phase.Schema.TypeExtensionImports, []}
         )
+        |> Absinthe.Pipeline.replace(
+          Absinthe.Phase.Schema.DirectiveImports,
+          {Absinthe.Phase.Schema.DirectiveImports, []}
+        )
       end
```

maartenvanvliet: yes, that'd be simpler 😅

MaeIsBad: In that case I'll close this PR. Again thanks for the help!

MaeIsBad: @maartenvanvliet you mentioned [here](https://github.com/absinthe-graphql/absinthe/issues/1233#issuecomment-1506714252) that

> Whilst investigating, I found a similar bug happens with the extend notation, this can also cause multiple functions with the same function heads to be generated

is this the same issue?

maartenvanvliet: It's related, the Schema.Notation module does a pass over the blueprint to lift the functions defined in the schema to the schema module. 
When it walks over the blueprint with a custom scalar, that is later extended using `extend` if will generate the same lifted function for both of them. 

E.g. 
```
  def __absinthe_function__(
      {Absinthe.Blueprint.Schema.ScalarTypeDefinition, :my_custom_scalar},
      :serialize
  ) do
    nil
  end
``` 
will be generated twice.

What I am as of yet unsure of is whether the first or second generated function can simply be removed. In the above case the second can be removed as `serialize` cannot be overridden by the extended scalar. I don't know whether there are cases currently where the second cannot be removed.

Regardless, supporting the overriding of `serialize` and similar in the `extend` syntax is something I've been thinking about for a while. Though it is beyond the graphql spec it would make it easier to override a custom scalar or even a resolver. A use case I have is overriding the introspection resolvers with my own. 

I'll have a look at the `directive` issue you're having

maartenvanvliet: > The problem with this solution is(I think) that on every "level" of schema inheritance a new `:deprecated` directive is defined
> 
> https://github.com/absinthe-graphql/absinthe/blob/5d4ef4c59b5adabd7be21890f13f0fccb42ae368/lib/absinthe/schema/prototype/notation.ex#L14
> 
> which still causes more of the same errors to appear

I cannot reproduce this. Do you have a repo showing this issue? Ideally with the OTP/elixir version.

MaeIsBad: hm, I am unable to reproduce this today as well, no idea why I thought that was what was happening yesterday.

I found that using the `schema do` notation causes more errors to be generated

```
defmodule MyAppWeb.Schema do
  use Absinthe.Schema

  schema do
   field :query, :query
   field :subscription, :subscription
  end

  subscription do
   field :name, :string
  end

  query do
    field :item, :string
  end
end
```

this doesn't happen if I remove the subscription

```
Compiling 1 file (.ex)
warning: this clause cannot match because a previous clause at line 1 always matches
  lib/asdf.ex:1

warning: this clause cannot match because a previous clause at line 1 always matches
  lib/asdf.ex:1

warning: this clause cannot match because a previous clause at line 1 always matches
  lib/asdf.ex:1

warning: this clause cannot match because a previous clause at line 1 always matches
  lib/asdf.ex:1

Compilation failed due to warnings while using the --warnings-as-errors option
```

maartenvanvliet: yes, the schema notation has the same issues as the extend notation. The fix should be similar as well.

benwilson512:
@maartenvanvliet do you have any concerns with not importing the default directives into prototypes?

maartenvanvliet:
> @maartenvanvliet do you have any concerns with not importing the default directives into prototypes?

No, that would work.

benwilson512 (approved):

benwilson512:
Thanks for handling this @MaeIsBad !

PR #1244, opened 2023-05-17
Do not unquote AST twice

josevalim:
Instead of generating

    def __absinthe_type__(:foo_bar), do: unquote(ast)
    def __absinthe_type__("FooBar"), do: unquote(ast)

We generate:

    def __absinthe_type__(:foo_bar), do: unquote(ast)
    def __absinthe_type__("FooBar"), do: __absinthe_type__(:foo_bar)

This considerably reduces the size of the generated code.
Compilation times went from 1.8s to 1.2s in a sample schema.

### DIFF
--- a/lib/absinthe/phase/schema/compile.ex
+++ b/lib/absinthe/phase/schema/compile.ex
@@ -103,7 +103,7 @@ defmodule Absinthe.Phase.Schema.Compile do
         end
 
         def __absinthe_type__(unquote(type.name)) do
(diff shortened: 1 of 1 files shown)


josevalim:
Btw, it may be more efficient at runtime to first generate all atom clauses and then all string clauses. But it may be the VM optimizes it anyway.

benwilson512 (approved):
Great call, thanks!

PR #1246, opened 2023-06-07
Modifying the example code of a resolve function

hmy3743:
The resolver functions of the example code are written to receive the id parameter, but arg is not declared in schema, so I add arg declaration.


benwilson512 (approved):
Great catch, thanks!

PR #1248, opened 2023-06-16
Add dialyxir to CI

paulo-ferraz-oliveira:
This:
* adds `dialyxir` to CI (cached PLT)
* tweaks some code to get rid of compilation warnings
* solves all existing Dialyzer -issued warnings

I may add comments to my own changes to clarify intention or source, if applicable (though the individual commits should tell part of the story already - they will have comments).


.dialyzer_ignore.exs
```diff
@@ -1,4 +1,8 @@
 [
   {":0:unknown_function Function :persistent_term.get/1 does not exist."},
-  {":0:unknown_function Function :persistent_term.put/2 does not exist."}
+  {":0:unknown_function Function :persistent_term.put/2 does not exist."},
+  {"lib/absinthe/middleware/async.ex", :unknown_function, 117},
+  {"lib/absinthe/middleware/batch.ex", :unknown_function, 213},
+  {"lib/absinthe/utils/render.ex", :improper_list_constr, 51},
```
paulo-ferraz-oliveira, line 6: Lemme know if you prefer a broader "no-line" entry.

lib/absinthe/lexer.ex
```diff
@@ -8,6 +8,8 @@ defmodule Absinthe.Lexer do
   @space 0x0020
   @unicode_bom 0xFEFF
 
+  @stopped_at_token_limit ":stopped_at_token_limit"
+
```
paulo-ferraz-oliveira, line 12: Because `nimble_parsec` specific errors as `String.t()`.

Another solution is to pull request there so the error reason can be `term()`, for example.

mix.exs
```diff
@@ -30,7 +30,8 @@ defmodule Absinthe.Mixfile do
       ],
       deps: deps(),
       dialyzer: [
-        plt_core_path: "priv/plts",
+        plt_add_deps: :apps_direct,
```
paulo-ferraz-oliveira, line 33: For lolspeed 😄

benwilson512 (approved):
Nice, Thanks!

benwilson512:
@paulo-ferraz-oliveira I ended up removing dialyxer from the CI in https://github.com/absinthe-graphql/absinthe/pull/1253 since it immediately had issues w/ PLT caches that I didn't understand. You're welcome to reintroduce it but some guidance about how to manage the caches when adding Elixir / OTP versions would be appreciated.

paulo-ferraz-oliveira:
Caches usually should depend on whatever would invalidate them (in terms of key naming), as in

```
key: "${{ runner.os }}-erlang-${{ matrix.otp }}-elixir-${{ matrix.elixir }}-${{ hashFiles('mix.lock') }}"
```

(OS + OTP version + Elixir version + mix.lock hash)

What problems did you run into?

As for guidance, I think the above should be enough, but maybe you want something different (?)

benwilson512:
You can see the job that failed here https://github.com/absinthe-graphql/absinthe/actions/runs/5408206586/jobs/9827039687

paulo-ferraz-oliveira:
Oh, from the top of my head this might be because we recently released a new `setup-beam` version that changed where stuff was kept (1.16), so the cached elements fail because the cache is no longer found in the same place (OS-wise) and the PLT is static in references.

Would you be willing to delete the caches? I'd pull request again to re-introduce `dialyxir` in CI and you shouldn't see that problem (well, at least not until `setup-beam` breaks again 😢).

Check https://github.com/erlef/setup-beam/issues/214 for some more history.

I just opened an issue for `setup-beam` (to discuss with maintainers) where this would potentially help avoid this class of issues.

benwilson512:
Actually so if you look at this job here https://github.com/absinthe-graphql/absinthe/actions/runs/5423460943/jobs/9861420841 I added a cache key prefix and if you look at the cache action you can see it didn't find any PLT cache, but it still failed.

paulo-ferraz-oliveira:
Yeah, I see that. Strange that it's complaining about a missing function: https://github.com/absinthe-graphql/absinthe/actions/runs/5423460943/jobs/9861420841#step:10:18.

paulo-ferraz-oliveira:
Yeah, so I also got it failing here: https://github.com/paulo-ferraz-oliveira/absinthe/actions/runs/5425961530/jobs/9867389890

Now on to figure stuff out 😄

paulo-ferraz-oliveira:
Not breaking using Elixir 1.14.5, but breaking using 1.15.1. IIrc, this is an intentional change for 1.15.x (lemme look that up).

paulo-ferraz-oliveira:
@benwilson512, I "fixed" it by doing:

```elixir
   def application do
    [extra_applications: [:crypto, :logger | extra_extra_applications(Mix.env())]]
   end
 
  def extra_extra_applications(:test), do: [:syntax_tools]
  def extra_extra_applications(_), do: []
```

but this seems kinda hackish. I don't know what is including `prettypr` in this case, since in other Elixir lib.s I don't see the same issue.

paulo-ferraz-oliveira:
A slightly less hackish (at least consistent with 1.15 release notes change) would be:

```elixir
  def project do
    [
      ...
      prune_code_paths: prune_code_paths(Mix.env())
    ]
  end

  defp prune_code_paths(:test), do: false
  defp prune_code_paths(_), do: true
```

paulo-ferraz-oliveira:
As per https://github.com/paulo-ferraz-oliveira/absinthe/actions/runs/5426289985/jobs/9868142420, from commit https://github.com/paulo-ferraz-oliveira/absinthe/commit/b6c39deef09d93103e84f4de4bb4b7b9cf6235ce, you can see that CI passes (your branch without further changes). If you find the test-specific option `prune_code_paths` set to `false` to have `dialyxir` pass (it's not only in CI, but locally too) I can add it. Otherwise we can wait for a `dialyxir` update that fixes it (?)

I'll open a pull request for that, and we can continue there.

benwilson512:
Thanks for all of the work following up @paulo-ferraz-oliveira!

PR #1249, opened 2023-06-18
Bugfix: multiple pushes per client for subscriptions that have a `context_id`

beligante:
This PR aims to close this issue: https://github.com/absinthe-graphql/absinthe/issues/1064

Originally this issue seems to be introduced by [this PR -- because of this specific change](https://github.com/absinthe-graphql/absinthe/pull/1006/files#diff-c889d345bee83f3e4e6ca1fb71519118225dddbab7b2340bc1fe50644a4719f4R144). It was introduced in the version `v1.6.0` 

On the `v1.5.5` the deduplication happened using `Map.new` when looking at the Registry. By checking the current version, using the same strategy seems to fix the problem since the documents should be identical.

### DIFF
--- a/lib/absinthe/subscription.ex
+++ b/lib/absinthe/subscription.ex
@@ -173,9 +173,7 @@ defmodule Absinthe.Subscription do
     pubsub
     |> registry_name
     |> Registry.lookup(key)
-    |> Enum.uniq_by(fn {_, {doc_id, _doc}} -> doc_id end)
-    |> Enum.map(fn match ->
-      {_, {doc_id, doc}} = match
+    |> Map.new(fn {_, {doc_id, doc}} ->
       doc = Map.update!(doc, :initial_phases, &PipelineSerializer.unpack/1)
 
       {doc_id, doc}
--- a/lib/absinthe/subscription.ex
+++ b/lib/absinthe/subscription.ex
@@ -173,13 +173,13 @@ defmodule Absinthe.Subscription do
     pubsub
     |> registry_name
     |> Registry.lookup(key)
+    |> Enum.uniq_by(fn {_, {doc_id, _doc}} -> doc_id end)
     |> Enum.map(fn match ->
       {_, {doc_id, doc}} = match
       doc = Map.update!(doc, :initial_phases, &PipelineSerializer.unpack/1)
 
       {doc_id, doc}
     end)
-    |> Map.new()
   end
 
   @doc false
--- a/lib/absinthe/subscription.ex
+++ b/lib/absinthe/subscription.ex
@@ -179,6 +179,7 @@ defmodule Absinthe.Subscription do
 
       {doc_id, doc}
     end)
+    |> Map.new()
   end
 
   @doc false
--- a/test/absinthe/execution/subscription_test.exs
+++ b/test/absinthe/execution/subscription_test.exs
@@ -691,8 +691,7 @@ defmodule Absinthe.Execution.SubscriptionTest do
       end)
 
     # assert that all documents are the same
-    first = hd(documents)
-    assert Enum.all?(documents, &(&1 == first))
+    assert [document] = Enum.dedup(documents)
 
     Absinthe.Subscription.publish(
       PubSub,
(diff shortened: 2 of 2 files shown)


test/absinthe/execution/subscription_test.exs
```diff
@@ -49,7 +49,7 @@ defmodule Absinthe.Execution.SubscriptionTest do
     @behaviour Absinthe.Subscription.Pubsub
 
     def start_link() do
-      Registry.start_link(keys: :unique, name: __MODULE__)
+      Registry.start_link(keys: :duplicate, name: __MODULE__)
```
beligante, line 52: Had to change this to allow duplicated keys to make the test fail and also because the registry in the sub supervisor allows duplicated keys: https://github.com/absinthe-graphql/absinthe/blob/master/lib/absinthe/subscription/supervisor.ex#L37

benwilson512: Yup, good call.

test/absinthe/execution/subscription_test.exs
```diff
@@ -678,6 +678,44 @@ defmodule Absinthe.Execution.SubscriptionTest do
...
+    documents =
+      Enum.map(1..5, fn _index ->
+        {:ok, doc} = run_subscription(@query, Schema, context: %{context_id: "global"})
+        doc
+      end)
+
+    # assert that all documents are the same
+    first = hd(documents)
+    assert Enum.all?(documents, &(&1 == first))
+
```
benwilson512, line 696:
```suggestion
    assert [_] = Enum.uniq(documents)
```

lib/absinthe/subscription.ex
```diff
@@ -179,6 +179,7 @@ defmodule Absinthe.Subscription do
 
       {doc_id, doc}
     end)
+    |> Map.new()
```
benwilson512, line 182: You should simply replace the `Enum.map/2` prior to this with a `Map.new/2` call.

benwilson512 (changes requested):
Some minor things to tweak, but overall this looks good.

lib/absinthe/subscription.ex
beligante: Ahhh, nice catch! That would save some unnecessary processing.

beligante: I did the dedup before this enum map: https://github.com/absinthe-graphql/absinthe/pull/1249/commits/2ee47cdc0102a802385e652407b4bac7c2bc665c

test/absinthe/execution/subscription_test.exs
beligante: https://github.com/absinthe-graphql/absinthe/pull/1249/commits/2ee47cdc0102a802385e652407b4bac7c2bc665c

lib/absinthe/subscription.ex
benwilson512: I see that and I think that's generally fine but since order isn't important here `Map.new` is going to involve several less traversals. I'll show what I mean.

lib/absinthe/subscription.ex
```diff
@@ -173,6 +173,7 @@ defmodule Absinthe.Subscription do
     pubsub
     |> registry_name
     |> Registry.lookup(key)
+    |> Enum.uniq_by(fn {_, {doc_id, _doc}} -> doc_id end)
     |> Enum.map(fn match ->
       {_, {doc_id, doc}} = match
```
benwilson512, line 178:
```suggestion
    |> Map.new(fn {_, {doc_id, _doc}} ->
```

lib/absinthe/subscription.ex
beligante: Ahhh gotcha! Let me fix that

beligante: https://github.com/absinthe-graphql/absinthe/pull/1249/commits/97ea4dc1ca699c47547496412b7d443e52092533

benwilson512 (approved):
Fantastic, thanks for tracking this down!

jonatanklosko:
Amazing, thank you!

PR #1255, opened 2023-06-30
Adapt to 1.15's code path pruning, for dialyxir'ing purposes

paulo-ferraz-oliveira:
As per the discussion that started [here](https://github.com/absinthe-graphql/absinthe/pull/1248#issuecomment-1612389235).

Overrides #1254, too.

(I didn't have issues when I pull requested, in #1248, since the update to include 1.15 was done after my proposal for change)

Note on 1.15 potential incompatibilities found [here](https://hexdocs.pm/elixir/1.15/changelog.html#potential-incompatibilities).

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -33,7 +33,8 @@ defmodule Absinthe.Mixfile do
         plt_add_deps: :apps_direct,
         plt_file: {:no_warn, "priv/plts/project.plt"},
         plt_add_apps: [:mix, :dataloader, :decimal, :ex_unit]
-      ]
+      ],
+      prune_code_paths: prune_code_paths(Mix.env())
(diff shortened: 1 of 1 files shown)


mix.exs
```diff
@@ -85,6 +86,9 @@ defmodule Absinthe.Mixfile do
     ]
   end
 
+  defp prune_code_paths(:test), do: false
```
benwilson512, line 89: Which dependency are we not declaring in test?

benwilson512: Oh I see in the other PR conversation, thanks.

benwilson512 (approved):
Seems good until Dialyxir patches.

Issue #1256, opened 2023-07-11
False positive error for non_null validation


fuelen:
### Environment

* Elixir version (elixir -v): 1.14.5
* Absinthe version (mix deps | grep absinthe): 1.7.4
* Client Framework and version (Relay, Apollo, etc): -

### Expected behavior

`non_null(list_of(:string))` should not emit `Cannot return null for non-nullable field` error on a list which contains `nil`

### Actual behavior
After update 1.7.1 ->1.7.4 our tests report an error
`Cannot return null for non-nullable field`

### Relevant Schema/Middleware Code
```
    field :my_test, non_null(list_of(:string)) do
      resolve(fn _, _ -> {:ok, ["V1", nil]} end)
    end
```
^ reports `Cannot return null for non-nullable field`

```
    field :my_test, non_null(list_of(:string)) do
      resolve(fn _, _ -> {:ok, nil} end)
    end
```
^ fails with exception `** (Protocol.UndefinedError) protocol Enumerable not implemented for nil of type Atom`

benwilson512:
Hi @fuelen, I will review, thanks!

PR #1259, opened 2023-08-03
Fix non-null-list-of-non-null error when root data is nil

tamanugi:
Thanks for awesome library ✨

This PR is fix  https://github.com/absinthe-graphql/absinthe/issues/1256

### DIFF
--- a/lib/absinthe/phase/document/execution/resolution.ex
+++ b/lib/absinthe/phase/document/execution/resolution.ex
@@ -285,7 +285,8 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
     |> propagate_null_trimming
   end
 
-  defp maybe_add_non_null_error([], values, %Type.NonNull{of_type: %Type.List{}}) when is_list(values) do
+  defp maybe_add_non_null_error([], values, %Type.NonNull{of_type: %Type.List{}})
+       when is_list(values) do
     values
     |> Enum.with_index()
     |> Enum.filter(&is_nil(elem(&1, 0)))
--- a/lib/absinthe/phase/document/execution/resolution.ex
+++ b/lib/absinthe/phase/document/execution/resolution.ex
@@ -285,7 +285,20 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
     |> propagate_null_trimming
   end
 
-  defp maybe_add_non_null_error([], values, %Type.NonNull{of_type: %Type.List{}})
+  defp maybe_add_non_null_error([], nil, %Type.NonNull{}) do
+    ["Cannot return null for non-nullable field"]
+  end
+
+  # Unwrap the non null so we can check again for the possible case of a non null
+  # inside of a list. We want that clause to handle both
+  # - `non_null(list_of(non_null(:thing)))`
+  # - `list_of(non_null(:thing))`
+  # Thus the single layer of unwrapping here.
+  defp maybe_add_non_null_error([], values, %Type.NonNull{of_type: type}) do
+    maybe_add_non_null_error([], values, type)
+  end
+
+  defp maybe_add_non_null_error([], values, %Type.List{of_type: %Type.NonNull{}})
        when is_list(values) do
     values
     |> Enum.with_index()
@@ -293,10 +306,6 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
     |> Enum.map(fn {_value, index} ->
       %{message: "Cannot return null for non-nullable field", path: [index]}
     end)
-  end
-
-  defp maybe_add_non_null_error([], nil, %Type.NonNull{}) do
-    ["Cannot return null for non-nullable field"]
   end
 
   defp maybe_add_non_null_error(errors, _, _) do
--- a/lib/absinthe/phase/document/execution/resolution.ex
+++ b/lib/absinthe/phase/document/execution/resolution.ex
@@ -285,7 +285,7 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
     |> propagate_null_trimming
   end
 
-  defp maybe_add_non_null_error([], values, %Type.NonNull{of_type: %Type.List{}}) do
+  defp maybe_add_non_null_error([], values, %Type.NonNull{of_type: %Type.List{}}) when is_list(values) do
     values
     |> Enum.with_index()
     |> Enum.filter(&is_nil(elem(&1, 0)))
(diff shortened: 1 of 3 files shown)


benwilson512 (changes requested):
Hey @tamanugi thanks for taking a crack at fixing this!

This fix may be useful on its own, but it is not testing the scenario that the issue outlines. At a core level, the issue is about fields with return type `non_null(list_of(:string))` but you are testing fields with return type `non_null(list_of(non_null(:string)))`.

You need a test that uses field return type `non_null(list_of(:string))` and exhibits the error message in the issue, that is then fixed by your changes.

Your change DOES fix a crash, which is great! This might be the same root cause as the crash described in the issue but since the return types don't match I'm not sure.

Please add in this test and we'll be good to go!

tamanugi:
Hi @benwilson512 ,  Thanks for review ✨ 

You are right, the test should have been for `non_null(list_of(:string))`.

I have fixed the test 👍 
556a8cc9a4d6a5199d10477cd1674e67d8c3cc03

benwilson512:
Hey @tamanugi thanks for your continued effort. I'm going to push some changes up I think that better distinguish between `child` and `element`. In the existing tests `child` refers to a subfield, not an item in the list. I think this ambiguity is making it difficult to determine which test is supposed to cover which scenario.

I'm not entirely convinced that this does the right thing for lists of nullable that return `[nil]` but I should be able to see that more clearly shortly.

test/absinthe/phase/execution/non_null_test.exs
```diff
@@ -263,10 +277,30 @@ defmodule Absinthe.Phase.Document.Execution.NonNullTest do
...
+      ]
+
+      assert {:ok, %{data: data, errors: errors}} == Absinthe.run(doc, Schema)
+    end
+
     test "list of non null things works when child is null" do
       doc = """
       {
-        nonNullListOfNonNull(makeNull: true) { __typename }
+        nonNullListOfNonNull(makeChildNull: true) { __typename }
```
benwilson512, line 303: I don't see why this test is modified.

test/absinthe/phase/execution/non_null_test.exs
```diff
@@ -290,7 +324,7 @@ defmodule Absinthe.Phase.Document.Execution.NonNullTest do
           nonNullListOfNonNull {
             nonNullListOfNonNull {
               nonNullListOfNonNull {
-                nonNullListOfNonNull(makeNull: true) { __typename }
+                nonNullListOfNonNull(makeChildNull: true) { __typename }
```
benwilson512, line 327: This test also should not be modified.

test/absinthe/phase/execution/non_null_test.exs
```diff
@@ -263,10 +277,30 @@ defmodule Absinthe.Phase.Document.Execution.NonNullTest do
...
+
+      errors = [
+        %{
+          locations: [%{column: 3, line: 2}],
+          message: "Cannot return null for non-nullable field",
+          path: ["nonNullListOfNullable"]
+        }
+      ]
+
+      assert {:ok, %{data: data, errors: errors}} == Absinthe.run(doc, Schema)
```
benwilson512, line 297: This test covers the case where an error is returned, but there is still no test case that shows that there is NOT an error for `[nil]` when it's allowed. I'll add one.

test/absinthe/phase/execution/non_null_test.exs
benwilson512: OH I see the semantics of `make_null` was changed.

benwilson512:
OK I see what you were doing now. In the future as a bit of advice for open source contributions please add some comments to the PR to explain why you're changing existing tests. I get why you did it now because we need to distinguish between making the whole field `null` vs making an element in the list `null`. However when PRs change existing tests it always complicates things cause I am not sure why existing test behavior would need to change.

benwilson512:
Right and the thing is if you're changing the argument name then you should change EVERY test that uses that argument because now they're testing different things.

test/absinthe/phase/execution/non_null_test.exs
```diff
@@ -263,10 +277,42 @@ defmodule Absinthe.Phase.Document.Execution.NonNullTest do
...
+      doc = """
+      {
+        nonNullListOfNullable(returnNullElement: true) { __typename }
+      }
+      """
+
+      data = %{"nonNullListOfNullable" => [nil]}
+
+      assert {:ok, %{data: data}} == Absinthe.run(doc, Schema)
+    end
```
benwilson512, line 310: OK! This is the test that the issue is about, and this test fails on this PR. Now to figure out why.

lib/absinthe/phase/document/execution/resolution.ex
```diff
@@ -285,7 +285,21 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
     |> propagate_null_trimming
   end
 
-  defp maybe_add_non_null_error([], values, %Type.NonNull{of_type: %Type.List{}}) do
+  defp maybe_add_non_null_error([], nil, %Type.NonNull{}) do
```
benwilson512, line 288: Moved this up as this the most fundamental check, best to do it first.

lib/absinthe/phase/document/execution/resolution.ex
```diff
@@ -285,7 +285,21 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
...
+  # Unwrap the non null so we can check again for the possible case of a non null
+  # inside of a list. We want that clause to handle both
+  # - `non_null(list_of(non_null(:thing)))`
+  # - `list_of(non_null(:thing))`
+  # Thus the single layer of unwrapping here.
+  defp maybe_add_non_null_error([], values, %Type.NonNull{of_type: type}) do
+    maybe_add_non_null_error([], values, type)
+  end
+
+  defp maybe_add_non_null_error([], values, %Type.List{of_type: %Type.NonNull{}})
```
benwilson512, line 301: Before, this clause would happen for any `non_null(list_of(` regardless of wether the inner type was `non_null` or not. Now we check the inner type properly.

benwilson512 (approved):
Thanks for digging into this, glad we could get it in a good spot!

tamanugi:
@benwilson512 
Thanks for your follow-up and awesome advice✨  
I have limited experience with OSS, so it was very educational for me.

benwilson512:
Thanks for your contribution!

PR #1260, opened 2023-08-11
Add __private__ field to Type.Argument and Type.Enum.Value structs

DReigada:
Most structs under Absinthe.Type have private fields, but `Argument` and `Enum.Value` don't.


lib/absinthe/type/argument.ex
```diff
@@ -15,6 +15,8 @@ defmodule Absinthe.Type.Argument do
   * `:deprecation` - Deprecation information for an argument, usually
     set-up using `Absinthe.Schema.Notation.deprecate/1`.
   * `:description` - Description of an argument, useful for introspection.
+
+  The `__private__` and `__reference__` fields are for internal use.
```
ren90, line 19: Can we specify what uses? 🤔

ren90 (approved):
Seems good

benwilson512 (approved):
Thanks!

PR #1261, opened 2023-08-11
Store less information in the registry

josevalim:
Instead of storing data in the registry keyed by self(),
we put that in the process dictionary. Although the process
dictionary is discouraged, they are equivalent here:
they are both a form of side-effect although pdict is
considerably less expensive.

### DIFF
--- a/lib/absinthe/subscription.ex
+++ b/lib/absinthe/subscription.ex
@@ -170,7 +170,6 @@ defmodule Absinthe.Subscription do
   @doc false
   def unsubscribe(pubsub, doc_id) do
     registry = pubsub |> registry_name
-    self = self()
 
     for field_key <- pdict_fields(doc_id) do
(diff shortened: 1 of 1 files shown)


lib/absinthe/subscription.ex
```diff
@@ -151,20 +151,32 @@ defmodule Absinthe.Subscription do
...
+  end
+
+  defp pdict_delete_fields(doc_id) do
+    Process.delete({__MODULE__, doc_id})
   end
 
   @doc false
   def unsubscribe(pubsub, doc_id) do
     registry = pubsub |> registry_name
     self = self()
```
benwilson512: This variable is now unused.

benwilson512 (approved):
This makes tons of sense, don't know why I didn't think of this originally! Thanks!

PR #1269, opened 2023-08-14
Also pack contexts and variables in subscriptions

josevalim:
This should reduce memory consumption.


benwilson512 (approved):
I need to step through this more to be sure it can't produce cycles in the reverse map but this looks great.

benwilson512:
Thanks!

Issue #1274, opened 2023-09-05
Resolving field of type `[[Value]!]!]` seems to fail with valid values


mattbaker:
### Environment

* Elixir version (elixir -v): Elixir 1.14.0 (compiled with Erlang/OTP 25)
* Absinthe version (mix deps | grep absinthe): master
* Client Framework and version (Relay, Apollo, etc): n/a

### Expected behavior

Example  — a field of type `[[String]!]!` resolving with value `[["test", nil]]` should successfully return that data to the user.

I _think_ this should work, though I do get turned around once things start nesting 😂 

### Actual behavior

In reality we end up with `%{data: nil}` and no error message that I can see. Very strange! My guess is Absinthe is viewing this as an error of some sort.

### Relevant Schema/Middleware Code

I haven't dug into Absinthe yet (and I will), but I am opening a PR now with a failing test just to kick things off: PR https://github.com/absinthe-graphql/absinthe/pull/1275

If anyone has any ideas about where to start let me know! Might save me some time :D

benwilson512:
Lists and non null strike again! Likely related to changes made in this PR https://github.com/absinthe-graphql/absinthe/pull/1259/files

mattbaker:
Weirdly enough we noticed this in our production system first which is on this commit from February, so it might go further back:
https://github.com/absinthe-graphql/absinthe/commits/7eec586ca47a17da265141ad44aad7d98cf3a355

```
  "absinthe": {:git, "https://github.com/absinthe-graphql/absinthe.git", "7eec586ca47a17da265141ad44aad7d98cf3a355", [ref: "7eec586ca47a17da265141ad44aad7d98cf3a355"]},
```

(We needed @stellanor's token limit work)

I'll start digging Thursday or Friday assuming no interruptions and let you know what I find!

PR #1275, opened 2023-09-05
Traverse deeply nested lists when searching for null violations

mattbaker:
Attempts to fix https://github.com/absinthe-graphql/absinthe/issues/1274 by traversing through layers of non-null/list wrappers when looking for non-null constraint violations.


test/absinthe/phase/execution/non_null_test.exs
```diff
@@ -217,6 +240,51 @@ defmodule Absinthe.Phase.Document.Execution.NonNullTest do
...
+      {
+        deeplyNestedNonNullableListOfNullable
+      }
+      """
+
+      assert {:ok, %{data: %{"deeplyNestedNonNullableListOfNullable" => [[[["ok", nil]]]]}}} ==
+               Absinthe.run(doc, Schema)
+    end
+
+    test "deeply nested non-nullable value inside non-nullable lists cannot be null" do
```
mattbaker, line 265: This wasn't the issue I saw when I started digging into this, but this test also fails on master. I think it's in the same theme of needing to traverse through multiple layers of lists/non-nulls.

test/absinthe/phase/execution/non_null_test.exs
```diff
@@ -217,6 +240,51 @@ defmodule Absinthe.Phase.Document.Execution.NonNullTest do
   end
 
   describe "lists" do
+    test "non-null list of non-null list of nullable value returns null value" do
```
mattbaker, line 243: This seems like it should be redundant when `test "deeply nested nullable value inside non-nullable lists can be null"` exists below, but the latter actually passes in master for the same reason `test "deeply nested non-nullable value inside non-nullable lists cannot be null"` fails (I think). 

So I ended up keeping this test just for my own sanity.

mattbaker:
@benwilson512: Full transparency — this is a complex file and I don't know if what I'm doing is the right thing here, it could definitely benefit from your eyes!

It seems like, one way or another, this boils down to the fact that a sufficiently deep wrapping of `non_null(list_of(non_null(list_of(......` will cause issues? I took a stab at making the "add errors" codes and the null propagation code more recursive and it seemed to do the trick. My only worry is I feel like I don't fully grok everything in this module and I may have done the wrong thing and just gotten lucky 😱

mattbaker:
@benwilson512 - Curious if you have any thoughts, but I'm also sure you're busy and I know this is probably more than a surface fix! I could try to set some time aside this month to dive deeper and build some more familiarity myself and maybe emerge with a bit more confidence too, if you think you won't have time to take a look. It's also not super urgent for us.

lib/absinthe/phase/document/execution/resolution.ex
```diff
@@ -285,30 +285,28 @@ defmodule Absinthe.Phase.Document.Execution.Resolution do
...
-       when is_list(values) do
-    values
-    |> Enum.with_index()
-    |> Enum.filter(&is_nil(elem(&1, 0)))
-    |> Enum.map(fn {_value, index} ->
-      %{message: "Cannot return null for non-nullable field", path: [index]}
+  defp maybe_add_non_null_error([], [_ | _] = values, %Type.List{of_type: type}, path) do
+    Enum.with_index(values)
+    |> Enum.flat_map(fn {value, index} ->
+      maybe_add_non_null_error([], value, type, path ++ [index])
```
benwilson512, line 305: Appending to this list here will be `O(n^2)` and the lists could potentially be quite large. I'm working to refactor this.

benwilson512 (approved):
Sorry this took me so long to get to, thanks Matt!

lib/absinthe/phase/document/execution/resolution.ex
mattbaker: Oh very good catch, I breezed right past that!

mattbaker:
Not a problem at all! It was an odd edge case that was on our "to do" list but wasn't urgent at all. 

Thanks so much!

PR #1277, opened 2023-10-27
Update incorrect GraphQL website URLs

xtian:
Not sure why these links would go to another (unmaintained) Elixir implementation of GraphQL


benwilson512 (approved):
That does seem odd!

PR #1287, opened 2023-11-16
update locked deps

inoas-nbw:


benwilson512:
Hi @inoas-nbw I think I'll handle this manually in main after #1286 and #1285 are in.

PR #1290, opened 2023-11-20
Improve typechecks on nested args

polleuretan:
This PR adds check to see if the variable type matches the argument type for nested arguments. Currently this check works only for "top level" arguments, but misses the ones used inside input objects (see the test case in PR).

Specifically, documents of the form:

```
  mutation ($value: String) {
    updateThing(id: "foo", thing: {value: $value}) {
      name
      value
    }
  }
```

where the variable `$value` is used inside of an input object literal were not checked properly prior to this PR.

This is a bugfix, but will have the effect that some previously valid documents are now marked as invalid.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## Unreleased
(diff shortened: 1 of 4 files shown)


gsychev:
This is a breaking change and it was a misleading thing to release it in patch version bump. This must've been an __at least__ minor version bump.

This thing has consumed days of work from several engineers in my project and we were very close to breaking production.

PR #1291, opened 2024-01-03
Support deprecated input object fields in introspection query (fixes #1241)

### LINKED ISSUE #1241: Support deprecated input object fields in introspection query
### Environment

* Elixir version (elixir -v): 1.14.2-otp-25
* Absinthe version (mix deps | grep absinthe): 1.7.1
* Client Framework and version (Relay, Apollo, etc): Apollo v3.7.5 

### Expected behavior

When making an introspection query to get data about fields, I'd like to query for deprecated fields on input objects as defined in the [graphql-spec](https://github.com/graphql/graphql-spec/pull/805) 
### Actual behavior

With this [PR](https://github.com/apollographql/apollo-kotlin/pull/4678/files) Apollo devs added `(includeDeprecated: true)` as a param to `inputFields` in the `fragment` portion of the GraphQL query to download a schema.
This param is not currently supported by Absinthe and the Gradle plugin has stopped working starting from this Apollo version (`3.7.5`).

Also Close #1113

jgautsch:
I was setting up Apollo iOS in a new mobile app project, hoping to have it talk to my absinthe API. Apollo introspects the schema to generate types, however the [introspection query](https://gist.github.com/jgautsch/d95309a809e9ef8afd3ed49606071558) wasn't processable by absinthe.

The issue was the `(includeDeprecated: true)` param passed to various fields including `inputFields`. 

The `includeDeprecated` arg for input fields was [merged](https://github.com/dchang-dchang/graphql/commit/9ef8c9b845f2329dfa64f88e02f854c0354c1d41) into the graphql spec back in Aug 2022.

This PR brings absinthe up to support that arg, and makes Apollo's introspection query work.

Feedback welcome!

### DIFF
--- a/lib/absinthe/type/built_ins/introspection.ex
+++ b/lib/absinthe/type/built_ins/introspection.ex
@@ -80,7 +80,7 @@ defmodule Absinthe.Type.BuiltIns.Introspection do
           source.args
           |> Map.values()
           |> Enum.filter(fn %{deprecation: is_deprecated} ->
-            !is_deprecated || (is_deprecated && show_deprecated)
+            !is_deprecated || show_deprecated
           end)
           |> Enum.sort_by(& &1.identifier)
 
@@ -122,7 +122,7 @@ defmodule Absinthe.Type.BuiltIns.Introspection do
                 Absinthe.Type.introspection?(field) ->
                   []
 
-                !is_deprecated || (is_deprecated && show_deprecated) ->
+                !is_deprecated || show_deprecated ->
                   [field]
 
                 true ->
@@ -185,7 +185,7 @@ defmodule Absinthe.Type.BuiltIns.Introspection do
             values
             |> Map.values()
             |> Enum.filter(fn %{deprecation: is_deprecated} ->
-              !is_deprecated || (is_deprecated && show_deprecated)
+              !is_deprecated || show_deprecated
             end)
             |> Enum.sort_by(& &1.value)
 
@@ -210,7 +210,7 @@ defmodule Absinthe.Type.BuiltIns.Introspection do
             fields
             |> Map.values()
             |> Enum.filter(fn %{deprecation: is_deprecated} ->
-              !is_deprecated || (is_deprecated && show_deprecated)
+              !is_deprecated || show_deprecated
             end)
             |> Enum.sort_by(& &1.identifier)
 
@@ -253,7 +253,7 @@ defmodule Absinthe.Type.BuiltIns.Introspection do
           args
           |> Map.values()
           |> Enum.filter(fn %{deprecation: is_deprecated} ->
-            !is_deprecated || (is_deprecated && show_deprecated)
+            !is_deprecated || show_deprecated
           end)
           |> Enum.sort_by(& &1.identifier)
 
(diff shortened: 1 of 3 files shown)


jgautsch:
Fixes #1241

lib/absinthe/type/built_ins/introspection.ex
```diff
@@ -191,11 +203,24 @@ defmodule Absinthe.Type.BuiltIns.Introspection do
...
           input_fields =
             fields
-            |> Map.values()
+            |> Enum.flat_map(fn {_, %{deprecation: is_deprecated} = value} ->
+              if !is_deprecated || (is_deprecated && show_deprecated) do
+                [value]
+              else
+                []
+              end
+            end)
```
benwilson512, line 223: Ah I see this is extracting the values and filtering in one go. I am not entirely sure that this is actually better, I think it'd be clearer as:

```
|> Map.values
|> Enum.filter(fn %{deprecation: is_deprecated} -> !is_deprecated || (is_deprecated && show_deprecated) end)
```

When you `some_map |> Enum.flat_map` Enumerable ends up calling `Enum.to_list` on the map anyway so we may as well do `Map.values |> Enum.filter`.

lib/absinthe/type/built_ins/introspection.ex
```diff
@@ -226,10 +251,22 @@ defmodule Absinthe.Type.BuiltIns.Introspection do
...
         args =
           args
-          |> Map.values()
+          |> Enum.flat_map(fn {_, %{deprecation: is_deprecated} = value} ->
+            if !is_deprecated || (is_deprecated && show_deprecated) do
+              [value]
+            else
+              []
+            end
+          end)
```
benwilson512, line 269: Same as the earlier comment related to filtering.

benwilson512 (changes requested):
Hey! Thanks for taking this on! I have those minor code tweaks and otherwise this looks good to go.

jgautsch:
@benwilson512 Good callout– updated PR. There was another existing chunk following the same `flat_map` pattern, I went ahead and updated it to `Map.values() |> Enum.filter()` as well for consistency. I also pushed a commit to simplify the boolean expressions. 

Thanks for creating & maintaining absinthe! It's wonderful.

lib/absinthe/type/built_ins/introspection.ex
```diff
@@ -226,10 +242,19 @@ defmodule Absinthe.Type.BuiltIns.Introspection do
...
+          default_value: false
+        ]
+      ],
+      resolve: fn %{include_deprecated: show_deprecated}, %{source: %{args: args}} ->
         args =
           args
           |> Map.values()
+          |> Enum.filter(fn %{deprecation: is_deprecated} ->
+            !is_deprecated || show_deprecated
+          end)
```
benwilson512, line 257: Sorry as one last revision: I'm not normally super DRY about things but this is the same function call 4 times over in the same module, can you extract this into a `defp filter_deprecated` function?

jgautsch: Good idea 👍

jgautsch:
@benwilson512 lmk if there are any other improvements you'd see prior to merging

jgautsch:
@benwilson512 any other changes you'd suggest prior to being able to merge?

PR #1292, opened 2024-01-10
assign correct field_def value

barnabasJ:
Hi, I think I found some kind of copy/paste issue


benwilson512 (approved):
Yes this looks like an issue, good catch!

PR #1295, opened 2024-02-05
Add `docset_runner` option to Subscriptions to use a custom docset runner

bryanjos:
Adds an optional `opts` parameter to both `Absinthe.Subscription.publish` and `Absinthe.Subscription.Local.publish_mutation`. The `opts` can include a `docset_runner` that allows for developers to use a custom docset runner to process subscriptions. 

It also updates the `Absinthe.Subscription` middleware to pass along the opts in the second parameter to the call to publish. This should allow developers to update the middleware to include the `docset_runner` option for triggered subscriptions.

### DIFF
--- a/lib/absinthe/subscription.ex
+++ b/lib/absinthe/subscription.ex
@@ -240,9 +240,9 @@ defmodule Absinthe.Subscription do
 
   ## Middleware callback
   @doc false
-  def call(%{state: :resolved, errors: [], value: value} = res, _) do
+  def call(%{state: :resolved, errors: [], value: value} = res, opts) do
     with {:ok, pubsub} <- extract_pubsub(res.context) do
-      __MODULE__.publish(pubsub, value, res)
+      __MODULE__.publish(pubsub, value, res, opts)
     end
 
     res
--- a/lib/absinthe/subscription.ex
+++ b/lib/absinthe/subscription.ex
@@ -99,22 +99,34 @@ defmodule Absinthe.Subscription do
     other_user_subscription_field: user.id,
   ])
   ```
+
+  ## Options
+  * `:docset_runner` - A custom function used iterate over the subscription docs and
+  publishes them using the pubsub module. The default docset_runner iterates over the subscription docs and runs them sequentially.`
   """
   @spec publish(
           Absinthe.Subscription.Pubsub.t(),
           term,
-          Absinthe.Resolution.t() | [subscription_field_spec]
+          Absinthe.Resolution.t() | [subscription_field_spec],
+          Keyword.t()
         ) :: :ok
-  def publish(_pubsub, _mutation_result, []), do: :ok
+  def publish(
+        _pubsub,
+        _mutation_result,
+        _subscribed_fields,
+        opts \\ []
+      )
 
-  def publish(pubsub, mutation_result, %Absinthe.Resolution{} = info) do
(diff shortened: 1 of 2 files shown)


bryanjos:
Hey @benwilson512 this is the PR my teammate @michaelcaterisano mentioned we would make in the conversation in #1218.

I noticed in the moduledoc for `Absinthe.Subscription` about plans for batching subscription and more control over back pressure. I don't know if this gets in the way of those plans at all. That is part of idea behind this change for us.

bryanjos:
Thinking about this more, is there a place that one configure that the custom function be used for all subscriptions? That would make it simpler for the triggered subscriptions case. And the publish functions could still have the option for the custom docset runner if they need to override it for a specific case.

michaelcaterisano:
> Thinking about this more, is there a place that one configure that the custom function be used for all subscriptions? That would make it simpler for the triggered subscriptions case. And the publish functions could still have the option for the custom docset runner if they need to override it for a specific case.

@benwilson512 I had the same thought and initially tried to add this in the `config` function opts, where custom `context_id` and `document_id` functions can be passed, but I got stuck trying to access that config from the code that chooses the docset. Do you have any suggestions about a better spot for this config?

michaelcaterisano:
@benwilson512 Just following up, any feedback would be appreciated. We'd love to move off of our fork. Happy to make adjustments to this approach if you have thoughts.

benwilson512:
Sorry just saw the question, let me dig in further.

michaelcaterisano:
> Sorry just saw the question, let me dig in further.

Thanks!

cschiewek:
@benwilson512 any chance we could get this merged? (and maybe a release?) 🥺

benwilson512:
Hey folks! I had a chance to look at this further, and I wonder if something of a middleground between https://github.com/absinthe-graphql/absinthe/pull/1218 and this would work. What if we added an `execute_documents` optional callback to the `Absinthe.Subscription.Pubsub` behaviour?

I think @bryanjos this would solve you concern about having to pass in the `opts` from different places, while still being a more minimal change than having a behaviour for just the docset runner specifically. Thoughts?

bryanjos:
@benwilson512 Adding an optional callback to `Absinthe.Subscription.Pubsub` makes sense to me. And yes it would solve the concern about the opts and triggered subscriptions. 

The doc storage/ retrieval mechanism as a behaviour also makes sense to me. In cases where developers might want to have a different strategy. It's something we've thought about before, but don't have a need for a this moment. But it would make it easier to do something like use redis instead of what's in place now for instance. That might not be the best example 😄 . Having more flexibility for subscriptions would definitely be a plus for us.

Is there a path we can take to allow for the custom `run_docset`/`execute_documents` functionality while leaving the door open for custom doc storage/ retrieval mechanism in case absinthe wants to go that route?

bryanjos:
@benwilson512 I had to close the PR, but I can open up another one with the changes you proposed to `Absinthe.Subscription.Pubsub`. Let me know if that's still ok to do and/or if `execute_documents` should be different from `run_docset` in any way.

michaelcaterisano:
> What if we added an execute_documents optional callback to the Absinthe.Subscription.Pubsub behaviour?

@benwilson512 I took this approach in #1305. Let me know what you think.

>  I want to make the whole doc storage / retrieval mechanism a behaviour...

We'd be happy to help out with this.

PR #1298, opened 2024-02-14
Use Task.yield/2 instead of Task.await/2 in batching fn, so we can log error

VladoPlavsic:
- Improve absinthe batch logging, add useful information on what module what function with what parameters is failing when timeout occurs

- Instead of calling `Task.await/2` which kills process after timeout has expired, use `Task.yield/2`, if it timeouts, kill task and log which function failed to resolve, after which we continue to act like `Task.await/2` and kill calling process (self).

### DIFF
--- a/lib/absinthe/middleware/batch.ex
+++ b/lib/absinthe/middleware/batch.ex
@@ -60,6 +60,8 @@ defmodule Absinthe.Middleware.Batch do
   @behaviour Absinthe.Middleware
   @behaviour Absinthe.Plugin
 
+  require Logger
+
   @typedoc """
   The function to be called with the aggregate batch information.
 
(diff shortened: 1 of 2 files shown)


benwilson512 (approved):
Great change, thanks!

rossvz:
@VladoPlavsic @benwilson512 I ended up trying to reimplement this today and noticed its merged to main, but not in latest -v1.7.6. Is there any chance of this making it into a release soon? This would help us a ton with tracking down failed batch resolvers in our API!

katafrakt:
By the way, I wonder if it's good for a library to just log things like this. What if someone has different logging format or needs or even some security considerations of logging the whole `fun`? Wouldn't it be better to execute a telemetry event instead and everyone could subscribe to it and log as they want?

PR #1299, opened 2024-02-23
sort_maps: true added to mix absinthe.schema.sdl, to reduce churn

jdewar:
### Summary

We save the output of `mix absinthe.schema.sdl` in our repo so our JS front end can consume/use it.  Minor changes to the Absinthe schema often result in a huge amount of churn in the outputted file; our last PR had an extra +7000 -7000 lines from this.   This PR adds an option from Inspect.Opts to make the output of the sdl file deterministic.

The :sort_maps option was added in Elixir 1.14.4+, but the :custom_options field that it is passed to has been around since 1.9, so Absinthe (which is compatible starting at 1.11) will ignore the option, but not get confused by it if your Elixir version is not high enough.

### Notable Changes
* added `sort_maps: true` to the inspect call that outputs SDL

### DIFF
--- a/lib/mix/tasks/absinthe.schema.sdl.ex
+++ b/lib/mix/tasks/absinthe.schema.sdl.ex
@@ -72,7 +72,7 @@ defmodule Mix.Tasks.Absinthe.Schema.Sdl do
              schema.__absinthe_blueprint__(),
              pipeline
            ) do
-      {:ok, inspect(blueprint, pretty: true)}
+      {:ok, inspect(blueprint, pretty: true, custom_options: [sort_maps: true])}
     else
       _ -> {:error, "Failed to render schema"}
     end


benwilson512 (commented):
Great call, thanks. This is definitely a bit of a tricky thing to test for. I'm also sort of surprised it has an impact because the SDL doesn't (as far as I am aware) display any Elixir maps.

Do you have a schema that demonstrates the ordering issue that this then helps?

jdewar:
A deeper on dive on what was going wrong makes me think there's not a real problem here, so I'm going to close the issue.  I tried to reproduce major churn with simple edits and could not.

I believe we were seeing huge churns because
1) We have been breaking large chunks of the object definitions out into separate files and import_types'ing them into the main file.  This causes much churn -- it would be nice if (assuming SDL doesn't care about ordering) any rearrangement of the Absinthe objects did not change the SDL.
2) *Because* we save our generated files, there may have been some times where we did not re-run the relay-compiler, so subsequent PRs that thought they were changing a single field would be met with multi hundred/thousand line changes.  This is fixed by asserting that the SDL file did not change in CI with an additional relay-compiler run.

PR #1301, opened 2024-02-28
Mc/camelize custom schema directives

michaelcaterisano:


michaelcaterisano:
oops, meant to open this against our branch first, but trying to address custom schema directives not being camelCased in sdl

PR #1302, opened 2024-02-28
Mc/camelize directives

michaelcaterisano:
We're using custom schema directives in an Apollo federated schema. I noticed that multi-word directives were showing up as snake_case in the schema rather than camelCase. The change to `notation.ex` took care of camelCasing directives as defined in the `sdl_render_test.exs`, but the additional change in `sdl_render.ex` seemed to be necessary to camelCase custom schema directives defined in a federated subgraph. I'm not sure how to test that here since defining them relies on Apollo's `composeDirective` directive:

```
  extend schema do
    directive(:link, url: "https://specs.apollo.dev/federation/v2.5", import: ["@composeDirective"])
    directive(:link, url: "https://some-url.com", import: ["@disallowedForSubscriptions"])
    directive(:composeDirective, name: "@disallowedForSubscriptions")
  end
```

Any feedback would be appreciated cc: @benwilson512

### DIFF
--- a/lib/absinthe/schema/notation.ex
+++ b/lib/absinthe/schema/notation.ex
@@ -1991,6 +1991,7 @@ defmodule Absinthe.Schema.Notation do
   defp default_name(Schema.DirectiveDefinition, identifier) do
     identifier
     |> Atom.to_string()
+    |> Absinthe.Utils.camelize(lower: true)
   end
 
   defp default_name(_, identifier) do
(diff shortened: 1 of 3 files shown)


benwilson512 (approved):
This looks good to me, thanks!

michaelcaterisano:
@benwilson512 I assumed the tests would run in CI; I just ran the whole suite locally and it seems this broke some tests

PR #1303, opened 2024-02-29
Fix #1302, camelize type system directives in sdl output

michaelcaterisano:
This is a follow-up to #1302. I camelized in one too many places and wasn't testing type system directives in the way I intended, e.g. defined in a prototype schema. This PR removes the extra call to `Absinthe.Utils.camelize` and corrects the test to use a prototype schema.

cc: @benwilson512

### DIFF
--- a/lib/absinthe/schema/notation.ex
+++ b/lib/absinthe/schema/notation.ex
@@ -1991,7 +1991,6 @@ defmodule Absinthe.Schema.Notation do
   defp default_name(Schema.DirectiveDefinition, identifier) do
     identifier
     |> Atom.to_string()
-    |> Absinthe.Utils.camelize(lower: true)
   end
 
   defp default_name(_, identifier) do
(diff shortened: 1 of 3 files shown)


lib/absinthe/schema/notation.ex
```diff
@@ -1991,7 +1991,6 @@ defmodule Absinthe.Schema.Notation do
   defp default_name(Schema.DirectiveDefinition, identifier) do
     identifier
     |> Atom.to_string()
-    |> Absinthe.Utils.camelize(lower: true)
```
michaelcaterisano, line 1994: This did succeed in `camelCasing` the directive in the test below, but it broke a bunch of other things.

test/absinthe/schema/sdl_render_test.exs
```diff
@@ -208,10 +208,6 @@ defmodule Absinthe.Schema.SdlRenderTest do
       on :field
     end
 
-    directive :foo_foo do
-      on :field
-    end
```
michaelcaterisano, line 213: I thought this was a type system directive definition, but we're defining those using a [prototype schema](https://hexdocs.pm/absinthe/Absinthe.Schema.Prototype.html). I'm not clear on what kind of directive definition this is.

benwilson512 (approved):
Thanks for this. The test suite around some of these macro things clearly isn't adequate, thanks for adding things!

PR #1308, opened 2024-03-25
Use versioned main page link

HarshBalyan:
Currently, the link takes to the latest version of the main page.
I was working with an older version and expected the link to navigate me to the docs of the absinthe version that I had selected.

This fix will keep take the user to the main page of selected version of docs.


benwilson512 (approved):
Thanks!

PR #1309, opened 2024-03-26
Bail from subscription if resolver has an error

barnabasJ:
Would it be possible to do something like this? Right now, null is sent down to the subscriber if the resolver produces an error. 

It might depend on the use case, but I was thinking about something where you have a table where users can only see some of the entries and you have a subscription to updates, for example and use the user_id as the context id. Now, in the resolver, you check if the user is authorized to see the updated value and if not, put an error into the resolution. With the current implementation, the user would still get a push from the subscription with the null value. But not notifying the user in this case might be the better thing to do.

### DIFF
--- a/lib/absinthe/subscription/local.ex
+++ b/lib/absinthe/subscription/local.ex
@@ -58,16 +58,20 @@ defmodule Absinthe.Subscription.Local do
           ]
         ]
 
-        {:ok, %{result: data}, _} = Absinthe.Pipeline.run(doc.source, pipeline)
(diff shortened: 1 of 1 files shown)


benwilson512:
Hey @barnabasJ I agree that it's awkward to send `null`. However, we can't just suppress all publications that have errors. GraphQL results can contain both data and errors.

My plan for a while was to have some sort of special return value from Subscription resolvers that told Absinthe to halt and not send. Can you open up an issue to discuss this? (Take a look around, one may already be open).

PR #1321, opened 2024-06-12
add telemetry phase when subscription config errors

michaelcaterisano:
### What

Ensuring that telemetry events are fired when subscription config returns an error

### Why

We would like to be able to hook into these events even when subscription config returns an error.

### Testing

Wrote a new test to check that this functionality works. I'd also note that there are three failing tests on `main` that are unrelated to this change.


cschiewek (approved):
I think there's room for improvement here, as this can still potentially discard phases we might want to run, but likely out of scope for this PR. 👍

benwilson512 (approved):

PR #1322, opened 2024-06-13
Check absinthe dependent libraries on CI runs

bryanjos:
This PR adds a GitHub action to check the libraries that depend on Absinthe to ensure they still work with the changes added in PRs.

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -9,27 +9,25 @@ on:
 jobs:
   test:
     name: Elixir ${{matrix.elixir}} / OTP ${{matrix.otp}}
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
 
     strategy:
       matrix:
         elixir:
-          - "1.13"
-          - "1.14"
-          - "1.15"
+          - '1.15'
+          - '1.16'
+          - '1.17'
         otp:
-          - "24"
-          - "25"
-          - "26"
-        include:
-          - elixir: "1.15"
-            otp: "26"
-            format: true
+          - '25'
+          - '26'
+          - '27'
         exclude:
-          - otp: 26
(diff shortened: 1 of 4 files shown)


benwilson512:
Oh this reminds me, we should update the Elixir versions we test against, particularly with 1.17 out. We should only do 1.15, 1.16, and 1.17

.github/workflows/elixir.yml
```diff
@@ -14,22 +14,21 @@ jobs:
...
+          - '1.17'
         otp:
-          - "24"
-          - "25"
-          - "26"
-        include:
-          - elixir: "1.15"
-            otp: "26"
-            format: true
+          - '24'
```
benwilson512, line 21: I'm fine dropping this one at this point.

maartenvanvliet:
Is it failing because of the change/fix in the `String.jaro_distance/2` in Elixir? https://github.com/elixir-lang/elixir/pull/13369

bryanjos:
@maartenvanvliet thanks for that! I've been trying to figure it out. Especially since it passes locally with Elixir 1.17 and OTP 27, though I'm on a mac. I'll look and see if that's the issue.

bryanjos:
@benwilson512 I think this is good to go now! Aside from the CI changes, I lowered the jaro_distance slightly (thanks again @maartenvanvliet!), and fixed an issue dialyzer found.

benwilson512 (approved):
Really love how clean this is, thanks!

maartenvanvliet:
> @benwilson512 I think this is good to go now! Aside from the CI changes, I lowered the jaro_distance slightly (thanks again @maartenvanvliet!), and fixed an issue dialyzer found.

No problem, saw the CI error and vaguely recalled the change in elixir a few months back

PR #1333, opened 2024-07-31
Set include_deprecated default value to true for backwards compatibility

### LINKED ISSUE #1327: Mix.Tasks.Absinthe.Schema.Json nolonger includes deprecated input fields

### Environment

* Elixir version (elixir -v): 
    Erlang/OTP 25 [erts-13.2.2.4] [source] [64-bit] [smp:10:10] [ds:10:10:10] [async-threads:1]
    Elixir 1.14.1 (compiled with Erlang/OTP 25)

    absinthe 1.7.7 (Hex package) (mix)
    locked at 1.7.7 (absinthe) 21455198

### Expected behavior

In prior versions, when running 
```
  mix absinthe.schema.json --schema MyApp.Schema --json-codec AbsintheSortingCodec --pretty ./schema.json > /dev/null
```
the generated schema.json file would contain input fields that were marked as deprecated.

### Actual behavior

In this version, running the same task omits the deprecated input fields.
No errors are display.

Hacking the absinthe/priv/graphql/introspection.graphql file at line 44 to be `inputFields(includeDeprecated: true) {` instead of just `inputFields {` adds back the deprecated input fields.

filipemonteiroth:
## Context

Absinthe introspection has always included deprecated fields, but there were a few misleading tests.

For example, [this test](https://github.com/absinthe-graphql/absinthe/blob/main/test/absinthe/introspection_test.exs#L214-L238) was targeting the ProfileInput object, which had no deprecated fields until #1291 added one.

The test mentioned above continued working successfully because the include_deprecated argument default was set to false.

## Purpose

Since Absinthe has always included deprecated fields for args, input_fields, and field, I recommend setting the default value of include_deprecated to true for these categories to maintain backwards compatibility (changed by #1291).

## Considerations

* `enum_values` and `__type` already had include_deprecated default set to false before adding include_deprecated to other fields. I understand the inconsistency this PR introduces, with some defaults set to true and others to false. However, the trade-off seems worth it to maintain backwards compatibility. If we decide to set include_deprecated to false by default, we could highlight this breaking change more explicitly in the CHANGELOG.

Fixes #1327

### DIFF
--- a/guides/introspection.md
+++ b/guides/introspection.md
@@ -91,6 +91,43 @@ Getting the name of the fields for a named type:
 Note that you may have to nest several depths of `type`/`ofType`, as
 type information includes any wrapping layers of [List](https://spec.graphql.org/October2021/#sec-List) and/or [NonNull](https://spec.graphql.org/October2021/#sec-Non-Null).
 
+## Configuration
+
+The introspection system can be configured through your application config:
+
+```elixir
+config :absinthe,
+  include_deprecated: false  # Defaults to false, set to true for backward compatibility
+```
+
+This configuration affects the default behavior of the `includeDeprecated` argument in introspection queries. When set to `false` (the default), deprecated fields and values will not be included in introspection results unless explicitly requested via the `includeDeprecated` argument. When set to `true`, deprecated fields and values will be included by default, matching the behavior of previous versions of Absinthe.
+
+For example, with the default configuration (`include_deprecated: false`), this query will not include deprecated fields:
+
+```graphql
+{
+  __type(name: "User") {
+    fields {
+      name
+    }
+  }
+}
+```
+
+To include deprecated fields, you must explicitly set the `includeDeprecated` argument:
+
+```graphql
+{
+  __type(name: "User") {
+    fields(includeDeprecated: true) {
+      name
+    }
+  }
+}
+```
+
+If you set `include_deprecated: true` in your configuration, the first query would include deprecated fields by default, matching the behavior of previous versions of Absinthe.
+
 ## Using GraphiQL
 
 The [GraphiQL project](https://github.com/graphql/graphiql) is
--- a/guides/introspection.md
+++ b/guides/introspection.md
@@ -102,6 +102,8 @@ config :absinthe,
 
 This configuration affects the default behavior of the `includeDeprecated` argument in introspection queries. When set to `false` (the default), deprecated fields and values will not be included in introspection results unless explicitly requested via the `includeDeprecated` argument. When set to `true`, deprecated fields and values will be included by default, matching the behavior of previous versions of Absinthe.
 
+Note: Setting `include_deprecated` to true will break compatibility with the GraphQL specification, which recommends not including deprecated fields by default. Only use this option if you need to maintain compatibility with existing code that expects deprecated fields to be included.
+
 For example, with the default configuration (`include_deprecated: false`), this query will not include deprecated fields:
 
 ```graphql
--- a/lib/absinthe/type/built_ins/introspection.ex
+++ b/lib/absinthe/type/built_ins/introspection.ex
@@ -72,7 +72,7 @@ defmodule Absinthe.Type.BuiltIns.Introspection do
       args: [
         include_deprecated: [
           type: :boolean,
-          default_value: false
+          default_value: true
         ]
       ],
       resolve: fn %{include_deprecated: show_deprecated}, %{source: source} ->
@@ -187,7 +187,7 @@ defmodule Absinthe.Type.BuiltIns.Introspection do
       args: [
         include_deprecated: [
           type: :boolean,
-          default_value: false
+          default_value: true
         ]
       ],
       resolve: fn
@@ -230,7 +230,7 @@ defmodule Absinthe.Type.BuiltIns.Introspection do
       args: [
         include_deprecated: [
           type: :boolean,
-          default_value: false
+          default_value: true
         ]
       ],
       resolve: fn %{include_deprecated: show_deprecated}, %{source: %{args: args}} ->
--- a/lib/absinthe/type/built_ins/introspection.ex
+++ b/lib/absinthe/type/built_ins/introspection.ex
@@ -72,7 +72,7 @@ defmodule Absinthe.Type.BuiltIns.Introspection do
       args: [
         include_deprecated: [
           type: :boolean,
-          default_value: true
+          default_value: default_include_deprecated()
         ]
       ],
       resolve: fn %{include_deprecated: show_deprecated}, %{source: source} ->
@@ -165,7 +165,7 @@ defmodule Absinthe.Type.BuiltIns.Introspection do
       args: [
         include_deprecated: [
           type: :boolean,
-          default_value: false
+          default_value: default_include_deprecated()
         ]
       ],
       resolve: fn
@@ -187,7 +187,7 @@ defmodule Absinthe.Type.BuiltIns.Introspection do
       args: [
         include_deprecated: [
           type: :boolean,
-          default_value: true
+          default_value: default_include_deprecated()
         ]
       ],
       resolve: fn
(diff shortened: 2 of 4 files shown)


filipemonteiroth:
@benwilson512 maybe I could have your review here?

cschiewek:
@filipemonteiroth The problem is that the spec requires that the default be false.  

From [the spec]:
```graphql
type __Type {
  kind: __TypeKind!
  name: String
  description: String
  # must be non-null for OBJECT and INTERFACE, otherwise null.
  fields(includeDeprecated: Boolean! = false): [__Field!]
  # must be non-null for OBJECT and INTERFACE, otherwise null.
  interfaces: [__Type!]
  # must be non-null for INTERFACE and UNION, otherwise null.
  possibleTypes: [__Type!]
  # must be non-null for ENUM, otherwise null.
  enumValues(includeDeprecated: Boolean! = false): [__EnumValue!]
  # must be non-null for INPUT_OBJECT, otherwise null.
  inputFields(includeDeprecated: Boolean! = false): [__InputValue!]
  # must be non-null for NON_NULL and LIST, otherwise null.
  ofType: __Type
  # may be non-null for custom SCALAR, otherwise null.
  specifiedByURL: String
}
```

> Since Absinthe has always included deprecated fields for args, input_fields, and field, I recommend setting the default value of include_deprecated to true for these categories to maintain backwards compatibility.

If the trade-off is backwards compatibility vs spec-compliance, I would we chose spec compliance, but I'm happy to reconsider if other maintainers feel differently cc: (@absinthe-graphql/core) 

I'd like to understand how this is preventing y'all from upgrading.  Maybe we can find an alternative fix?

filipemonteiroth:
> @filipemonteiroth The problem is that the spec requires that the default be false.
> 
> From [the spec](https://github.com/graphql/graphql-spec/blob/main/spec/Section%204%20--%20Introspection.md):
> 
> ```graphql
> type __Type {
>   kind: __TypeKind!
>   name: String
>   description: String
>   # must be non-null for OBJECT and INTERFACE, otherwise null.
>   fields(includeDeprecated: Boolean! = false): [__Field!]
>   # must be non-null for OBJECT and INTERFACE, otherwise null.
>   interfaces: [__Type!]
>   # must be non-null for INTERFACE and UNION, otherwise null.
>   possibleTypes: [__Type!]
>   # must be non-null for ENUM, otherwise null.
>   enumValues(includeDeprecated: Boolean! = false): [__EnumValue!]
>   # must be non-null for INPUT_OBJECT, otherwise null.
>   inputFields(includeDeprecated: Boolean! = false): [__InputValue!]
>   # must be non-null for NON_NULL and LIST, otherwise null.
>   ofType: __Type
>   # may be non-null for custom SCALAR, otherwise null.
>   specifiedByURL: String
> }
> ```
> 
> > Since Absinthe has always included deprecated fields for args, input_fields, and field, I recommend setting the default value of include_deprecated to true for these categories to maintain backwards compatibility.
> 
> If the trade-off is backwards compatibility vs spec-compliance, I would chose spec compliance, but I'm happy to reconsider if other maintainers feel differently cc: (@absinthe-graphql/core)
> 
> I'd like to understand how this is preventing y'all from upgrading. Maybe we can find an alternative fix?

It's mostly because it forces us to go to all queries and `includeDeprecated`. The more queries you have, the more you have to update and do QA regression etc. 

I understand your point though, that it'd be better to be spec-compliant. I wonder if we could make `include_deprecated` a configuration which defaults to false. We could then set it on our end to true and slowly migrate our queries. 🤔 

What do you think?

filipemonteiroth:
> > @filipemonteiroth The problem is that the spec requires that the default be false.
> > From [the spec](https://github.com/graphql/graphql-spec/blob/main/spec/Section%204%20--%20Introspection.md):
> > ```graphql
> > type __Type {
> >   kind: __TypeKind!
> >   name: String
> >   description: String
> >   # must be non-null for OBJECT and INTERFACE, otherwise null.
> >   fields(includeDeprecated: Boolean! = false): [__Field!]
> >   # must be non-null for OBJECT and INTERFACE, otherwise null.
> >   interfaces: [__Type!]
> >   # must be non-null for INTERFACE and UNION, otherwise null.
> >   possibleTypes: [__Type!]
> >   # must be non-null for ENUM, otherwise null.
> >   enumValues(includeDeprecated: Boolean! = false): [__EnumValue!]
> >   # must be non-null for INPUT_OBJECT, otherwise null.
> >   inputFields(includeDeprecated: Boolean! = false): [__InputValue!]
> >   # must be non-null for NON_NULL and LIST, otherwise null.
> >   ofType: __Type
> >   # may be non-null for custom SCALAR, otherwise null.
> >   specifiedByURL: String
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > > Since Absinthe has always included deprecated fields for args, input_fields, and field, I recommend setting the default value of include_deprecated to true for these categories to maintain backwards compatibility.
> > 
> > 
> > If the trade-off is backwards compatibility vs spec-compliance, I would chose spec compliance, but I'm happy to reconsider if other maintainers feel differently cc: (@absinthe-graphql/core)
> > I'd like to understand how this is preventing y'all from upgrading. Maybe we can find an alternative fix?
> 
> It's mostly because it forces us to go to all queries and `includeDeprecated`. The more queries you have, the more you have to update and do QA regression etc.
> 
> I understand your point though, that it'd be better to be spec-compliant. I wonder if we could make `include_deprecated` a configuration which defaults to false. We could then set it on our end to true and slowly migrate our queries. 🤔
> 
> What do you think?

We would have something along those lines:

```
config :absinthe,
  include_deprecated: true # defaults to false though
```

filipemonteiroth:
> @filipemonteiroth The problem is that the spec requires that the default be false.
> 
> From [the spec](https://github.com/graphql/graphql-spec/blob/main/spec/Section%204%20--%20Introspection.md):
> 
> ```graphql
> type __Type {
>   kind: __TypeKind!
>   name: String
>   description: String
>   # must be non-null for OBJECT and INTERFACE, otherwise null.
>   fields(includeDeprecated: Boolean! = false): [__Field!]
>   # must be non-null for OBJECT and INTERFACE, otherwise null.
>   interfaces: [__Type!]
>   # must be non-null for INTERFACE and UNION, otherwise null.
>   possibleTypes: [__Type!]
>   # must be non-null for ENUM, otherwise null.
>   enumValues(includeDeprecated: Boolean! = false): [__EnumValue!]
>   # must be non-null for INPUT_OBJECT, otherwise null.
>   inputFields(includeDeprecated: Boolean! = false): [__InputValue!]
>   # must be non-null for NON_NULL and LIST, otherwise null.
>   ofType: __Type
>   # may be non-null for custom SCALAR, otherwise null.
>   specifiedByURL: String
> }
> ```
> 
> > Since Absinthe has always included deprecated fields for args, input_fields, and field, I recommend setting the default value of include_deprecated to true for these categories to maintain backwards compatibility.
> 
> If the trade-off is backwards compatibility vs spec-compliance, I would chose spec compliance, but I'm happy to reconsider if other maintainers feel differently cc: (@absinthe-graphql/core)
> 
> I'd like to understand how this is preventing y'all from upgrading. Maybe we can find an alternative fix?

I have changed my approach to add a config option instead of setting it back to true. I think this is safe to merge and would unblock us from upgrading to newer versions.

cc @benwilson512

guides/introspection.md
```diff
@@ -91,6 +91,43 @@ Getting the name of the fields for a named type:
...
+
+The introspection system can be configured through your application config:
+
+```elixir
+config :absinthe,
+  include_deprecated: false  # Defaults to false, set to true for backward compatibility
+```
+
+This configuration affects the default behavior of the `includeDeprecated` argument in introspection queries. When set to `false` (the default), deprecated fields and values will not be included in introspection results unless explicitly requested via the `includeDeprecated` argument. When set to `true`, deprecated fields and values will be included by default, matching the behavior of previous versions of Absinthe.
+
```
cschiewek, line 104: @filipemonteiroth can we add a sentence here indicating that setting this to true will also break compatibility?

cschiewek: Perfect ty

cschiewek (approved):

PR #1336, opened 2024-08-03
Unregsiter duplicate (listening to the same topic) subscriptions individually

### LINKED ISSUE #1335: With duplicate subscriptions unsubscribing of one of the subs kills both records in Registry
### Environment

* Elixir version (elixir -v): Elixir 1.16.0 (compiled with Erlang/OTP 26)
* Absinthe version (mix deps | grep absinthe): 1.7.6
* Client Framework and version (Relay, Apollo, etc): Apollo, version is probably irrelevant

### Expected behavior
We want to subscribe to the same subscription twice: one subscription is alive throughout the lifecycle of the app (counter for notifications), second is only alive when a component rendering the content of notifications is open.

### Actual behavior
When subscribing to the same subscription twice (with different set of fields), unsubscribing of one of the subs kills the second sub.
It happens due to unregistering by field_key, which is a total duplicate in this case: https://github.com/absinthe-graphql/absinthe/blob/5705d9a0e16bf425d29f5fc7a945bbff4d751ec5/lib/absinthe/subscription.ex#L179
I believe it should check the record and only remove references to doc_ids.

youroff:
This PR fixes #1335. Having two subscriptions listening to the same topic, unsubscribe on one subscription would remove both references from the registry disrupting messages heading for the second subscription. Thе fix is to use `unregister_match` with correspondent doc_id instead of plain `unregister` targening field_key only.

### DIFF
--- a/lib/absinthe/subscription.ex
+++ b/lib/absinthe/subscription.ex
@@ -176,7 +176,7 @@ defmodule Absinthe.Subscription do
     registry = pubsub |> registry_name
 
     for field_key <- pdict_fields(doc_id) do
-      Registry.unregister(registry, field_key)
+      Registry.unregister_match(registry, field_key, doc_id)
     end
 
     Registry.unregister(registry, doc_id)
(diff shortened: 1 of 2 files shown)


benwilson512 (approved):
Fantastic, thanks!

Issue #1337, opened 2024-09-01
Absinthe.Resolution.path_string/1 crashes if path includes a list


mdg:
I have a patch for this one ready. Just submitting the issue for referencing from the PR.

### Environment

* Elixir version (elixir -v): 1.16.2
* Absinthe version (mix deps | grep absinthe): 1.7.8
* Client Framework and version (Relay, Apollo, etc): n/a

### Expected behavior

Expected [Absinthe.Resolution.path_string/1](https://hexdocs.pm/absinthe/Absinthe.Resolution.html#path_string/1) to always return a list of strings and not raise a FunctionClauseError

### Actual behavior

If the query path contains a list, the path value list contains an integer for the node in the path that is a list item. The [fn passed to the map matches on maps](https://github.com/absinthe-graphql/absinthe/blob/3d0823bd71c2ebb94357a5588c723e053de8c66a/lib/absinthe/resolution.ex#L239) and a FunctionClauseError is raised when it gets to the integer. case statement fails with a FunctionClause

### Relevant Schema/Middleware Code

Schema has to include a list upstream from the field being resolved when path_string is called.
https://github.com/absinthe-graphql/absinthe/blob/3d0823bd71c2ebb94357a5588c723e053de8c66a/lib/absinthe/resolution.ex#L238-L246

mdg:
This function doesn't actually have any documentation so it's not entirely clear how it's different from `path/1`. In attempting to add the documentation, it's unclear how to describe what it does.

PR #1338, opened 2024-09-01
fix path_string crash when path includes list

mdg:
in Absinthe.Resolution; also:

* add documentation for path_string/1
* add specs for path/1 and path_string/1

this fix is #1337

### DIFF
--- a/lib/absinthe/resolution.ex
+++ b/lib/absinthe/resolution.ex
@@ -128,7 +128,7 @@ defmodule Absinthe.Resolution do
   In this case `5` is the 0 based index in the list of users the field is currently
   at.
   """
-  @spec path(t()) :: [String.t()]
+  @spec path(t()) :: [String.t() | integer()]
   def path(%{path: path}) do
     path
     |> Enum.reverse()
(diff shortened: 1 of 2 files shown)


lib/absinthe/resolution.ex
```diff
@@ -128,6 +128,7 @@ defmodule Absinthe.Resolution do
   In this case `5` is the 0 based index in the list of users the field is currently
   at.
   """
+  @spec path(t()) :: [String.t()]
   def path(%{path: path}) do
```
benwilson512, line 132: I don't think this spec is correct. This returns the query nodes, not strings. Also, in the case of lists, it returns integers.

lib/absinthe/resolution.ex
```diff
@@ -235,8 +236,30 @@ defmodule Absinthe.Resolution do
...
+  ```graphql
+  query {users { email }}
+  ```
+
+  If you called this function inside a resolver on the users email field it
+  returns a value like:
+
+  ```elixir
+  resolve fn _, _, resolution ->
+    Absinthe.Resolution.path(resolution) #=> ["email", "users", "RootQueryType"]
```
benwilson512, line 254: The desired output here is probably to stringify the integer so that the information isn't lost. This is also a breaking change so we'll need to release this under a new minor version.

benwilson512 (changes requested):
I think the correct

lib/absinthe/resolution.ex
mdg: I'll switch to `[String.t() | integer()]`

lib/absinthe/resolution.ex
mdg: I've added back the integers as strings. I would suggest this isn't a breaking change, b/c no one was using the previous behavior b/c it was broken.

mdg:
I've tried to address the feedback

Issue #1340, opened 2024-09-26
Allow subscription config errors to return spec compliant errors


bryanjos:
Currently subscription config functions expect either `{:ok, config}` or `{:error, msg}` to be returned. In the error case, this is what's returned ultimately:

```elixir
%{errors: [%{message: msg}]}
```

We would like to send spec compliant errors with extra information from the config errors. 

Right now if we return the following from config

```elixir
{:error, %{extensions: %{code: "FAILED"}, message: "failed"}}
```

We get

```elixir
%{errors: [%{message: %{extensions: %{code: "FAILED"}, message: "failed"}}]}
```

Which breaks some tools that expect spec compliant errors.

There are a couple of ways we've thought about fixing it and could use some feedback before making a PR.

The first way would be to update `Absinthe.Phase.Document.Result` to look for a map in the `message` property of the `Absinthe.Phase.Error` struct created and return what's there as the error.

So returning `{:error, %{extensions: %{code: "FAILED"}, message: "failed"}` from the config function would give

```elixir
%{errors: [%{extensions: %{code: "FAILED"}, message: "failed"}]}
```

The second way we were thinking is updating `Absinthe.Phase.Subscription.SubscribeSelf` to let config return a 3-tuple, `{:error, msg, extra}` and put the `extra` there onto the `extra` property of the `Absinthe.Phase.Error` struct created.

With that, and with `spec_compliant_errors` set to true, if `{:error, "failed", %{code: "FAILED"}` is returned, you would get the above as well.

Or some other way we aren't thinking of.

michaelcaterisano:
Thanks @bryanjos! We've already discussed, but I'll add my thoughts here as well.

My take is that the `config/2` function's API should match the resolver API. That is, it should support the same kind of error tuples. Currently, resolvers allow returning `{:error, "message"}` or `{:error, map}`, where map is `%{message: "message", extensions: %{}}`. https://github.com/absinthe-graphql/absinthe/commit/32db847a137834e618d6f66751b1fa3bcdc4a5fe introduced a `spec_compliant_errors` option, which, when set to true, creates a map like the one above, with "extra errors" being added to the extensions key. I think the idea is for Absinthe to help ensure that errors are spec compliant, but as the PR mentions, the change needs to be opt-in because it's breaking. I also wasn't able to find this option in the docs, but maybe I missed it.

[...]

PR #1341, opened 2024-09-27
Allow config/2 to send errors in spec compliant format

bryanjos:
Related to #1340

This PR makes it so that if `config/2` returns an error tuple where the second element is a map, then that map is treated as the error response.

Subscription config errors currently expect a tuple of `{:error, msg }`. If message is an map, when returned to the user, it looks like:

```json
"errors": [
    {"message": {"extensions": {"code": "BAD_REQUEST"}, "message": "Invalid argument"}}
]
```

This causes trouble for us as we return spec compliant errors and they get wrapped in the message.

This PR makes it so returning the following from `config/2`

```elixir
{:error, %{message: "Invalid argument", extensions: %{code: "BAD_REQUEST"}}
```

Will return 

```json
"errors": [
    {"extensions": {"code": "BAD_REQUEST"}, "message": "Invalid argument"}
]
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Feature: [Treat maps returned from config/2 as response](https://github.com/absinthe-graphql/absinthe/pull/1341)
 - Feature: [Add async option to Absinthe.Subscription](https://github.com/absinthe-graphql/absinthe/pull/1329)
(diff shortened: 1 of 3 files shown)


michaelcaterisano (approved):
Seems reasonable to me!

benwilson512:
I'm perfectly happy to let ya'll decide and move forward on this. I would recommend adding a test case after this to `absinthe_phoenix` to make sure this actually works end to end.

bryanjos:
I wrote a test case for absinthe_phoenix locally and verified everything passes. I can make a PR there

PR #1345, opened 2024-10-28
Include deprecated input_fields for absinthe.schema.json

aaronrenner:
This is an attempt to solve https://github.com/absinthe-graphql/absinthe/issues/1327, where the deprecated input fields are no longer included in the `mix absinthe.schema.json` file. I added test coverage to ensure these test fields are actually being included.


cschiewek:
This was fixed in https://github.com/absinthe-graphql/absinthe/pull/1333.  The default is still to exclude them, to ensure introspection is spec compliant, but you can opt-in to the older non-compliant behaviour via config: https://hexdocs.pm/absinthe/introspection.html#configuration

PR #1349, opened 2024-12-08
Make task implementation for async configurable

drtheuns:
Projects that don't use open-telemetry or with more complex process propagation requirements can now implement it themselves by configuring the `Task` module to use.

### DIFF
--- a/guides/telemetry.md
+++ b/guides/telemetry.md
@@ -89,3 +89,11 @@ Instead, you can add the `:opentelemetry_process_propagator` package to your
 dependencies, which has a `Task.async/1` wrapper that will attach the context
 automatically. If the package is installed, the middleware will use it in place
 of the default `Task.async/1`.
+
+Alternatively, you can configure the `Task` implementation to use:
+
+```elixir
+config :absinthe, task_module: MyTaskModule
+```
+
+The provided module must support the `async/1` function.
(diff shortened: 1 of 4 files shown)


lib/absinthe/utils.ex
```diff
@@ -151,4 +151,24 @@ defmodule Absinthe.Utils do
     #{types ++ directives}
     """
   end
+
+  @spec async((-> any)) :: Task.t()
+  def async(fun) do
+    case Application.fetch_env(:absinthe, :task_module) do
```
benwilson512, line 157: This should be a compile env property, not something we swap out at runtime.

benwilson512 (changes requested):
I think this is a fine idea but we definitely want this to be compile time.

guides/telemetry.md
```diff
@@ -89,3 +89,11 @@ Instead, you can add the `:opentelemetry_process_propagator` package to your
 dependencies, which has a `Task.async/1` wrapper that will attach the context
 automatically. If the package is installed, the middleware will use it in place
 of the default `Task.async/1`.
+
+Alternatively, you can configure the `Task` implementation to use:
+
+```elixir
+config :absinthe, task_module: MyTaskModule
+```
```
benwilson512, line 97: We should make a comment here that this is a compile time option and thus needs to be specified in one of the compile time config files like `config.exs` or `prod.exs` and not `runtime.exs`

PR #1360, opened 2025-04-24
Replace regex with pattern as string to support OTP 28

TBK145:
OTP 28 changes the implementation of the regex module, and as a result you can no longer use regexes as module attributes.

By changing the regex to a pattern as a string it can still be used in the error message, and just needs to be compiled before usage.

### DIFF
--- a/lib/absinthe/phase/schema/validation/names_must_be_valid.ex
+++ b/lib/absinthe/phase/schema/validation/names_must_be_valid.ex
@@ -4,7 +4,6 @@ defmodule Absinthe.Phase.Schema.Validation.NamesMustBeValid do
   use Absinthe.Phase
   alias Absinthe.Blueprint
 
-  @valid_name_regex ~r/^[_A-Za-z][_0-9A-Za-z]*$/
 
   def run(bp, _) do
     bp = Blueprint.prewalk(bp, &validate_names/1)
@@ -30,7 +29,7 @@ defmodule Absinthe.Phase.Schema.Validation.NamesMustBeValid do
   end
 
   defp valid_name?(name) do
-    Regex.match?(@valid_name_regex, name)
+    Regex.match?(valid_name_regex(), name)
   end
 
   defp error(object, data) do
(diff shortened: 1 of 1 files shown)


lib/absinthe/phase/schema/validation/names_must_be_valid.ex
```diff
@@ -4,7 +4,7 @@ defmodule Absinthe.Phase.Schema.Validation.NamesMustBeValid do
   use Absinthe.Phase
   alias Absinthe.Blueprint
 
-  @valid_name_regex ~r/^[_A-Za-z][_0-9A-Za-z]*$/
+  @valid_name_regex "^[_A-Za-z][_0-9A-Za-z]*$"
```
josevalim, line 8:
```suggestion
```

josevalim:
This is a change on Erlang/OTP 28, we can no longer precompile them in the module body, but we only need to move it inside. Elixir v1.19 will emit a warning telling you to do exactly that.

TBK145:
Hi @josevalim , at first I wanted to solve it in the way you propose, but the regex is also referenced [here](https://github.com/absinthe-graphql/absinthe/pull/1360/files#diff-ecc121586489de928a255795e682021839fce270da5fcbe3f515a03341e23e5eR48), so I did it in this way to only have to maintain the regex in one place.
But I understand it's better to maintain the regex twice in favour of performance, thanks @benwilson512 I did not think about that.

josevalim:
In this case it is best to move the regex to a private function and reuse it. :)

TBK145:
That was actually the first thing I tried, but since it's a module attribute, the function is not defined yet before calling.
I could move the whole `@description` block into the function, but I don't know if that has any negative side-effects.

josevalim:
What I mean is this:

```
defp valid_regex_name, do: ~r/foobar/
```

And use it as `valid_regex_name()`.

TBK145:
I understand, but what I mean is that the regex is also used in module attribute `@description`, which cannot call a function from that same module.

josevalim:
You can also move it to a function. I assume there isn't much benefit for those being attributes anyway, besides user convenience.

benwilson512:
Thank you @josevalim for the assist!

Issue #1351, opened 2025-01-23
Pipeline module not compiled at schema compilation time


jdewar:
I ran into the error described here: https://github.com/absinthe-graphql/absinthe/issues/1331#issuecomment-2251382907 sometime ago and used the unreleased commit (`ref: 3d0823bd71c2ebb94357a5588c723e053de8c66a`) to fix my compilation problem, but the problem has returned.  

```
== Compilation error in file lib/sga/frontend_schema.ex ==
** (ArgumentError) could not load module SGA.Pipeline.PublicFieldVisibility due to reason :nofile
    (elixir 1.17.2) lib/code.ex:1837: Code.ensure_loaded!/1
    (absinthe 1.7.8) lib/absinthe/schema.ex:382: anonymous fn/3 in Absinthe.Schema.apply_modifiers/3
    (elixir 1.17.2) lib/enum.ex:2531: Enum."-reduce/3-lists^foldl/2-0-"/3
    (absinthe 1.7.8) lib/absinthe/schema.ex:405: Absinthe.Schema.__after_compile__/2
    (stdlib 5.2) lists.erl:1594: :lists.foldl/3
```

Removing `_build` fixes it at the time, but it has recurred 3 times across 2 different developers in the past few weeks.  Here is the pipeline definition and how it is used, in case I'm 'holding it wrong':

----

public_field_visibility.ex:
```elixir
defmodule SGA.Pipeline.PublicFieldVisibility do
  alias Absinthe.{Blueprint, Phase, Pipeline}

  # Add this module to the pipeline of phases
  # to run on the schema
  def pipeline(pipeline) do
    Pipeline.insert_after(pipeline, Phase.Schema.TypeImports, __MODULE__)
  end

  # Here's the blueprint of the schema, let's do whatever we want with it.
  def run(%Blueprint{} = blueprint, _) do
    blueprint =
      update_in(
        blueprint,
        [
          Access.key!(:schema_definitions),
          Access.all(),
          Access.key!(:type_definitions),
          Access.all()
        ],
        &update_type_def/1
      )

    {:ok, blueprint}
  end

  defp update_type_def(%Absinthe.Type.Object{fields: type_fields} = type_def) do
    public_fields = Absinthe.Type.meta(type_def, :public_fields) || []
    public_fields = [:id | public_fields]
    filtered_fields = filter_fields(public_fields, type_fields)

    %{type_def | fields: filtered_fields}
  end

[...]

PR #1361, opened 2025-04-28
fix: use `Code.ensure_compiled!` at compile time

zachdaniel:
While investigating the changes in this PR:
https://github.com/elixir-lang/elixir/pull/14451

I encountered an error about not being able to load a module at this line, implying this is called at compile time.

@josevalim pointed out that `Code.ensure_loaded!/1` will have no effect at compile time, and so this was really just relying on compile order and would be non-deterministic. `Code.ensure_compiled!/1` is a superset of `Code.ensure_loaded!/1` and will work at compile time to ensure the depended on module is compiled.

This would appear to solve #1351.


zachdaniel:
It should be noted that this error may affect people more significantly in the next version of Elixir so we may want to get this merged sooner rather than later 😄

josevalim (approved):

PR #1372, opened 2025-06-24
Updates for Elixir 1.19

cschiewek:
This PR makes the necessary changes to get Absinthe working with Elixir 1.19

Changes include:
- Update mix.exs 
- Updating CI to test against 1.19
- Removing a `Absinthe.Language.Render.do_block/2` function head that was never being called and was raising a dialyzer error.
- Updating a test to make the formatter happy across all versions

### DIFF
--- a/lib/absinthe/language/render.ex
+++ b/lib/absinthe/language/render.ex
@@ -416,10 +416,6 @@ defmodule Absinthe.Language.Render do
     do_block(docs)
   end
 
-  defp block(:doc_nil, docs) do
-    do_block(docs)
-  end
-
   defp block(name, docs) do
     glue(
       name,


.github/workflows/elixir.yml
```diff
@@ -18,13 +18,16 @@ jobs:
...
+          - "1.19"
         otp:
           - "25"
           - "26"
           - "27"
           - "28"
         # see https://hexdocs.pm/elixir/compatibility-and-deprecations.html#between-elixir-and-erlang-otp
         exclude:
+          - elixir: 1.19
+            otp: 25
```
binaryseed, line 30: Maybe this should be 28?

binaryseed (changes requested):

.github/workflows/elixir.yml
cschiewek: This is the exclusion list.  

Why would be want to exclue 1.19 on OTP 28 when it's support from 26-28? https://hexdocs.pm/elixir/1.19.0-rc.0/compatibility-and-deprecations.html#between-elixir-and-erlang-otp

binaryseed: Oh! Sorry, I just misread that

binaryseed (approved):

PR #1373, opened 2025-07-02
Add field description inheritance from referenced types during introspection

jwaldrip:
## Summary

This PR enhances GraphQL introspection by making fields inherit descriptions from their referenced types when no explicit description is provided. This improves API documentation by automatically propagating type descriptions to fields that reference those types.

## Motivation

Currently, when a field has no description, introspection returns `null` even if the field references a well-documented type. This leads to incomplete API documentation where developers must manually duplicate type descriptions on every field that references that type.

### Example

```elixir
object :user do
  description "A user in the system"
  field :id, :id
  field :name, :string
end

object :post do
  field :author, :user  # Currently returns null description in introspection
end
```

With this change, the `author` field will automatically inherit the description "A user in the system" during introspection.

## Implementation Details

- Modified the `__field` introspection resolver to check if a field has no description and attempt to use the referenced type's description as a fallback
- Properly handles wrapped types (`non_null`, `list_of`) by unwrapping them first before looking up the base type
- Works with custom types, built-in scalars, interfaces, and all GraphQL type constructs

## Testing

Added comprehensive test coverage in `test/absinthe/introspection/field_description_inheritance_test.exs` including:
- Direct type references
- Wrapped types (non_null, list_of)
- Built-in scalar types
- Interface types
- Query fields
- Nested type references

All existing tests continue to pass.

## Documentation

Updated the field documentation in `lib/absinthe/type/field.ex` to explain the new behavior.

## Breaking Changes

None. This is a backwards-compatible enhancement that only affects fields without explicit descriptions.

🤖 Generated with [Claude Code](https://claude.ai/code)


jwaldrip:
@benwilson512 the motivator here was to create better documentation overall. I was also fixing some other bugs along the way with SDL generation related to directives. I will update the PR to add a :inherit option.

Issue #1359, opened 2025-04-07
Using `import_types` introduces transitive compile dependencies on resolvers


jesenko:
Scenario: split large schema module to multiple smaller submodules, use `import_types` and `import_fields` to combine them.

This introduces compile-time dependency of combined schema module on all submodules from which we import fields and types. Submodules however have runtime dependencies on various resolvers (which themselves have runtime dependencies all over the place). All there runtime dependencies become transitive compile-time dependencies of schema module, resulting in recompilation of large number of files on each change in any transitive compile-time dependency.

This blogs sums it quite nicely and I guess is still relevant with current version of absinthe: https://devonestes.com/trying-to-speed-up-compilation-with-absinthe

Not sure what could be done at absinthe level with respect to this issue... I guess we could obscure resolver calls so that compiler does not detect them as runtime dependencies, but not sure if that would introduce other issues. At least, there should probably be an "official" way on how to avoid described recompilation issue?

PR #1380, opened 2025-09-25
feat: Support module concat in middleware to avoid compile-time dependencies

axelson:
This allows using `Module.concat/1,2` to provide the middleware. The benefit of this is that the resolver no longer has a compile-time dependency on the middleware, and more importantly a transitive dependency on all the middleware's runtime dependencies.

Here is an example of a supported field definition:
```elixir
field :folder, non_null(:folder) do
  arg :id, non_null(:string)
  middleware Module.concat(MyAppWeb.GraphQL, MyMiddleware), some_options: [42]
  resolve &SomeResolver.project_folder/3
end
```

This almost works today, but `MyMiddleware` doesn't receive the provided options.

Fix is related to https://github.com/absinthe-graphql/absinthe/issues/1359

This is more of a PoC/RFC right now. We may actually want to go with an approach that targets the compiler directly, like these two related ecto commits:
- https://github.com/elixir-ecto/ecto/pull/1670
- https://github.com/elixir-ecto/ecto/commit/cadaced80a4fe200f21003da111ce801b48d1c07

### DIFF
--- a/lib/absinthe/schema/notation.ex
+++ b/lib/absinthe/schema/notation.ex
@@ -1873,6 +1873,10 @@ defmodule Absinthe.Schema.Notation do
               {:{}, [], [{env.module, atom}, opts]}
           end
 
+        # Special handling for Module.concat to allow breaking of compilation cycles
+        {{:., _meta, [Module, :concat]}, _call_meta, _args} = val ->
+          {:{}, [], [{val, :call}, opts]}
+
         val ->
           val
       end


s3cur3:
I am _extremely_ interested in this fix for @Jump-App's codebase. Right now the compile-time dependencies via our Absinthe schema means any change to what is conceptually a leaf module results in 30+ modules being recompiled. If there's anything I can do to help, let me know.

PR #1383, opened 2025-10-06
allow dirctives on nameless queries

bsanyi:
Accordig to the [GraphQL spec](https://spec.graphql.org/September2025/#sec-Language.Operations) it is valid to query (or mutate, or subscribe) without name, but with directives, like
    
    query @cache(ttl: 500) { someQuery }
    
Absinthe currently does not parse a qurey like the above, except with a name included, like this:

    query name @cache(ttl: 500) { someQuery }

The fix is a little modification of the yecc grammer in this PR.


benwilson512 (changes requested):
Hey @bsanyi this is great thanks. Can you please add a test case?

benwilson512 (approved):

Issue #1083, opened 2021-06-18
`mix clean` is sometimes required to see changes take effect


bmidgley:
### Environment

* Elixir version (elixir -v): Erlang/OTP 23 [erts-11.2] [source] [64-bit] [smp:12:12] [ds:12:12:10] [async-threads:1] [hipe]
* Absinthe version (mix deps | grep absinthe): absinthe 1.6.4 / absinthe_plug 1.5.8
* Client Framework and version (Relay, Apollo, etc): graphiql

### Expected behavior

Changing my types file from
field(:id, :id)
to
field :id, :string do
     resolve(fn object, _, _ ->
        {:ok, GlobalId.encode(Spiff.Comment, object.id)}
      end)
    end

Sometimes does not register as a change

### Actual behavior

id continued to behave as a plain db column until I ran `mix clean` and restarted. Then it works as expected. Restarting iex on its own is not enough to break the stale cache.

kyleboe:
I believe this is related to https://github.com/absinthe-graphql/absinthe/issues/1139. I have to do the same thing.

PR #1402, opened 2025-12-23
fix: render default values in SDL output for schema.sdl task

### LINKED ISSUE #1222: mix absinthe.schema.sdl does not include `default_values`
### Environment

* Elixir version (elixir -v): Elixir 1.13.4 (compiled with Erlang/OTP 24)
* Absinthe version (mix deps | grep absinthe): master

### Expected behavior

When generating SDL via `mix absinthe.schema.sdl --schema ...` Absinthe does not include any fields or input args etc. with the `default_value` keyword. With the test Absinthe schema below, the task outputs:

Task output:
```graphql
schema {
  mutation: RootMutationType
  query: RootQueryType
}

type RootMutationType {
  createItem(input: TestInput!, other: String = "other"): String
}

type RootQueryType {
  item(show: Boolean = true): String
}

input TestInput {
  value: String! = "test"
}
```

Test absinthe schema:
```elixir
defmodule MyTestSchema do
  use Absinthe.Schema

  query do
    field :item, :string do
      arg :show, :boolean, default_value: true
    end
  end

  mutation do
    field :create_item, :string do
      arg :input, non_null(:test_input)
      arg :other, :string, default_value: "other"
    end
  end

  input_object :test_input do
    field :value, non_null(:string), default_value: "test"
  end
end
```

> The `absinthe.schema.json` does however have the default values in it

### Actual behavior

The generator returns
```graphql
schema {
  mutation: RootMutationType
  query: RootQueryType
}

input TestInput {
  value: String!
}

type RootMutationType {
  createItem(input: TestInput!, other: String): String
}

type RootQueryType {
  item(show: Boolean): String
}
```

Test schema I used
```
defmodule MyTestSchema do
  use Absinthe.Schema

  query do
    field :item, :string do
      arg :show, :boolean, default_value: true
    end
  end

  mutation do
    field :create_item, :string do
      arg :input, non_null(:test_input)
      arg :other, :string, default_value: "other"
    end
  end

  input_object :test_input do
    field :value, non_null(:string), default_value: "test"
  end
end
```

This test I wrote for `test/mix/tasks/absinthe.schema.sdl_test.exs` fails but I would expect the strings to all be present.
```
test "with a default input" do
  argv = ["ouput.graphql", "--schema", "#{MyTestSchema}"]
  opts = Task.parse_options(argv)

  {:ok, schema} = Task.generate_schema(opts)
  assert schema =~ "value: String! = \"test\""
  assert schema =~ "other: String = \"other\""
  assert schema =~ "show: Boolean = true"
end
```

jruhland:
Fixes #1222

Previously, the `mix absinthe.schema.sdl` task would not include default values for arguments and input object fields in the generated SDL output, even though the `mix absinthe.schema.json` task would properly include them in its output.

Changes:
- Modified SDL renderer to handle default values for InputValueDefinition (arguments) and FieldDefinition (input object fields)
- Added support for DSL-defined schemas where default_value is an Elixir term that needs conversion to Blueprint.Input structures
- Added support for SDL-parsed schemas where default_value_blueprint is already a Blueprint structure
- Skip non-serializable values (refs, functions)

Tests:
- Added DefaultValueTestSchema with default value coverage to prevent regression while rendering boolean, string, and input object default values
- All existing tests updated to expect default values in output. There were 2 that confirmed this behavior wasn't working
- We integrated this change into our production app, and have validated the expected outcomes against a significantly large schema that also employs the federation extensions.

Note:

Hit #1083 quite a bit, will see if I can address in a separate PR.

### DIFF
--- a/lib/absinthe/schema/notation/sdl_render.ex
+++ b/lib/absinthe/schema/notation/sdl_render.ex
@@ -324,12 +324,8 @@ defmodule Absinthe.Schema.Notation.SDL.Render do
         concat([" = ", string(enum_name)])
 
       :error ->
-        try do
-          blueprint_value = Blueprint.Input.parse(default_value)
-          concat([" = ", render_value(blueprint_value)])
-        rescue
-          _ -> empty()
-        end
+        blueprint_value = Blueprint.Input.parse(default_value)
+        concat([" = ", render_value(blueprint_value)])
     end
   end
 
@@ -339,12 +335,8 @@ defmodule Absinthe.Schema.Notation.SDL.Render do
        )
        when is_binary(default_value) or is_number(default_value) or is_boolean(default_value) or
               is_list(default_value) or is_map(default_value) do
-    try do
-      blueprint_value = Blueprint.Input.parse(default_value)
-      concat([" = ", render_value(blueprint_value)])
-    rescue
-      _ -> empty()
-    end
+    blueprint_value = Blueprint.Input.parse(default_value)
+    concat([" = ", render_value(blueprint_value)])
   end
 
   defp default(%Blueprint.Schema.FieldDefinition{}, _type_definitions) do
--- a/lib/absinthe/schema/notation/sdl_render.ex
+++ b/lib/absinthe/schema/notation/sdl_render.ex
@@ -70,7 +70,7 @@ defmodule Absinthe.Schema.Notation.SDL.Render do
       string(@adapter.to_external_name(input_value.name, :argument)),
       ": ",
       render(input_value.type, type_definitions),
-      default(input_value.default_value_blueprint),
+      default(input_value, type_definitions),
       directives(input_value.directives, type_definitions)
     ])
     |> description(input_value.description)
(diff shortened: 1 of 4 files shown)


lib/absinthe/schema/notation/sdl_render.ex
```diff
@@ -306,12 +306,54 @@ defmodule Absinthe.Schema.Notation.SDL.Render do
...
+
+  defp default(%Blueprint.Schema.FieldDefinition{default_value: default_value})
+       when is_atom(default_value) or is_binary(default_value) or is_number(default_value) or
+              is_boolean(default_value) or is_list(default_value) or is_map(default_value) do
+    try do
+      blueprint_value = Blueprint.Input.parse(default_value)
+      concat([" = ", render_value(blueprint_value)])
+    rescue
+      _ -> empty()
+    end
```
binaryseed: What led to using `rescue` here? Ideally we don't use exceptions for control flow... This could swallow actual errors

jruhland: sorry, just missed cleaning these up while debugging, let me polish this up

PR #1412, opened 2026-02-10
fix: leaked async tasks in Incremental.Supervisor

QuinnWilton:
Note that this issue was found while running a static analysis tool I'm working on. It seems like a legitimate finding, but it wasn't an issue I ran into while using `Absinthe`.

`start_deferred_task/1` and `start_stream_task/1` used `async_nolink` without awaiting, leaking a result message and `:DOWN` message into the caller's mailbox per task. Replace with `start_child`, which is true fire-and-forget.


cschiewek:
@QuinnWilton any chance you can update your branch to resolve the conflicts?  Then I can review and hopefully merge.  Thanks!

PR #1416, opened 2026-03-19
feat: PartitionSupervisor for subscription tasks

jhchen:
Task.Supervisor.start_child can become a bottleneck in large systems publishing many subscriptions since the task supervisor is a single process. [PartitionSupervisor](https://hexdocs.pm/elixir/main/PartitionSupervisor.html) can be used to easily resolves this by partitioning as of Elixir 1.14.0. This change is essentially their DynamicSupervisor example.


cschiewek (approved):
Thank you!

PR #1417, opened 2026-04-03
feat: add DigitAwareLanguageConventions adapter for digit-boundary roundtrip

### LINKED ISSUE #298: Capitalized field names yields bad error message
Defining a field with a capitalized name does not work, but the error message still suggests using the capitalized name when attempting to query it.

```
query do
    field :Foo, type: :string, resolve: fn
      _, _ -> {:ok, "Bar"}
    end
end
```

When running the query `query {Foo}` this returns:
```
{
  "errors": [
    {
      "message": "Cannot query field \"Foo\" on type \"RootQueryType\". Did you mean \"Foo\"?",
      "locations": [
        {
          "line": 1,
          "column": 0
        }
      ]
    }
  ]
}
```

### LINKED ISSUE #312: Fields with numbers in the name not working
I have a `work_phone_2` field in my database and in my Absinthe types:

```elixir
object :contact do
  field :id, :id
  ...
  field :work_phone_2, :string
end
```

If I run a query against this:
```gql
query Contact {
  contact(id: 1) {
    id
    workPhone2
  }
}
```

I get an error saying:

    Cannot query field "workPhone2" on type "Contact". Did you mean "workPhone2"?

If I change it to snake case (`work_phone_2`) in my query it works fine (`workPhone_2` in the query works too). It appears it only fails when a number immediately follows an underscore. For example, if I set up `field :work_phone_a2` then querying against `workPhoneA2` works fine.

### LINKED ISSUE #367: field names ending with _ number
I have a set of fields like:

```
      logo_url
      logo_url_1
```

When entering a query into graphiql it says that logo_url_1 should be queried as logoUrl1 but this generates an error. I can query with logo_url_1 or logoUrl_1 but not logoUrl1 as graphiql suggests. I can work around this by having folks using the API use logo_url_1 so it isn't critical. Just thought I'd mention it.

donleandro:
Hey! Rewrote this based on @benwilson512's feedback. Instead of modifying LanguageConventions, this adds a new opt-in adapter: `Absinthe.Adapter.DigitAwareLanguageConventions`.

## The problem

`Macro.underscore/1` doesn't insert underscores at letter-to-digit boundaries. So a field like `:requires_2fa` gets camelized to `"requires2fa"`, but converting back gives `"requires2fa"` instead of `"requires_2fa"`. The roundtrip breaks and the field becomes unreachable from GraphQL.

This affects any field with an underscore before a digit: `:work_phone_2`, `:address_line_1`, `:field_2_name`, etc.

## The fix

A new adapter that behaves identically to `LanguageConventions` for `to_external_name/2`, but after `Macro.underscore/1` in `to_internal_name/2`, inserts underscores at `[a-z][0-9]` boundaries via a single `String.replace/3` call.

```elixir
# In your Phoenix endpoint or Absinthe config:
forward "/api",
  to: Absinthe.Plug,
  init_opts: [schema: MyAppWeb.Schema, adapter: Absinthe.Adapter.DigitAwareLanguageConventions]
```

```elixir
Adapter.to_internal_name("requires2fa", :field)   #=> "requires_2fa"
Adapter.to_internal_name("workPhone2", :field)     #=> "work_phone_2"
Adapter.to_internal_name("field2Name", :field)     #=> "field_2_name"
Adapter.to_internal_name("addressLine1", :field)   #=> "address_line_1"
```

## What changed from the previous version

- No longer modifies LanguageConventions - zero breaking change risk
- New opt-in adapter that users choose explicitly
- Simpler implementation: `Macro.underscore` + one `String.replace` for digit boundaries
- Full test coverage including roundtrip verification

## Trade-off

This adapter always inserts underscores at letter-digit boundaries. If a schema has a field like `:req1` (digit without preceding underscore), this adapter would convert it to `req_1` on input, which wouldn't match. The moduledoc documents this clearly. In practice, underscore-before-digit (`:requires_2fa`) is far more common in Elixir than digit-glued-to-word (`:req1`).

We hit this originally building [Shiko](https://shiko.vet) (veterinary SaaS in Elixir) where `:requires_2fa` on the login response was completely unreachable from GraphQL.

Closes #312, closes #298, closes #367.

### DIFF
--- a/lib/absinthe/adapter.ex
+++ b/lib/absinthe/adapter.ex
@@ -8,7 +8,7 @@ defmodule Absinthe.Adapter do
   Adapters aren't a part of GraphQL, but a utility that Absinthe adds so that
   both client and server can use use conventions most natural to them.
 
-  Absinthe ships with four adapters:
+  Absinthe ships with five adapters:
 
   * `Absinthe.Adapter.LanguageConventions`, which expects schemas to be defined
     in `snake_case` (the standard Elixir convention), translating to/from `camelCase`
@@ -24,6 +24,9 @@ defmodule Absinthe.Adapter do
   * `Absinthe.Adapter.StrictLanguageConventions`, which expects schemas to be
     defined in `snake_case`, translating to `camelCase` for outgoing results.
     This adapter requires incoming query documents to use `camelCase`.
+  * `Absinthe.Adapter.DigitAwareLanguageConventions`, which extends the behavior
+    of `Absinthe.Adapter.LanguageConventions` with correct roundtrip handling of
+    field names containing digits (e.g., `requires_2fa`, `address_line_1`).
 
   To set an adapter, you pass a configuration option at runtime:
 
--- a/lib/absinthe/adapter/digit_aware_language_conventions.ex
+++ b/lib/absinthe/adapter/digit_aware_language_conventions.ex
@@ -1,0 +1,88 @@
+defmodule Absinthe.Adapter.DigitAwareLanguageConventions do
+  use Absinthe.Adapter
+  alias Absinthe.Utils
+
+  @moduledoc """
+  An adapter that extends the behavior of `Absinthe.Adapter.LanguageConventions`
+  with correct roundtrip handling of field names containing digits.
+
+  ## The problem with `LanguageConventions`
+
+  `Macro.underscore/1` does not insert underscores at letter-to-digit boundaries.
+  This means that names like `requires_2fa` get camelized to `"requires2fa"`, but
+  converting back with `Macro.underscore/1` returns `"requires2fa"` instead of
+  `"requires_2fa"` - the roundtrip breaks and the field becomes unreachable.
+
+  ## How this adapter works
+
+  - `to_external_name/2` behaves identically to `LanguageConventions` (camelCase output).
+  - `to_internal_name/2` applies `Macro.underscore/1` and then inserts underscores
+    at letter-to-digit boundaries. This ensures that names like `"requires2fa"`
+    correctly resolve to `"requires_2fa"`.
+
+  ## Examples
+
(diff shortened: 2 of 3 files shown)


benwilson512 (changes requested):
So the issue here is that this is a breaking change. We can't just change how the language_conventions adapter works on people.

The adapter pattern is pluggable by design. I am open to creating a _new_ adapter that follows some sort of standardized camel to snake definition or spec. I don't really love regexing our way out of the problem though.

donleandro:
Hey @benwilson512, rewrote this from scratch based on your feedback. The key changes:

1. **New adapter, not a modification** - this is now `Absinthe.Adapter.DigitAwareLanguageConventions`, a separate opt-in adapter. `LanguageConventions` is completely untouched, so zero breaking change risk.

2. **Simpler implementation** - dropped the camelCase-detection heuristic from the previous version. The adapter just does `Macro.underscore` followed by a single `String.replace(name, ~r/([a-z])(\d)/, "\\1_\\2")` to insert underscores at letter-digit boundaries. That's it.

The trade-off is that if someone has a field like `:req1` (digit as part of the word, no underscore), this adapter would convert the external `req1` to `req_1` internally, which wouldn't match. But in practice, `letter_digit` patterns (`:requires_2fa`, `:address_line_1`) are way more common in Elixir schemas than `letterdigit` patterns (`:req1`). And since it's opt-in, people only switch to it when they actually need it.

We've been using this approach in production at Shiko for a while now (we had a `:requires_2fa` field on login that was completely unreachable via GraphQL). Works great.

All 1484 existing tests pass, plus 15 new tests covering digit boundaries and roundtrip verification.

benwilson512:
Hey @donleandro I don't necessarily object to the LLM development here but do make sure to run the formatter so that the CI passes :)

Issue #1352, opened 2025-01-30
Performance Degradation Under Load From Fragments (Blueprint Walk Churns Memory?)


jeffutter:
# Summary

My team and I have been struggling to track down a GraphQL related performance 'regression' for the past couple of weeks. We have two versions of a query that are effectively identical, one has more of it's fragments as Spread/Named fragments one has more of them as in-lined fragments. The text document for both are > 200kb and the Blueprints are very large (> 5MB from `:erts_debug.size` IIRC).

The version with more inline fragments has a slightly smaller text size and blueprint size.

Under heavy load the version with the in-lined fragments has **2-5x worse p95 latency** and uses ~2x the CPU.

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/d6a2b531-14b7-4e79-8b65-a1475508fbeb" />

**I'm hoping folks can read this explanation and see if my hypothesis makes sense. I'm making some logical jumps that I haven't been able to specifically measure and substantiate yet.**

# Details Background

- The two versions of the query are:
    - A.) Handwritten by our clients devs **(Referred to as the "Good query" from here on out)**
    - B.) That same query passed through Apollo Router with only one subgraph behind the router (essentially Apollo doing the optimizations it thinks are appropriate **(Referred to as the "Bad query" from here on out - because who likes slow p95s?)**
- Both versions
    - Have a lot of fragments
    - Often have `@skip` or `@include` directives on the fragments
    - Return the exact same data
    - _Seem_ to execute the same as far as resolvers go
- The "Bad" query has a smaller byte size
- The "Bad" query's blueprint is smaller
- The "Bad" query seems to have many of the fragments that were only used once converted to inlined fragments (by Apollo Router)
- We've seen similar poor performance with a 3rd variation of the query where _all_ fragments were converted to named fragments
- We have completely removed Apollo Router from the equation by dumping the queries it makes and injecting them into our load test framework
    - Now the load tests make the same requests the router would make as if the router were in place
    - This removes any theory about it being some kind of connection pooling / load balancing issue with Apollo Router
- In any kind of isolated testing the "Bad" query actually performs faster (It's slightly smaller, why wouldn't it?)
    - We've tried looking at the following
        - Traces in Datadog
        - eprof profiles
        - cprof profiles
        - flamegraphs
    - In any of these situations the "Bad" query is actually (as one might expect) better

## The only difference

The only inkling of a difference I can find between the two queries is this:

[...]

PR #1425, opened 2026-04-10
feat: add per-phase telemetry events to pipeline

aloukissas:
**What**

Adds `[:absinthe, :pipeline, :phase, :start]` and `[:absinthe, :pipeline, :phase, :stop]` telemetry events around each `phase.run` call in `Absinthe.Pipeline.run_phase/3`. The phase module atom (e.g. `Absinthe.Phase.Document.Arguments.CoerceLists`) is included in event metadata.

**Why**

To enable per-phase memory profiling in live-api. Each GraphQL operation runs through ~20+ phases that walk the entire blueprint tree (see absinthe-graphql/absinthe#1352). By measuring memory allocation and GC churn per phase, we can identify which phases are the most wasteful for our specific queries and target optimizations.

No handler is attached by default — zero overhead unless a consumer explicitly attaches to these events.

**Tests**

- `mix compile --warnings-as-errors` passes
- Telemetry events are no-ops when no handler is attached (verified by existing test suite)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/absinthe/pipeline.ex
+++ b/lib/absinthe/pipeline.ex
@@ -14,6 +14,9 @@ defmodule Absinthe.Pipeline do
 
   alias Absinthe.Phase
 
+  @phase_start [:absinthe, :pipeline, :phase, :start]
+  @phase_stop [:absinthe, :pipeline, :phase, :stop]
+
   @type data_t :: any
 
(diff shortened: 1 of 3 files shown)


aloukissas:
Opened against wrong repo by mistake, closing.

PR #1377, opened 2025-09-05
feat: Implement @defer and @stream directives for incremental delivery

jwaldrip:
## Summary

This PR implements the GraphQL `@defer` and `@stream` directives for incremental delivery as specified in the [GraphQL Incremental Delivery RFC](https://github.com/graphql/graphql-spec/blob/main/rfcs/DeferStream.md).

> **Note:** These directives are still in draft/RFC stage and not yet part of the finalized GraphQL specification. The implementation may change as the specification evolves.

### Core Directives
- **@defer directive**: Defer execution of fragments to reduce initial response time
- **@stream directive**: Stream list fields incrementally with configurable batch sizes
- **Opt-in**: Requires `import_directives Absinthe.Type.BuiltIns.IncrementalDirectives` in your schema

### Unified Streaming Architecture
- **Subscriptions with @defer/@stream**: Subscriptions now support incremental delivery automatically
- **Shared execution path**: Both queries and subscriptions use the same `TaskExecutor` for deferred tasks
- **Backwards compatible**: Existing PubSub implementations work unchanged

### Pluggable Executors
- **Custom backends**: Implement `Absinthe.Streaming.Executor` for Oban, RabbitMQ, GenStage, etc.
- **Flexible configuration**: Schema-level, per-request, or application-wide
- **Default executor**: `Task.async_stream` with configurable concurrency and timeouts

### Transport & Integration
- **Transport support**: GraphQL-WS, Server-Sent Events, and extensible transport layer
- **Dataloader integration**: Maintains efficient batching with incremental delivery
- **Relay compatibility**: Full support for streaming Relay connections

### Observability
- **Telemetry events**: `[:absinthe, :incremental, :delivery, :*]` for monitoring
- **on_event callback**: Custom monitoring integrations (Sentry, DataDog, etc.)
- **Complexity analysis**: Proper cost calculation for deferred/streamed operations

## Architecture

```
Absinthe.Streaming
├── Executor        - Behaviour for pluggable execution backends
├── TaskExecutor    - Default executor (Task.async_stream)
└── Delivery        - Handles pubsub delivery for subscriptions

Query/Mutation Path:
  Request → StreamingResolution → Transport → TaskExecutor → Client

Subscription Path:
  Mutation → Subscription.Local → StreamingResolution → Streaming.Delivery
           → TaskExecutor → pubsub.publish_subscription/2 → Client
```

### Enable in Schema

```elixir
defmodule MyApp.Schema do
  use Absinthe.Schema

[...]

PR #1426, opened 2026-04-12
refactor(streaming): resolve-then-split strategy for @defer

jwaldrip:
## Summary
- Rewrites `StreamingResolution` to resolve everything in one pass (standard resolution), then store `@defer` metadata in the streaming context
- Removes complex sub-blueprint re-resolution that caused multiple crashes (nil selections in Projector, missing `:path` key in Execution struct, broken path tracking across siblings in prewalk)
- The transport layer (absinthe_plug) handles splitting the final result into initial/incremental SSE payloads

## What changed
**`lib/absinthe/phase/document/execution/streaming_resolution.ex`** — full rewrite

Old approach: skip deferred nodes during initial resolution → re-resolve them later via sub-blueprints → broken
New approach: resolve everything → collect defer metadata via AST walk → let plug split the result

Follow-up commit addresses Bugbot review feedback:
- Handle `Fragment.Spread` nodes (not just `Fragment.Inline`) when walking for `@defer`, including traversal of named fragments
- Populate backwards-compat keys (`deferred_fragments`, `streamed_fields`, `deferred_tasks`, `stream_tasks`) on `__streaming__` context so downstream `Absinthe.Incremental.*` consumers don't raise `KeyError`

## Test plan
- [ ] `@defer` query returns initial payload without deferred fields + incremental payload with deferred data
- [ ] Normal queries (no `@defer`) continue to work unchanged
- [ ] SSE transport delivers 3 events: initial (with `pending`/`hasNext`), incremental, complete
- [ ] `@defer` on fragment spreads (`...FragmentName @defer`) is honored
- [ ] `@defer` on inline fragments continues to work

## Depends on
- gigsmart/absinthe_plug PR for SSE wiring (companion PR)

## Upstream note
Companion to gigsmart/absinthe#8. Original author: @juscyllan.


benwilson512:
Hey @jwaldrip I think we are rapidly approaching an impasse here.

I have not received any communication public or private on https://github.com/absinthe-graphql/absinthe/pull/1377 which I had to unship due to a mixture of CI failures, AI fragments, and a general lack of human oversight.

This PR clearly builds on the previous work and doesn't even realize I unshipped it. I am more than happy to have a larger conversation here given the important areas of work at hand here but I want to be very clear that PRs like this are entirely unsuitable.

jwaldrip:
@benwilson512 I disagree with many of your assertions here and the reasoning to unship the original code. It DID go through community human review beyond the review I gave it with my own eyes and additional back pressure from claude. I understand you are the maintainer of this project. But @gigsmart will maintain their own fork and stop contributing to the main repo. The review cycle is too long, and frankly our direction on LLM and our internal advancements on such don't align with your philosophy. Thank you for your original contributions and I hope to see some advancements come to the core repo in addition to our fork.

PR #1428, opened 2026-04-15
fix: do not remove comments when formatting

### LINKED ISSUE #1427: `Absinthe.Formatter` removes comments
The formatter removes comments from queries, even from files. We use GraphQL for documentation and testing, and need to keep these comments.

Minimal reproduction:
```elixir
iex> Mix.install([:absinthe])
iex> Absinthe.Formatter.format("""
        # This query does that
        query {
          entity { id } }
        """)
"query {\n  entity {\n    id\n  }\n}\n"
```

I did not dig in to the spec, but comments are documented in the [Schema](https://graphql.com/learn/schema/#annotating-the-schema) part of the documentation on graphql.com.

Gladear:
An attempt to fix #1427

Include comments when formatting using `Absinthe.Formatter`.

~~Since I'd rather not change the internal structure of documents too much, I chose to rather rely on a more hacky solution, extracting the comments without relying on the lexer or parser in the formatter, and force including them on render. Feel free to tell me if you think the solution is wrong, and I should rather go "the long way" and modify the internal structure.~~
See https://github.com/absinthe-graphql/absinthe/pull/1428#issuecomment-4259019807

### DIFF
--- a/lib/absinthe/blueprint/draft.ex
+++ b/lib/absinthe/blueprint/draft.ex
@@ -6,7 +6,10 @@ end
 
 defimpl Absinthe.Blueprint.Draft, for: List do
   def convert(nodes, root) do
-    Enum.map(nodes, &Absinthe.Blueprint.Draft.convert(&1, root))
+    nodes
+    |> Enum.map(&Absinthe.Blueprint.Draft.convert(&1, root))
+    # Comments are converted to `nil` values to be ignored
+    |> Enum.reject(&is_nil/1)
   end
 end
 
--- a/lib/absinthe/blueprint/draft.ex
+++ b/lib/absinthe/blueprint/draft.ex
@@ -6,10 +6,12 @@ end
 
 defimpl Absinthe.Blueprint.Draft, for: List do
   def convert(nodes, root) do
-    nodes
-    |> Enum.map(&Absinthe.Blueprint.Draft.convert(&1, root))
-    # Comments are converted to `nil` values to be ignored
-    |> Enum.reject(&is_nil/1)
+    Enum.flat_map(nodes, fn node ->
+      case Absinthe.Blueprint.Draft.convert(node, root) do
+        nil -> []
+        converted -> [converted]
+      end
+    end)
   end
 end
 
--- a/lib/absinthe/language/comment.ex
+++ b/lib/absinthe/language/comment.ex
@@ -14,4 +14,8 @@ defmodule Absinthe.Language.Comment do
     # Comments are not part of the executable document; skip them silently.
     def convert(_comment, _doc), do: nil
   end
+
+  defimpl Inspect do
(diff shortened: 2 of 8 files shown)


benwilson512:
Hey @Gladear definitely appreciate the effort at a bugfix here. Unfortunately the hacky way here is I think too hacky. I don't have a ton of confidence in Regexes for this purpose, and shoving the data in the pdict has its own risks as well.

Gladear:
@benwilson512 I updated the PR to use the lexer and parser, as well as introduce an `Absinthe.Language.Comment` node.

_AI disclaimer_: I used Claude Code to help me understand the code structure and make the first changes. I thoroughly reviewed the generated code (and modified it in some places), but there are still limitations to my understanding:
1. I reviewed the code that was modified and ensured it looked OK, but I don't have "the bigger picture", the whole project in mind
2. the `:absinthe_parser` module is pretty hard to understand to me 😵‍💫 I never worked with erlang's yecc before, and well, let's say it's not intuitive to me 😅 So I kept the code as it was generated in this file, and couldn't review properly.

Gladear:
@benwilson512 anything I can do to help the review? If this is too much internal and/or touchy changes, I can understand it too, feel free to tell me you don't want this PR merged 😄

lib/absinthe/blueprint/draft.ex
```diff
@@ -6,7 +6,10 @@ end
 
 defimpl Absinthe.Blueprint.Draft, for: List do
   def convert(nodes, root) do
-    Enum.map(nodes, &Absinthe.Blueprint.Draft.convert(&1, root))
+    nodes
+    |> Enum.map(&Absinthe.Blueprint.Draft.convert(&1, root))
+    # Comments are converted to `nil` values to be ignored
+    |> Enum.reject(&is_nil/1)
```
maartenvanvliet: can you use flat_map here to remove the extra pass?

```
|> Enum.flat_map(fn node ->
  case Absinthe.Blueprint.Draft.convert(node, root) do
    nil -> []
    converted -> [converted]
  end
end)
```

maartenvanvliet:
how does it deal with graphql multiline comments?

lib/absinthe/blueprint/draft.ex
Gladear: done in [0ec10cb](https://github.com/absinthe-graphql/absinthe/pull/1428/commits/0ec10cb9b902bf5ed28cd8e6884b546a0d22e9ed)

Gladear:
@maartenvanvliet
> how does it deal with graphql multiline comments?

I'll be honest, I did not know these existed 😄 Apparently, they are already handled in main, and still work on this branch. They seem to be treated as block strings, rather than comments. I added a test in 
8afb14d to ensure they are formatted correctly.
It appears there is an issue when adding a multi-line comment before a field, but the issue already exists on main.

PR #1430, opened 2026-04-22
fix(compilation): remove compile time dependency creation in Notation

novaugust:
Overrides the macro env of import_types ast to tell the compiler that it's used within a function, not at compile-time.

In Frame's codebase, this removes the majority of compilation dependencies across schema / resolvers / submodules therein.

Very open to feedback. This change is inspired by the same technique in Phoenix/Plug:

https://github.com/phoenixframework/phoenix/blob/6ceade4288188651c189ce2ef15b0197e5507df7/lib/phoenix/router.ex#L851-L863

The difference here is, I believe this work is _always_ used entirely at runtime, so there's no need for config checking etc. That said, I know very little about Absinthe, so please correct me if I'm wrong! Given no tests are failing I hope that I'm on the right track.

### DIFF
--- a/lib/absinthe/schema/notation.ex
+++ b/lib/absinthe/schema/notation.ex
@@ -1416,12 +1416,21 @@ defmodule Absinthe.Schema.Notation do
   ```
   """
   defmacro import_types(type_module_ast, opts \\ []) do
-    env = __CALLER__
+    # Ideally prevent introducing compile-time dependencies.
+    ast =
+      if Macro.quoted_literal?(type_module_ast) do
+        Macro.prewalk(type_module_ast, &expand_alias(&1, __CALLER__))
+      else
+        Macro.expand(type_module_ast, __CALLER__)
+      end
 
-    type_module_ast
-    |> Macro.expand(env)
-    |> do_import_types(env, opts)
+    do_import_types(ast, __CALLER__, opts)
   end
+
+  defp expand_alias({:__aliases__, _, _} = alias, env),
+    do: Macro.expand(alias, %{env | function: {:__absinthe_function_n__, 1}})
+
+  defp expand_alias(other, _env), do: raise(inspect(other)) #TODO debugging - can remove if nothing triggers this
 
   @placement {:import_directives, [toplevel: true]}
   @doc """
--- a/lib/absinthe/schema/notation.ex
+++ b/lib/absinthe/schema/notation.ex
@@ -1416,21 +1416,22 @@ defmodule Absinthe.Schema.Notation do
   ```
   """
   defmacro import_types(type_module_ast, opts \\ []) do
-    # Ideally prevent introducing compile-time dependencies.
-    ast =
-      if Macro.quoted_literal?(type_module_ast) do
-        Macro.prewalk(type_module_ast, &expand_alias(&1, __CALLER__))
-      else
-        Macro.expand(type_module_ast, __CALLER__)
-      end
+    type_module_ast
+    |> expand_aliases(__CALLER__)
+    |> do_import_types(__CALLER__, opts)
+  end
 
-    do_import_types(ast, __CALLER__, opts)
+  defp expand_aliases(ast, env) do
+    # Ideally prevents introducing compile-time dependencies by setting the alias's env as occurring within a function
+    if Macro.quoted_literal?(ast),
+      do: Macro.prewalk(ast, &expand_alias(&1, env)),
+      else: Macro.expand(ast, env)
   end
 
   defp expand_alias({:__aliases__, _, _} = alias, env),
     do: Macro.expand(alias, %{env | function: {:__absinthe_function_n__, 1}})
 
-  defp expand_alias(other, _env), do: raise(inspect(other)) #TODO debugging - can remove if nothing triggers this
+  defp expand_alias(other, _env), do: other
 
   @placement {:import_directives, [toplevel: true]}
   @doc """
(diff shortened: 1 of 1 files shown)


lib/absinthe/schema/notation.ex
novaugust, line 2448: fwiw, i noticed while digging into this that this function is called on the same ast multiple times. for example, the field/2 macro calls `handle_field_attrs` which calls `expand_ast` before returning the expanded ast, then that ast is passed to `record!` which calls this function again

lib/absinthe/schema/notation.ex
```diff
@@ -1416,13 +1416,23 @@ defmodule Absinthe.Schema.Notation do
...
+
+  defp expand_aliases(ast, env) do
+    # Ideally prevents introducing compile-time dependencies by setting the alias's env as occurring within a function
+    if Macro.quoted_literal?(ast),
+      do: Macro.prewalk(ast, &expand_alias(&1, env)),
+      else: Macro.expand(ast, env)
   end
 
+  defp expand_alias({:__aliases__, _, _} = alias, env),
+    do: Macro.expand(alias, %{env | function: {:__absinthe_function_n__, 1}})
```
novaugust, line 1432: notably i'm using a made up function name here. absinthe creates a bunch of functions, and i'm not sure... which one to blame as the function this alias will be used in. so i used a name to represent all the names. it's possible this will break at some point in the future, but for now the compiler doesn't seem to care

lib/absinthe/schema/notation.ex
```diff
@@ -1416,13 +1416,23 @@ defmodule Absinthe.Schema.Notation do
   ```
   """
   defmacro import_types(type_module_ast, opts \\ []) do
```
Gladear, line 1418: If I recall correctly, I did try that, but this caused an important issue : when modifying the imported module, the schema wasn't recompiled, and therefore the changes were not taken into account.

Honestly, I already spent quite some time trying to get rid of these compile-connected dependencies created by Absinthe schemas, but my conclusion has been that they are a symptom of the declaration API, and will be hard to get rid of without a new API 😕 I'd be happy to be proven wrong, though.

novaugust: Thanks for the heads up! Like I said, Absinthe neophyte here. What do I need to do to compile the schema and prove out whether this is an issue or not? I assume it's a mix task of some sort?

benwilson512: @Gladear so I think the issue goes away if we tie this to using the Persistent Term backend, because the full import graph is re-evaluated at runtime in that scheme.

Can you confirm if you see that issue with the Persistent Term backend?

Gladear: There appears to be issues with the persistent term backend as well. I just tried the following with this branch:

**Confirm my theory (with "compiled" backend)**
1. Launch the project and go to GraphiQL
2. Fetch the schema, run a query
3. Delete the query from the source code and force recompile using `recompile` in IEx
4. Refetch the schema : the query still appears
5. Run the query again : it can be run again

So, theory confirmed I think.

**Try exactly the same with persistent_term backend**
1, 2, 3, same as above

4. Refetch the schema : the query still appears (unexpected ?)
5. Run the query again : it crashed during complexity analysis with the following stacktrace

```
** (FunctionClauseError) no function clause matching in AppGraphQL.AbsintheFields."-inlined-__absinthe_function__/2-"/2
    (app 0.0.1) AppGraphQL.AbsintheFields."-inlined-__absinthe_function__/2-"({Absinthe.Blueprint.Schema.FieldDefinition, {:query_fields, :entity}}, :complexity)
    (absinthe 1.10.0) lib/absinthe/phase/document/complexity/analysis.ex:87: Absinthe.Phase.Document.Complexity.Analysis.handle_node/3
```

novaugust: ah, unfortunate! perhaps it's still possible to find optimizations by continuing the plug analogy, where `init` is compile time in production but configurably runtime (for dev, where perhaps it doesn't matter as much?)

further, even if `import_types` and other schema-defining macros require compile-time dependencies, i wonder if we can still nix that requirement just for the `resolve` argument, which in our codebase spiders this web of dependencies _outside_ absinthe-specific modules of `... resolve: &NonSchemaModule.resolve/3`

Gladear: Just for sake of clarity, this may already be what you say but I'm not entirely sure: `resolve` does not cause a compile-time dependency, but a runtime one. The issue is that a compile-time dependency _into_ a runtime one creates a compile-connected dependency (see https://hexdocs.pm/mix/Mix.Tasks.Xref.html#module-a-brief-introduction-to-xref).

`resolve` would need to completely ignore the dependency to the other module. It is likely doable, iirc Phoenix does it in the router - I must say I don't know how. I think it's a great idea, however!

novaugust:
Came back to this after a good bit off. Indeed, the solution offered up here only goes halfway. 

For this to work, Absinthe would need something like plug builder's `init_mode` that specifies runtime or compile (aka phoenix's `plug_init_mode` setting). Like ben said, this would seemingly work with the persistent term backend since it's doing that work of building the schema at startup, but this isn't shippable for when people want the work done at compile time.

I guess I  _could_ introduce a new setting and only conditionally expand AST in this way, but it'd require some complex guidance for the user to tell them to _also_ swap to the persistent term backend in those cases. Just seems too easy to mess up if it gets implemented piecemeal like that.

PR #1434, opened 2026-05-08
feat: Emit batch :exception telemetry event on raise/throw/exit

iacutone:
When a batch function raised, the task crashed and the exit signal propagated through the `Task.async` link to kill the caller — there was no telemetry hook for observability tooling to see the failure, only :start, :stop, and :timeout events.

This PR wraps the batch function call in a `try/catch` block which raises/throws/exits inside the task. In the event of an exception, emit a new `[:absinthe, :middleware, :batch, :exception]` event with :kind, :reason, and :stacktrace metadata, then re-raise so the exception still surfaces to the caller.

If y'all are happy with this PR, I would like to apply similar exception telemetry events in `Absinthe.Phase.Telemetry`.


distortia (approved):
🙏

cschiewek (approved):

PR #1436, opened 2026-05-27
fix: allow nulls in nullable enum list defaults

jbellenger:
## Summary
Hello! I'm one of the maintainers of the graphql-conformance project, which verifies spec conformance of popular graphql implementations.

In the [conformance results for absinthe](https://jbellenger.github.io/graphql-conformance/impl/absinthe), I noticed some errors that seemed like they would be a straightforward fix (... for codex). This PR fixes handling of allowed null values for lists of enums.

For example, given this schema definition:
```graphql
enum MovieGenre { ACTION }

type Query {
  movies(genres: [MovieGenre] = [ACTION, null]): Int
}
```

Absinthe today throws an error:
```
  Compilation failed:
  ---------------------------------------
  ## Locations
  nofile:4

  The default_value for an enum must be present in the enum values.

  Could not use default value of `nil` for %Absinthe.Blueprint.TypeReference.List{of_type: %Absinthe.Blueprint.TypeReference.Name{name: "MovieGenre", schema_node: nil, errors: []}, errors: []}.

  Valid values are:

   * :action
```

This PR updates enum default handling so that this is allowed.

From what I can tell, Absinthe is far above average in spec conformance, but has a small number of areas where it behaves differently from graphql-js. I encourage you to check out [the absinthe conformance page](https://jbellenger.github.io/graphql-conformance/impl/absinthe) for other behavioral differences (a relatively simple one is that absinthe does not allow operations named "ON", "repeatable", "fragment", etc).

### DIFF
--- a/lib/absinthe/phase/schema/validation/default_enum_value_present.ex
+++ b/lib/absinthe/phase/schema/validation/default_enum_value_present.ex
@@ -64,8 +64,16 @@ defmodule Absinthe.Phase.Schema.Validation.DefaultEnumValuePresent do
     |> Enum.find({:ok, value}, &match?({:error, _}, &1))
   end
 
+  defp value_conforms_to_enum(%Blueprint.TypeReference.NonNull{}, nil, _) do
+    {:error, nil}
+  end
+
   defp value_conforms_to_enum(%_{of_type: of_type}, value, enum_values) do
     value_conforms_to_enum(of_type, value, enum_values)
+  end
+
+  defp value_conforms_to_enum(_, nil, _) do
+    {:ok, nil}
   end
 
   defp value_conforms_to_enum(_, value, enum_values) do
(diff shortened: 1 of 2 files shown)


maartenvanvliet:
cool project!

cschiewek (commented):
Thank you for your submissions!

The changes look good, but the tests could use some work.  The majority of these should be restructured as unit tests against the public functions defined in the `Absinthe.Phase.Schema.Validation.DefaultEnumValuePresent`.  

I know that your additions just followed the existing test style and structure, but ideally, we'd clean them up.

If you're amenable to those changes, let me know.  Thanks!

jbellenger:
Thanks @cschiewek! 
I'm not familiar with elixir, but I tasked codex with rewriting the tests that I added and this is the result. 
Very happy to make other changes, including rewriting the existing tests in this file.

cschiewek (approved):
Tests aren't _ideal_, but good enough to fix the bug.  Thanks!

PR #1438, opened 2026-05-28
fix: Expose oneOf via schema introspection

howleysv:
Added the `isOneOf` introspection field as defined in the spec:
https://spec.graphql.org/draft/#sec-Schema-Introspection.Schema-Introspection-Schema


cschiewek (approved):
Thank you.

PR #1440, opened 2026-06-03
chore(security): Add Sobelow scanning to CI

s3cur3:
Hi there! In response to the rash of CVEs from vulnerabilities that static analysis can detect (see [the EEF's list of most common weaknesses](https://cna.erlef.org/common-weaknesses), of which uncontrolled resource consumption like as atom table exhaustion is number 1), I thought it might be valuable to add Sobelow scanning to the project. It didn't find anything actionable (yay! 🎉), but I figured it might be valuable to have it running in CI so that future potential vulnerabilities can be called out.


cschiewek (commented):
@s3cur3 Thanks for submitting this.

A few changes I would love to see:

- Can you remove all the duplicative comments description each `["DOS.StringToAtom"]`.  A single comment (maybe in the config?) that covers all the ListToAtom warnings.

- You need to run `mix format`, it's failing on this branch. 

Thanks!

s3cur3:
@cschiewek I'd recommend against removing the inline `sobelow_skip` comments. It's _possible_ to set up a `.sobelow-skips` file and put both the lines to skip and the rationale for skipping them there, but doing so is *extremely* annoying in my experience, because Sobelow ties the skip rule to the exact line number _and_ a hash of the line that has to be generated via `mix sobelow --mark-skip-all`. Regenerating that file when the line numbers change is both tedious and does not preserve the order, so the comments are liable to become nonsense over time. 🙁 (There's a [PR](https://github.com/sobelow/sobelow/pull/24) to fix this, but it hasn't been acted on in 6 months.)

cschiewek:
> @cschiewek I'd recommend against removing the inline `sobelow_skip` comments. It's _possible_ to set up a `.sobelow-skips` file and put both the lines to skip and the rationale for skipping them there, but doing so is _extremely_ annoying in my experience, because Sobelow ties the skip rule to the exact line number _and_ a hash of the line that has to be generated via `mix sobelow --mark-skip-all`. Regenerating that file when the line numbers change is both tedious and does not preserve the order, so the comments are liable to become nonsense over time. 🙁 (There's a [PR](https://github.com/sobelow/sobelow/pull/24) to fix this, but it hasn't been acted on in 6 months.)

Apologies, I should have been clearer.  I'm not saying to remove the skips themselves, but the duplicative comments associated with each, specifically, I don't think we need this
```
# Schema SDL is developer-authored and processed at build time; identifiers
# are atoms by Absinthe's design, so this is not attacker-controlled input.
```
for _every_ module that skips it.  That being said, I don't think it's a deal breaker, so I'm fine if others feel strongly about keeping them.

cschiewek (approved):

cschiewek:
@s3cur3 if you address the conflicts, I'm happy to merge this.

PR #1441, opened 2026-06-04
fix: elixir 1.20 redundant clause warnings

### LINKED ISSUE #1429: Elixir 1.20 compatibility: redundant clause warnings from duplicate directive artifacts
While fixing Elixir 1.20 compatibilities on `absinthe_federation` I found this compatibility issue in absinthe, that it's important because it's in macros so the warnings happens on user's projects.

Elixir 1.20's new type inference detects redundant function clauses, surfacing a latent bug where directive_artifacts contains duplicates when a schema uses a custom @prototype_schema combined with import_directives from the same module.

 ImportPrototypeDirectives sets directive_artifacts from the prototype, then Build appends directives built from directive_definitions (which includes the imported ones) without deduplication:

 ```elixir
   # lib/absinthe/phase/schema/build.ex:13
   directive_artifacts: schema.directive_artifacts ++ directive_artifacts
 ```

 This generates duplicate __absinthe_directive__/1 clauses in the compiled module. Previously silent, Elixir 1.20 now emits warnings for these redundant clauses, which breaks builds using --warnings-as-errors.

 ### Reproduction

 ```elixir
   defmodule MyPrototype do
     use Absinthe.Schema.Prototype

     directive :custom do
       on [:schema]
     end
   end

   defmodule MySchema do
     use Absinthe.Schema

     @prototype_schema MyPrototype
     import_directives MyPrototype

     query do
       field :hello, :string
     end
   end
 ```

 ```
   $ mix compile --force

   warning: the following clause is redundant:

       def __absinthe_directive__(:custom)

   └─ deps/absinthe/lib/absinthe/phase/schema/compile.ex:93: MySchema.Compiled.__absinthe_directive__/1
 ```

 All prototype directives are duplicated (:custom, :deprecated, :one_of, :specified_by).

 ### Suggested fix

 Deduplicate by identifier in Build.run/2:

 ```elixir
   directive_artifacts =
     (schema.directive_artifacts ++ directive_artifacts)
     |> Enum.uniq_by(& &1.identifier)
 ```

 ### Environment

 - Absinthe master branch
 - Elixir 1.20.0-rc.4
 - Erlang/OTP 28

novaugust:
closes #1429 , thanks to the author for the fix

in addition, removed 2 unused require Logger, and an unnecessary if statement caught by 1.20


lib/absinthe/phase/schema/build.ex
```diff
@@ -5,13 +5,14 @@ defmodule Absinthe.Phase.Schema.Build do
...
 
-    schema = %{
-      schema
-      | type_artifacts: types,
-        directive_artifacts: schema.directive_artifacts ++ directive_artifacts
-    }
+    directive_artifacts =
+      blueprint
+      |> build_directives()
+      |> Enum.concat(schema.directive_artifacts)
```
novaugust, line 12: note that i flipped the list ordering (purely for pipe ergonomics) of schema.artifacts vs blueprint artifacts. i figure order is irrelevant since they're all unique anyways.
if someone knows better please say so :)

lib/absinthe/type/custom/decimal.ex
```diff
@@ -21,8 +21,7 @@ if Code.ensure_loaded?(Decimal) do
     end
 
     def parse(%Input.Integer{value: value}) when is_integer(value) do
-      decimal = Decimal.new(value)
-      if Decimal.nan?(decimal), do: :error, else: {:ok, decimal}
```
novaugust, line 25:
```
    warning: the following clause cannot match because the previous clauses already matched all possible values:

        _ ->

    it attempts to match on the result of:

        Decimal.nan?(decimal)

    which has the already matched type:

        dynamic(false)

    type warning found at:
    │
 25 │       if Decimal.nan?(decimal), do: :error, else: {:ok, decimal}
    │       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    │
    └─ (absinthe 1.10.2) lib/absinthe/type/custom/decimal.ex:25: Absinthe.Type.Custom.Decimal.parse/1
```

.github/workflows/ci.yml
cschiewek:
```suggestion
- elixir: 1.20
  otp: 25
- elixir: 1.20
  otp: 26
- elixir: 1.19
  otp: 25
- elixir: 1.17
  otp: 28
```

cschiewek (approved):
Thank you @novaugust

rockneurotiko:
Thanks!

Issue #1245, opened 2023-05-26
Huge max_heap_size compared to the actual data size


IvanIvanoff:
### Environment

* Elixir version (elixir -v): Elixir 1.15.0-dev (e0c1084) (compiled with Erlang/OTP 25)
* Absinthe version (mix deps | grep absinthe): absinthe 1.7.1 

### Actual behavior

A few days ago our application started getting killed with `Out of memory` error.
What we found is that for some queries, even if the size of the response is small, the memory consumed by the process that processes the GraphQL query is huge. When multiple such queries happen concurrently, the RAM usage jumps from 1Gb to over 6Gb. As a temporary solution we set the `:max_heap_size` flag in the resolver before processing the query, so these queries can be killed instead of killing the whole application.

Here is a project that reproduces the issue: https://github.com/IvanIvanoff/absinthe_memory

In this project, we observe that when the HTTP response is small in size at 828 kB, Absinthe peaks at over 50 Mb of RAM usage when processing it.

### Expected behavior

The peak memory usage should not be so big.

benwilson512:
Happy to look into it. I suspect that using persistent term will help. Overall though for bulk data like timeseries data the fact the remains that there is going to be a fair bit over overhead, both in processing and memory, compared to raw JSON. Absinthe is going to type annotate and type check every single value and if you want to move thousands of values, that adds up.

benwilson512:
Overall I would suggest that you treat bulk data transfers a bit like you'd treat images or binary data in GraphQL: Link to them. We have a bulk data controller that takes a set of signed parameters and returns JSON. Then in our GraphQL we return links to those endpoints with the signed params embedded.

crayment:
Just wanted to voice that we are also running into this issue. We would love to be able to return large numbers of records through GraphQL - but these massive memory spikes make that impossible and we are having to move to other solutions.

I don't understand the problem well enough to know if type annotation and checking is something that requires using as much memory as we are seeing. The sample app linked above using 50x the memory of the actual response seems excessive, but perhaps that's just what it is? Feels tempting to look for some memory wins here...

benwilson512:
Hey @crayment There is certainly no low hanging fruit here. For some thing as simple as a json response of `{"hello":"world"}` the key `"hello"` is just 5 bytes, but there is easily several hundred bytes worth of data structures holding on to the type information for casting, validation, and traceback information in case there is an error.

[...]

PR #1448, opened 2026-07-10
Reduce RAM memory usage by 60-90% in some cases

IvanIvanoff:
### Changes - Reduce memory footprint of large list results

Disclosure: Claude Fable helped write the code.

This PR addresses issues reported in [#1245](https://github.com/absinthe-graphql/absinthe/issues/1245).

Returning big lists used to blow up process memory way out of
proportion to the response - a ~0.9 MB payload peaking at 30+ MB of heap.
Turns out most of that wasn't the type checking itself, just avoidable duplication while building the result tree.

Three small, self-contained commits (details in each git message):
- share the fieldemitter across list elements instead of rebuilding it per element,
-  drop someper-element allocation churn
- collapse nullable scalar/enum lists into a single node instead of one struct per value.

Peak heap on the #1245 repro and a couple of variants:

| Query | Before | After |
|-------|--------|-------|
| 20k objects / 40k leaves | 31.26 MB | 11.58 MB (−63%) |
| list of 100k strings | 24.97 MB | 2.44 MB (−90%) |

No public API change — `Absinthe.run/3` output is byte-identical. The only thing
to flag: the scalar-list commit adds a new result node, `Result.LeafList`, so a
custom result phase that walks the tree needs one extra clause: https://github.com/absinthe-graphql/absinthe_phoenix/pull/111

PS: I see that bench tests are intentionally ignored.
I pushed the bench tests to my fork here: https://github.com/santiment/absinthe/commit/b1ee3c88d7a876cb4b5bf03bc116d3b29b801df1

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,8 @@
-/bench
+# Track the committed benchmark suite (scripts + README) but keep any local
+# scratch output under bench/ ignored.
+/bench/*
+!/bench/*.exs
+!/bench/README.md
 /_build
 /cover
 /deps
--- a/bench/README.md
+++ b/bench/README.md
@@ -1,0 +1,62 @@
+# Memory benchmarks
+
+Benchmarks for the memory footprint of query execution, originally written for
+[#1245](https://github.com/absinthe-graphql/absinthe/issues/1245) (large list
+results using far more RAM than the response size).
+
+## Running
+
+Benchee is a `:dev`-only dependency, so run these under `MIX_ENV=dev`:
+
+```sh
+# Peak process heap (what trips :max_heap_size). Best of 3 per query.
+MIX_ENV=dev mix run bench/peak_heap.exs
+
+# Total bytes allocated per request (churn / GC pressure), via Benchee.
+MIX_ENV=dev mix run bench/memory.exs
+```
+
+Both scripts load `bench/bench_schema.exs`, which defines a small in-memory
+schema (no database, no custom types) with four queries: a large nested-object
+list (the #1245 shape), the same with a non-null element type, a flat list of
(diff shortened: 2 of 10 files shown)


benwilson512:
Hey there! This is great, definitely happy to move this towards a release. Adding a new blueprint node means that this will probably be a 1.12 release. If you can review the CI steps that are failing that'd be a great start.

IvanIvanoff:
Hi @benwilson512,

I actually intended to first push to my fork only and let my app work on stage for a few days and on prod for a few days before I publish it here. But I pushed here, too, so I moved it to draft which cancelled the checks.

One of the broken checks is that absinthe_phoenix tests don't work - but for them to work first we need to release this https://github.com/absinthe-graphql/absinthe_phoenix/pull/111, right?

Either way, I'll first let the new code work in a real environment for a few days and if everything looks good - I'll continue updating here. It works fine locally and all tests pass, so I don't expect any significant issues to be found.

benwilson512:
Ah ya no worries. As it happens I was also in the middle of some refactors to the Resolution phase, I may fold some of this stuff in since they both impact some of the common functions. Will be sure to provide credit where credit is due!

IvanIvanoff:
@benwilson512 I'm running the code on stage for ~3 days and on prod for ~1 day without issues.

I see you've included part of the PR in yours - in such case we can close this one and open a new one for the LeafList after yours is merged?

benwilson512:
Hey @IvanIvanoff yes please can you re-open with just the leaf list part? Happy to merge in prep for 1.12.

IvanIvanoff:
Closing in favor of:
- #1450 includes everything except the LeafList struct
- #1451 inlucdes the LeafList struct

PR #1450, opened 2026-07-12
perf: Avoid rewalking the tree for suspended fields

### LINKED ISSUE #1094: Null trimming not applied when child field uses an async resolver
I think we have an issue with null trimming specifically with async resolvers, but I'm not sure how to solve it.

If I'm understanding the issue, "null trimming" happens on the resolution of a field by looking at its children to see if any non-null violations have occurred. For a field with children using async resolvers, those children might actually be `Resolution` structs in `state: :suspended`, not `Result.*` structs, when the parent field is deciding whether any trimming needs to occur.

If one of those child fields finally resolved `null` in a second resolution pass, the error might be added, but I think the nullable parent field won't be revisited, so no trimming will occur?

I'm not super familiar with this chunk of the codebase, so I'm not sure what the right solution is, assuming my hypothesis is even correct. 

Maybe we could add another phase to be responsible for null trimming? Not sure if that should happen before or after `Absinthe.Phase.Document.Result`. 

Or maybe the fix is actually quite simple but not obvious to my novice eye.... 🤞 

I'll open a PR with some failing tests I added as an example. 

### Environment

* Elixir version (elixir -v): `Elixir 1.12.1 (compiled with Erlang/OTP 24)`
* Absinthe version (mix deps | grep absinthe): `master`

### Expected behavior

A query like this:

```
{
  nullable { nullable { nonNull(makeNull: true) { __typename }}}
}
```

Should return data that looks something like this:

```
%{
      data: %{"nullable" => %{"nullable" => nil}},
      errors: [
        %{
          locations: [%{column: 25, line: 2}],
          message: "Cannot return null for non-nullable field",
          path: ["nullable", "nullable", "nonNull"]
        }
      ]
    }
```

### Actual behavior

Instead we produce this:
```
%{
      data: %{"nullable" => %{"nullable" => %{"nonNull" => nil}}},
      errors: [
        %{
          locations: [%{column: 25, line: 2}],
          message: "Cannot return null for non-nullable field",
          path: ["nullable", "nullable", "nonNull"]
        }
      ]
    } 
```

`Absinthe.Phase.Document.Execution.Resolution`

benwilson512:
# Rework resolution to stop re-walking the result tree

## The problem

The resolution phase re-runs once per "round" of suspension (async, batch,
dataloader). Each re-run walked the *entire* result tree (including subtrees
that fully resolved in an earlier pass) just to find the suspended fields.
For a document with S result nodes and K suspension rounds (chained dataloader
batches being the common worst case), that's O(K × S) work and allocation,
even though the useful work per round is only the suspended fields themselves.

## What changed

**Suspended fields now live in a flat pool instead of the tree.** When a field
suspends, it leaves a lightweight `%Result.Pending{ref: ref}` placeholder in
the result tree and the suspended `%Absinthe.Resolution{}` goes into
`execution.pending`. Re-runs of the resolution phase iterate only that pool —
the tree is never re-walked. Once nothing remains suspended, a single
substitution pass splices the resolved subtrees in over their placeholders,
applying non-null propagation along the changed paths.

**This fixes #1094.** Previously, a batched/async/dataloader field that
resolved to `nil` in a `non_null` position produced the error but did *not*
nullify its ancestors, returning spec-invalid data (the old re-walk rebuilt
ancestor objects without re-applying null trimming). Null propagation is now
identical whether a field resolves synchronously or suspends — covered by a
new test suite that runs the same documents in sync/batch/async modes and
requires identical results. This is a behavior change for anyone depending on
the old, non-compliant shape.

**Plus a round of sync-pass optimizations**: fields using the default
`MapGet` resolver skip building a resolution struct entirely (a pure
`Map.get/2` cannot error or suspend, so the struct and middleware loop were
pure overhead), list elements share one prepared emitter instead of
re-expanding the field's type per element (borrowed from #1448 — purely
internal, result nodes carry identical values), the concrete parent type is
recorded once in the cached projection instead of copied onto every field per
element, and lists skip the null-trimming rebuild when nothing needs
trimming.

Deliberately **not** included: the rest of #1448 — in particular
`Result.LeafList`, which changes the result node set that custom result
phases walk and so needs a coordinated absinthe_phoenix release; it should
ship as its own PR.

## Numbers

Resolution + result phases, 10k-item list with 4 scalar fields each
(`benchmarks/resolution_rewalk.exs`; chain_24 = 24 chained batch levels next
to the wide tree, i.e. 25 resolution passes):

| Scenario | before | after |
|---|---|---|
| sync (1 pass) | 33.2 ms / 65.9 MB | **19.3 ms / 26.3 MB** |
| batched N+1 (2 passes) | 60.8 ms / 126 MB | **45.9 ms / 83.7 MB** |
| chain_24 (25 passes) | 81.9 ms / 418 MB | **15.0 ms / 31.3 MB** |

The suspension-round penalty is effectively gone: each additional round now
costs O(pending fields) instead of O(tree), so chain_24 runs ~5x faster with
~13x less allocation. The all-sync pass is ~40% faster with ~60% less
allocation, and the dense single-round N+1 case is ~25% faster with ~35%
less allocation.

## Compatibility notes

- `Absinthe.run/3` output is unchanged except for the #1094 fix above.
- A new internal result node exists (`Result.Pending`) and
  `%Blueprint.Execution{}` grew `pending`/`resolved` fields. `Pending` never
  survives to the result phase in normal operation, so custom result phases
  are unaffected.
- Fields resolved by the default `MapGet` middleware no longer construct an
  `%Absinthe.Resolution{}` at any point. Nothing can observe the difference
  (no middleware runs for such fields, and abstract-typed fields keep the
  regular path), but the invariant is worth knowing.
- Within a re-run, suspended fields resume in document order from the pool;
  interleaving with newly-suspended nested fields can differ slightly from
  the old tree-order walk. Batch/async/dataloader are order-insensitive.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,7 +77,7 @@ jobs:
         with:
           path: |
             priv/plts
-          key: "v1-${{ runner.os }}-\
+          key: "v2-${{ runner.os }}-\
             erlang-${{ matrix.otp }}-\
             elixir-${{ matrix.elixir }}-\
             ${{ hashFiles('mix.lock') }}"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,19 +1,37 @@
 # Changelog
 
+## [1.12.0](https://github.com/absinthe-graphql/absinthe/compare/v1.11.0...v1.12.0) (Pending)
+
+### Features
+
+- Significantly Improved performance when using suspended fields (Dataloader,
+  Batching)
+- Significantly Improved memory usage when returnign lists of scalars
+
+
 ## [1.11.0](https://github.com/absinthe-graphql/absinthe/compare/v1.10.0...v1.11.0) (2026-06-04)
 
 
 ### Features
 
-* add description support for operations and fragments ([#1404](https://github.com/absinthe-graphql/absinthe/issues/1404)) ([c029a61](https://github.com/absinthe-graphql/absinthe/commit/c029a611640a764b687f6e3b5ce647880eff99fd))
-* PartitionSupervisor for subscription tasks ([#1416](https://github.com/absinthe-graphql/absinthe/issues/1416)) ([18665af](https://github.com/absinthe-graphql/absinthe/commit/18665af918a1f135171feec4f6b0d494ac105339))
+* add description support for operations and fragments
+  ([#1404](https://github.com/absinthe-graphql/absinthe/issues/1404))
+  ([c029a61](https://github.com/absinthe-graphql/absinthe/commit/c029a611640a764b687f6e3b5ce647880eff99fd))
+* PartitionSupervisor for subscription tasks
+  ([#1416](https://github.com/absinthe-graphql/absinthe/issues/1416))
+  ([18665af](https://github.com/absinthe-graphql/absinthe/commit/18665af918a1f135171feec4f6b0d494ac105339))
 
 
 ### Bug Fixes
 
-* elixir 1.20 redundant clause warnings ([#1441](https://github.com/absinthe-graphql/absinthe/issues/1441)) ([0fd4781](https://github.com/absinthe-graphql/absinthe/commit/0fd478115b0450cca00a3881ec756524e5c15bdc))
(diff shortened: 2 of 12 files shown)


bruce (approved):
Nice approach, and the initial benchmarks are impressive!

PR #1454, opened 2026-08-11
fix(compilation): mark `middleware`, `resolve`, `resolve_type` macro ast as runtime

novaugust:
Greetings! It's me again, taking another stab at making some parts of the absinthe notation DSL mark aliases as runtime references to prevent adding compile-time dependencies to the graph when possible.

## Context

The last attempt at this was #1430, which applied this technique to the schema's shape itself. That proved untenable without major absinthe redesigns, but I'm hoping that _these macros_ at least really do only apply their code during runtime and so can be marked as the same.

## Benefit

As for the level of benefit here, it'll obviously depend on how users write their modules. But for a rough metric, the large codebase I work with saw a 3% decrease in the number of transitive dependencies as totaled by the [`mix recompile_buster`](https://github.com/MrYawe/recompile_buster) tool. 3% doesn't sound large, but keeping in mind that our graphql api is only a small portion of our files, i assure you it's pretty big!

## Other work

I'm confident many other macros in the notation world could have this change applied to them as well. I'd love feedback on what else qualifies as a runtime macro, but just wanted to get eyes to see if it's tenable before doing more work

### DIFF
--- a/lib/absinthe/schema/notation.ex
+++ b/lib/absinthe/schema/notation.ex
@@ -684,10 +684,13 @@ defmodule Absinthe.Schema.Notation do
   end
 
   defp mark_aliases_as_runtime(ast, env) do
-    Macro.prewalk ast, fn
-      {:__aliases__, _, _} = alias -> Macro.expand(alias, %{env | function: {:__absinthe_resolver__, 1}})
-      other -> other
-    end
+    Macro.prewalk(ast, fn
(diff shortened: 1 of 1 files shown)


bryanjos (approved):
Looks good to me! Thanks! I'm going to keep it open a bit longer to see if @cschiewek or @benwilson512 have any opinions

cschiewek (approved):
Makes sense!  Merging.

PR #1460, opened 2026-09-21
fix: handle Unicode surrogate escapes in string values

### LINKED ISSUE #1458: A lone surrogate escape, or a surrogate pair escape, leaks Erlang's ArgumentError text
Versions: absinthe 1.12.0, Elixir 1.20.1 on OTP 29. Present on the default branch at `1372ceb` (checked 2026-09-15).

Found by an AI assistant working with a human who was fuzz-testing their own application; reproduced in a minimal project with default settings.

## Request

Any document with a string literal containing an escaped lone surrogate, for example `{ echo(value: "a\ud800b") }`, or an escaped surrogate pair, `{ echo(value: "\ud83d\ude00") }` (the fixed-width spelling of U+1F600 that the spec keeps for compatibility). Self-contained, no database:
```elixir
Mix.install([{:absinthe, "1.12.0"}])

defmodule Schema do
  use Absinthe.Schema

  query do
    field :echo, :string do
      arg :value, non_null(:string)
      resolve fn %{value: value}, _ -> {:ok, value} end
    end
  end
end

Absinthe.run(~S|{ echo(value: "a\ud800b") }|, Schema)
Absinthe.run(~S|{ echo(value: "\ud83d\ude00") }|, Schema)
```

## Observed

```
{:ok, %{errors: [%{message: "An unknown error occurred during parsing: errors were found at the given arguments:\n\n  * 1st argument: not valid character data (an iodata term)\n"}]}}
```
for both documents. The error carries no `locations`. (The second line of the message is OTP's error formatting and may read differently on older OTP.) `\u{1F600}`, `\uZZZZ` and `\u12` pass through as literal text.

## Expected

A syntax error with a location for the lone surrogate, and U+1F600 for the pair, with no exception text in either case.

## Failure point

`unescape_unicode/5`, [lib/absinthe/lexer.ex#L367-L372](https://github.com/absinthe-graphql/absinthe/blob/v1.12.0/lib/absinthe/lexer.ex#L367-L372), ignores the `{:error, _, _}` return of `:unicode.characters_to_binary/1`; `extract_quoted_string_token/1`, [src/absinthe_parser.yrl#L422-L423](https://github.com/absinthe-graphql/absinthe/blob/v1.12.0/src/absinthe_parser.yrl#L422-L423), then raises; `Absinthe.Phase.Parse` rescues at [lib/absinthe/phase/parse.ex#L86-L88](https://github.com/absinthe-graphql/absinthe/blob/v1.12.0/lib/absinthe/phase/parse.ex#L86-L88) and renders `Exception.message/1` at L116-L127.

## Reproduction

* Description, with the failure point at this release: https://github.com/grempe/ash-fuzz-repros#absinthelone-surrogate-escape-leaks-argument-error
* Failing test: https://github.com/grempe/ash-fuzz-repros/blob/main/test/absinthe/lone_surrogate_escape_test.exs
* Run: `mix setup && mix test test/absinthe/lone_surrogate_escape_test.exs` in https://github.com/grempe/ash-fuzz-repros

ethanstoner:
## Summary

- A string literal containing an escaped surrogate crashed the lexer's output into the parser, which raised `ArgumentError`; the rescue in `Absinthe.Phase.Parse` rendered the raw Erlang text with no `locations`. Unpaired surrogates are now rejected as a located syntax error.
- A surrogate pair escape such as `"\ud83d\ude00"` is now decoded to the single code point it denotes (U+1F600) instead of failing.

Fixes #1458

## Background

`unescape_unicode/5` passed the result of `:unicode.characters_to_binary/1` straight into the token stream. For an escape describing a surrogate that call returns `{:error, binary, rest}` rather than a binary, so the tuple was accumulated into the `:string_value` token:

```elixir
{:string_value, {1, 14}, [34, 97, {:error, "", [55296]}, 98, 34]}
```

`extract_quoted_string_token/1` then reached `unicode:characters_to_binary/1` with a tuple embedded in its iodata and raised, producing:

```
An unknown error occurred during parsing: errors were found at the given arguments:

  * 1st argument: not valid character data (an iodata term)
```

with an empty `locations` list.

The pair case surfaced the same crash, but its root cause is different: `"\ud83d\ude00"` is the fixed-width spelling of U+1F600, and each half was decoded on its own. This works in graphql-js without any special handling because JavaScript strings are UTF-16 and the two escapes combine by themselves — graphql-js says as much where it handles the pair explicitly for the braced form:

> Had JavaScript not been internally based on UTF-16, then this surrogate pair would be decoded to retrieve the supplementary code point.

Erlang binaries are UTF-8, so Absinthe has to do that decoding itself. Clients that escape non-ASCII characters emit this form routinely, so the practical effect was that any such document was rejected.

## Changes

- `lib/absinthe/lexer.ex`
  - `unescape_unicode/5` matches the error and incomplete returns of `:unicode.characters_to_binary/1` and fails the lexer instead of emitting a bad token.
  - A leading-surrogate/trailing-surrogate escape pair is matched in the grammar and combined into one code point with the arithmetic from `StringValue()`.
  - `tokenize/2` rewinds to the start of the offending `\u` and returns `{:error, {:invalid_unicode_escape, location}}`; its `@spec` is updated. A catch-all clause is also added for any other lexer `post_traverse` error, so an unrecognized reason yields a located error rather than a `CaseClauseError`.
- `lib/absinthe/phase/parse.ex` renders that as `"Invalid Unicode escape sequence"` with line and column.

Behavior, using the schema from the issue:

```elixir
Absinthe.run(~S|{ echo(value: "a\ud800b") }|, Schema)
# before: {:ok, %{errors: [%{message: "An unknown error occurred during parsing: ..."}]}}  (no locations)
# after:  {:ok, %{errors: [%{message: "Invalid Unicode escape sequence",
#                            locations: [%{line: 1, column: 17}]}]}}

Absinthe.run(~S|{ echo(value: "\ud83d\ude00") }|, Schema)
# before: {:ok, %{errors: [%{message: "An unknown error occurred during parsing: ..."}]}}
# after:  {:ok, %{data: %{"echo" => "😀"}}}
```

Nothing that parsed before parses differently: the only inputs affected are ones that previously raised. BMP escapes (`\u0041`, `\u00f3`, `\u04F9`) are unchanged, and block strings do not process escapes, so they are untouched.

## Which spec text this follows

The surrogate rules live in the [Working Draft](https://spec.graphql.org/draft/#sec-String-Value), which is what the README says Absinthe implements. Both June 2018 and October 2021 define `EscapedUnicode` as `/[0-9A-Fa-f]{4}/` with no surrogate handling at all — under either of those editions `"a\ud800b"` has no defined value, so raising is not sanctioned there either. The draft added the braced form, the pair-combining arithmetic, and the requirement that an escape describe a Unicode scalar value or else be a parse error. I followed the draft for the pair arithmetic and the lone-surrogate rejection, and it agrees with graphql-js, the reference implementation.

Deliberately out of scope, both pre-existing and unchanged here:

- The braced `\u{...}` form is still not implemented, so `"\u{1F600}"` continues to pass through as literal text, as the issue notes. Adding it is a separate grammar addition.
- That passthrough is why: `StringCharacter` in the lexer falls back to `any_unicode`, which accepts a bare `\`, so malformed escapes like `"\uZZZZ"` and `"\u12"` become literal text rather than syntax errors. Tightening that would change behavior for documents that parse today, so it seemed wrong to fold in here.

## Validation

Run locally on Elixir 1.20.4 / OTP 29 (the top of the CI matrix is 1.20 / 28):

- `mix test` — 1509 passed, 3 excluded, 0 failures
- `SCHEMA_PROVIDER=persistent_term mix test` — 1509 passed, 3 excluded, 0 failures
- `mix format --check-formatted` — clean
- `mix dialyzer` — `Total errors: 0, Skipped: 0, Unnecessary Skips: 6`

Six tests were added and confirmed to fail against the current behavior before the fix (all six failed with the `ArgumentError` text above). In `test/absinthe/lexer_test.exs`: a surrogate pair, an unpaired leading surrogate, an unpaired trailing surrogate, and a leading surrogate followed by a non-surrogate escape. In `test/absinthe/phase/parse_test.exs`: the located error, and pair decoding through the parse phase in both letter cases.

No `@doc`/`@moduledoc` changes: the functions involved are private and `Absinthe.Lexer` has no `@moduledoc`. I did not touch `CHANGELOG.md`, since release-please generates it from commit messages.

The two commits are kept separate — the crash fix, then the pair decoding — in case you would rather take only the first. The issue asks for both.

---

Generated with Claude Code on behalf of @ethanstoner.


grempe:
Thanks for picking this up. I filed #1458 and ran its reproduction against this branch: with `absinthe` pinned to 2067bdb in https://github.com/grempe/ash-fuzz-repros, both cases in https://github.com/grempe/ash-fuzz-repros/blob/main/test/absinthe/lone_surrogate_escape_test.exs pass. The lone surrogate escape `"\ud800"` is now a syntax error without the Erlang exception text, the surrogate pair escape resolves to the code point, and the control with a valid escape still passes. Nothing else in that suite changed.

